PR #4, opened 2015-12-28
add rdata from vignette so it is usable; generate in vignette-data

GitBrianLaw:
(@malecki doing this; brian to finish branch and notify when we can review)


vignettes/datasets.Rmd
```diff
@@ -26,8 +26,10 @@ To create new datasets, multiple paths exist. In the web application, you can up
 
 You can create a dataset from any `data.frame` you have in your R session with `newDataset`:
 
-```{r, eval=FALSE}
-load("economist.RData")
+This package provides one as `economist` which loads a `data.frame` named `df` because old habits die hard.
```
nealrichardson: Why the snark? If there's a problem with the naming convention, please propose an alternative.

vignettes/datasets.Rmd
```diff
@@ -26,8 +26,10 @@ To create new datasets, multiple paths exist. In the web application, you can up
...
+
+```{r}
+data("economist", package = "crunch") # named df
```
nealrichardson: I see below that vignettes/getting-started.RData is modified in this PR, but the file is generated from https://github.com/Crunch-io/rcrunch/blob/master/vignette-data/1-getting-started.R#L24, so any changes to it need to have a corresponding change in vignette-data/1-getting-started.R.

vignettes/datasets.Rmd
malecki: Ha that was me. I do consider `df` not a great name, a result of longstanding old habits, and not great pedagogically. `economistExampleDf` or `economist` would be a better name here.

vignettes/datasets.Rmd
malecki: @GitBrianLaw can add that. I thought we did. All of the issues can be resolved by renaming in `1-getting-started.R`

```
economist <- df
save(economist, file="../data/economist.RData")
```

nealrichardson: Almost. Also need to edit the line that follows this (here, in datasets.Rmd), which is `dim(df)`, as well as L44 of https://github.com/Crunch-io/rcrunch/blob/master/vignettes/variables.Rmd.

nealrichardson:
Cleaning up our old issues. I'm going to close this and open a new issue (if we don't already have one) about using a public dataset in the vignettes so that you can play along at home.

PR #8, opened 2016-01-29
Combine mrs

persephonet:
I haven't made tests for combineArrays but to be honest I'm not sure anyone else would ever want/need that function


R/combine-arrays.R
```diff
@@ -0,0 +1,52 @@
...
+##' for the new variable that is being created. This function uses the data from 
+##' the arrays in the order that they are given. If the first variable is NA for a given 
+##' row of a subvariable, it will use the data from the second variable and so on.
+##' @param variable 1 to combine
+##' @param variable 2 to combine
+##' @param other desired attributes such as new variable alias, name, 
+##' description as well as any other variables you would like to combine
+##' @return a new variable with combine values from the arrays
+##' @export
+
```
nealrichardson: I think you need to eliminate this space and run `make doc` again. I suspect this is why you don't have an entry in your NAMESPACE file for these functions.

R/combine-mrs.R
```diff
@@ -0,0 +1,61 @@
...
+##' @param a boolean describing whether subvariables that are NA in the
+##' original variables should be made into subvariables in the new
+##' variable
+##' @param a boolean describing whether 'selected' should be given priority
+##' if false, 'not selected' will be given priority
+##' @return a new variable with combined responses from the given variables
+##' @export
+## NOTE THAT THIS IS ONLY GOING TO WORK FOR VARIABLES WHERE THE CATEGORIES ARE SELECTED AND NOT SELECTED. IF THE CATEGORIES ARE MADE BY SPSS, THIS WON'T WORK.
+## I've thought of a fix but it'll take more time (and is not currently useful to me).
+
```
nealrichardson: Same, and move the two non-documentation comments above elsewhere (perhaps inside the function)

R/add-subvariable.R
```diff
@@ -6,12 +6,26 @@
 ##' @param subvariable the subvariable to add
 ##' @return a new version of \code{variable} with the indicated subvariables
 ##' @export
+<<<<<<< HEAD
```
nealrichardson: This means that you have a merge conflict you haven't resolved.

R/add-subvariable.R
```diff
@@ -6,12 +6,26 @@
...
+addSubvariable <- function(variable, subvariable, ds=NULL){
```
nealrichardson: Instead of "ds", call it "dataset".

R/add-subvariable.R
```diff
@@ -31,11 +39,19 @@ addSubvariable <- function (variable, subvariable){
 
     ## Prune subvariable name prefix, or otherwise reset the names
     subvars <- Subvariables(crGET(absoluteURL("subvariables/", new_url)))
+    print(names(subvars))
+    print(urls(subvars))
+    print(c(subvar.urls, new_subvar.url))
+    subvars <- subvars[match(urls(subvars),c(subvar.urls, new_subvar.url))]
+    names(subvars) <- c(subvar.names[na.omit(match(urls(subvars), subvar.urls))], new_subvar.name)
+    print(names(subvars))
```
nealrichardson: You can delete these 6 lines. The three lines below do this more robustly.

tests/testthat/test-add-subvariable.R
```diff
@@ -3,6 +3,25 @@ context('Adding subvariables')
 if (run.integration.tests) {
     with(test.authentication, {
         with(test.dataset(newDatasetFromFixture("apidocs")), {
+<<<<<<< HEAD
```
nealrichardson: Another merge conflict. I reworked the test we wrote a little bit, which is the part starting on L25. That should replace your L7-13, and I recommend adapting it to cover the new case you added of supplying a VarDef.

R/combine-mrs.R
```diff
@@ -0,0 +1,60 @@
...
+##' @param other desired attributes such as new variable alias, name, 
+##' description as well as any other variables you would like to combine
+##' @param a boolean describing whether subvariables that are NA in the
+##' original variables should be made into subvariables in the new
+##' variable
+##' @param a boolean describing whether 'selected' should be given priority
+##' if false, 'not selected' will be given priority
+##' @return a new variable with combined responses from the given variables
+##' @export
+combineMRs <- function(var1, var2, ..., selectedFirst=TRUE){
```
nealrichardson: If I read the tests for this correctly, the purpose of this function is to take the subvariables from two or more MR variables and make a single MR from them--is that right? If so, how about we keep the first MR, unbind the rest, and use addSubvariables to add them to the first? Like:

```
combineMRs <- function (variable, ...) {
    newsubvars <- unlist(lapply(list(...), unbind)) ## unbind returns the URLs of the unbound variables
    return(addSubvariables(variable, newsubvars))
}
```

addSubvariables would need to take more than just one subvar, and it would need to accept URLs, but neither are too challenging (and they're more generally useful in that function itself too). And also this way, you wouldn't need separate functions to add to categorical arrays and multiple responses--same function would work.

PR #9, opened 2016-02-04
remove conflict message testing

percious:


nealrichardson:
Commenting out tests is not enough.

percious:
I understand, the tests were testing something that is no longer there, so perhaps they should be removed entirely.  I just commented them pending your review.

PR #21, opened 2016-09-12
tests support for autorollback/savepoint in appends

amol-:


R/new-dataset.R
```diff
@@ -65,7 +65,7 @@ newDatasetFromFile <- function (file, name=basename(file), ...) {
         halt("File not found")
     }
     ds <- createDataset(name=name, ...)
-    ds <- addSourceToDataset(ds, createSource(file))
+    ds <- addSourceToDataset(ds, createSource(file), savepoint=FALSE)
```
nealrichardson: What is `savepoint=FALSE`, and how does it relate to `autorollback` as an attribute on POST /batches/? Do we need two new attributes here?

It looks like you've added it everywhere there's a POST to /batches/ except for appending a dataset, where instead there's autorollback=TRUE. If it's important that they be distinct, then maybe savepoint=FALSE should be the default (in these functions) so that it's not passed around everywhere.

nealrichardson: If it's truly a new argument that we want to support, then I'd like to see an integration test here that exercises the other case (savepoint=TRUE).

amol-: `autorollback` will be the only option that users will provide usually, `savepoint=false` is a way to disable not only the rollback but the savepoint at all. Usually, like for merge the savepoint is left behind even if the user decided not to rollback to it, so it can diagnose the issue and then decided to rollback later.

`savepoint=false` totally disables the savepoint. This is done by functions that create new datasets, as there is nothing we should rollback to (previously the dataset didn't exist) and some clients might decide to delete the dataset if creation fails (whaam does so) a rollback would be pointless and inefficient. 

`savepoint=true` is actually the default so everything that appends data is working with `savepoint=true`, only the `newDataset*` functions provide `savepoint=false` and I expect we don't expose that option to users is they rely on the `new/append` functions

PR #24, opened 2016-10-13
Make rcrunch use `select` instead of identify

jjdelc:


R/export-dataset.R
```diff
@@ -46,7 +46,8 @@ variablesFilter <- function (dataset, map=FALSE) {
                 .Names=ids(allvars))))
         } else {
             ## Old
-            return(list(`function`="identify", args=list(list(id=I(urls(allvars))))))
+            return(list(`function`="select", args=list(list(map=structure(lapply(urls(allvars), function (x) list(variable=x)),
+                .Names=urls(allvars))))))
```
nealrichardson: This is unnecessary copy/paste. Do we even need the "Old" version anymore? Move the function=select to the function that calls this and consolidate the code paths.

jjdelc: This is updating the "Old" statement, I don't know where the "new" scenario is used, not supported by the API as it sends a map straight instead of a function, I think should be removed unless it has some case that needed to be supported?

nealrichardson: https://github.com/Crunch-io/rcrunch/search?utf8=%E2%9C%93&q=variablesFilter&type=Code

PR #27, opened 2017-01-31
Allow a 'message' when sharing a dataset

persephonet:
I still don't understand git so I don't know how to change this fork name to be useful. But here you go:


R/share.R
```diff
@@ -42,22 +42,28 @@ setMethod("is.editor", "PermissionTuple", function (x) {
 #' of the number of emails given.
 #' @param notify logical: should users who are getting new privileges on this
 #' dataset be sent an email informing them of this fact? Default is
+#' @param message character: what message should users who are getting notified
+#' about new permissions on this dataset receive?
```
nealrichardson: Move this new code down a line. L47 goes with L44. Also indicate the default (no message).

R/share.R
```diff
@@ -42,22 +42,28 @@ setMethod("is.editor", "PermissionTuple", function (x) {
...
 #' \code{TRUE}.
 #' @return Invisibly, the dataset.
 #' @seealso \code{\link{unshare}}
 #' @export
-share <- function (dataset, users, edit=FALSE, notify=TRUE) {
+share <- function (dataset, users, edit=FALSE, notify=TRUE, message=NA) {
```
nealrichardson: For consistency, the default should be `NULL`, not `NA`.

R/share.R
```diff
@@ -42,22 +42,28 @@ setMethod("is.editor", "PermissionTuple", function (x) {
...
     perms <- permissions(dataset)
     if (length(edit) == 1) {
         edit <- rep(edit, length(users))
     }
     if (length(edit) != length(users)) {
         halt("Must supply `edit` permissions of equal length as the number of `emails` supplied")
     }
+    if (!is.na(message) & !notify){
+        halt("Cannot send message if not notifying")
```
nealrichardson: If you think this check is necessary, add a test that shows this error being thrown.

R/share.R
```diff
@@ -42,22 +42,28 @@ setMethod("is.editor", "PermissionTuple", function (x) {
...
+    }
     payload <- lapply(edit,
         function (e) list(dataset_permissions=list(edit=e, view=TRUE)))
     names(payload) <- users
     payload$send_notification <- notify
+    if (!is.na(message)) payload$message <- message
```
nealrichardson: Move this inside the `if (notify)` block below, and since `NULL` is the default you can just do `payload$message <- message` with no further checking.

R/weight.R
```diff
@@ -44,3 +44,13 @@ weight <- function (x) {
...
 }
+
+weightVariables <- function(x) {
+    stopifnot(is.dataset(x))
+    
+    return(tryCatch(VariableCatalog(crGET(shojiURL(x, "catalogs", "weight_variables"))),
+        error=function(e){
+            if (e$message == 'No URL “weight_variables” in collection “catalogs”') return(VariableCatalog(data.frame))
+            else print(e)
+        }))
```
nealrichardson: I don't understand this error handling. You can assume that `weight_variables` will exist. That's part of the data contract.

R/weight.R
```diff
@@ -44,3 +44,13 @@ weight <- function (x) {
     }
     return(x)
 }
+
+weightVariables <- function(x) {
```
nealrichardson: This function will need to be documented and exported in order for it to be useable.

tests/testthat/test-weighting.R
```diff
@@ -20,12 +20,17 @@ with_mock_HTTP({
...
+    
     test_that("Errors are properly handled when setting weight", {
         expect_error(weight(newds) <- "a",
             "Weight must be a Variable or NULL")
         ## test error handling when trying to set non-numeric (need backend?)
     })
+    
+    test_that("weightVariables function for dataset with no weight", {
+        expect_is(weightVariables(oldds), "VariableCatalog")
+    })
```
nealrichardson: Good start, but we need to assert more about the weight variables catalog than just its existence. If there are no weights in the dataset, assert that it is length 0.

tests/testthat/test-weighting.R
```diff
@@ -77,6 +82,10 @@ with_test_authentication({
             weight(ds) <- ds$w2
             expect_equivalent(weight(ds), ds$w2)
         })
+
+        test_that("weightVariables function", {
+            expect_is(weightVariables(ds), "VariableCatalog")
```
nealrichardson: Assert that the variable(s) we expect to be in weight variables are there.

R/share.R
nealrichardson: Also, see the travis build failure for this pull request. You will need to update the man pages as well. `make doc` from the command line, or otherwise call roxygen to do that.

nealrichardson:
Ok, very close. Two final things:

1. You checked in your .Rhistory file by accident. To fix this, I recommend:

```
git checkout c1f5465ce2b76a043acb19d39969f25e679f24d5
git push -f origin weight-variables
```

That will roll your branch back to before that commit, to the place where you merged my PR.

2. Then add a line to NEWS.md noting your change, and commit that.

persephonet:
I still can't tell if I've done this right. It looks to me like there's an open pull request. Is that good enough?

nealrichardson:
It doesn't look like you've done the two steps I described above (https://github.com/Crunch-io/rcrunch/pull/27#issuecomment-279116349).

PR #32, opened 2017-05-08
api for copying multitable from another dataset.

percious:


R/multitables.R
```diff
@@ -85,25 +85,32 @@ setMethod("is.public<-", "Multitable", function (x, value) {
...
-    ## Validate inputs
-    if (missing(formula)) {
-        halt("Must provide a formula")
-    }
-    if (missing(data) || !is.dataset(data)) {
-        halt(dQuote("data"), " must be a Dataset")
-    }
+newMultitable <- function (formula, data, name, multitable=NULL, ...) {
+
+    if (missing(multitable)){
```
nealrichardson: Since adding this argument turns the whole function into an if-else, let's just make it a different function. Call it `importMultitable` or something. 

Don't forget to document it and export it from the namespace.

nealrichardson: I also would want an assignment method that would let me do e.g. `multitables(ds)[["Copied mult"]] <- multitables(other_ds)[["mult"]]` that calls importMultitable, but I won't make you learn all of the R sorcery required to do that.

R/multitables.R
```diff
@@ -85,25 +85,32 @@ setMethod("is.public<-", "Multitable", function (x, value) {
...
-    template <- formulaToQuery(formula, data)
-    if (missing(name)) {
-        name <- formulaRHS(formula)
+        payload <- wrapEntity(
+            name=name,
+            template=lapply(template$dimensions,
+                function (x) list(query=x, variable=findVariableReferences(x)))
+        )
+    } else {
+        payload = wrapEntity(name=name,
```
nealrichardson: Use `<-` for assignment in R. 

And is "name" a required attribute of this payload? I'd expect it would/should use the name of the multitable I'm copying from.

tests/testthat/test-multitables.R
```diff
@@ -190,6 +190,7 @@ with_mock_HTTP({
```
nealrichardson: I'd recommend that we set up a test with mocks so that we can test this R code quickly without a backend.

R/multitables.R
```diff
@@ -85,25 +85,32 @@ setMethod("is.public<-", "Multitable", function (x, value) {
...
+                             multitable=self(multitable))
     }
 
-    payload <- wrapEntity(
-        name=name,
-        template=lapply(template$dimensions,
-            function (x) list(query=x, variable=findVariableReferences(x)))
-    )
     ## TODO: remove "variable" from that--no longer required?
```
nealrichardson: While you're in here, how about this TODO? I think this "variable" thing was cleaned up from the API a while ago.

R/multitables.R
```diff
@@ -104,7 +105,33 @@ newMultitable <- function (formula, data, name, ...) {
         template=lapply(template$dimensions,
             function (x) list(query=x, variable=findVariableReferences(x)))
     )
-    ## TODO: remove "variable" from that--no longer required?
+
```
nealrichardson: Either restore the TODO or do it, please.

R/multitables.R
```diff
@@ -104,7 +105,33 @@ newMultitable <- function (formula, data, name, ...) {
...
+#' a default name will be derived from \code{formula}.
+#' @param ... Additional multitable attributes to set. Options include
+#' \code{is_public}.
+#' @return An object of class \code{Multitable}
+#' @examples
+#' \dontrun{
+#' m <- importMultitable(multitable, data=ds)
+#' name(m) # [1] "gender + age4 + marstat"
+#' }
+#' @export
```
nealrichardson: Since this is a new function, you'll need to run `make doc` to build the man page and add it to the NAMESPACE so that it is accessible to users.

tests/testthat/test-multitables.R
```diff
@@ -228,6 +230,12 @@ with_test_authentication({
         expect_identical(names(refresh(multitables(ds))), "Yet another name")
     })
 
+    test_that("Can copy the multitable to a new multitable", {
+        m <- importMultitable(name='copied_multitable', data=ds, multitable=mult)
+        expect_identical(name(m), "copied_multitable")
+        is.public(multitables(ds))[2] <- TRUE
```
nealrichardson: For the integration test, (1) no need to make the multitable public on L236 (rogue copy/paste, I'm guessing); (2) it should also assert something about the contents of the multitable that has been copied to show that the copy was correct. If Multitable objects had a nice print method, that would be easy, but I don't see that they do. Perhaps @jonkeane can help with that.

tests/testthat/test-multitables.R
nealrichardson: Still want this; otherwise the code coverage report will show a drop and will prevent merging the pull request.

R/multitables.R
```diff
@@ -104,7 +105,33 @@ newMultitable <- function (formula, data, name, ...) {
...
+#' multitable
+#' @param multitable an object of class \code{Multitable} that you want copied
+#' to the new dataset
+#' @param name character name to give the new multitable object. If omitted,
+#' a default name will be derived from \code{formula}.
+#' @param ... Additional multitable attributes to set. Options include
+#' \code{is_public}.
+#' @return An object of class \code{Multitable}
+#' @examples
+#' \dontrun{
```
nealrichardson: Start the example like the `newMultitable` one, then do the copy

m <- newMultitable(~ gender + age4 + marstat, data=ds)
copied_m <- importMultitable(another_ds, m)
name(copied_m) # [1] "gender + age4 + marstat"

R/multitables.R
```diff
@@ -104,7 +105,33 @@ newMultitable <- function (formula, data, name, ...) {
...
+#' m <- importMultitable(multitable, data=ds)
+#' name(m) # [1] "gender + age4 + marstat"
+#' }
+#' @export
+importMultitable <- function (data, name, multitable, ...) {
```
nealrichardson: order the args data, multitable, name, ...

R/multitables.R
```diff
@@ -104,7 +105,33 @@ newMultitable <- function (formula, data, name, ...) {
...
+
+    payload <- wrapEntity(name=name, multitable=self(multitable))
```
nealrichardson: What happens if "name" isn't provided? (Can write a unit test for this.)

R/multitables.R
nealrichardson: And consider whether "name" should be required as an argument

percious: done.

tests/testthat/test-multitables.R
percious: i wanted to maintain that a public multitable when copied remained public, so it was an intentional copy.  Also, it verifies that the multitable object actually exists.

R/multitables.R
percious: done.

tests/testthat/test-multitables.R
```diff
@@ -228,6 +230,12 @@ with_test_authentication({
         expect_identical(names(refresh(multitables(ds))), "Yet another name")
     })
 
+    test_that("Can copy the multitable to a new multitable", {
+        m <- importMultitable(ds, mult, name='copied_multitable')
+        expect_identical(name(m), "copied_multitable")
+        is.public(multitables(ds))[2] <- TRUE
+    })
+
```
percious: still need to add unit test for missing name.

tests/testthat/test-multitables.R
nealrichardson: L235 tests that the copied multitable exists. L236 doesn't assert anything: it sets a multitable to be public. If you want to assert that a public multitable, when copied, remains public, then (a) shouldn't that be tested with unit tests on the server side, and (b) to assert that here, you'd do `expect_true(is.public(m))`

R/multitables.R
```diff
@@ -108,3 +108,27 @@ newMultitable <- function (formula, data, name, ...) {
...
+
+#' Import a Multitable
+#'
+#' Allows you to import a multitable from a different dataset
+#' @param data an object of class \code{CrunchDataset} in which to create the
+#' multitable
+#' @param multitable an object of class \code{Multitable} that you want copied
+#' to the new dataset
+#' @param ... Additional multitable attributes to set. Options include
+#' \code{is_public}.
```
nealrichardson: add "name" as a possible attribute. And if you don't mind, while you're here, swap these \code{} things to either ```` or `[functionName()]` to link to docs, if appropriate.

R/show.R
```diff
@@ -254,6 +254,21 @@ setMethod("show", "CrunchLogicalExpr", function (object) {
     invisible(object)
 })
 
+showMultitable <- function (x) {
+    out <- paste("Multitable", dQuote(name(x)))
+    
+    # TODO: check variable types to alert users as needed
+    # * check if numerics are binned
+    # * check for datetime rollup resolutions
```
nealrichardson: I'm not sure that we care. If they're binned, `formatExpression` will print `bin(var)`, so the standard behavior would be fine.

tests/testthat/test-multitables.R
```diff
@@ -87,6 +87,29 @@ with_mock_HTTP({
             expect_is(mtable, "Multitable")
         })
     })
+    
+    test_that("importMultitable", {
+        with_POST("https://app.crunch.io/api/datasets/1/multitables/4de322/", {
+            mtable <- newMultitable(~ gender + mymrset, data=ds, name="New multitable")
+            expect_is(mtable, "Multitable")
+        })
+        expect_POST(m <- importMultitable(ds, mtable, name='copied_multitable'),
```
nealrichardson: For future reference, you don't need to assign (`m <-`) here. The POST request will stop before the function returns.

R/multitables.R
jonkeane: Done, do you know if `[class name]` will link to the appropriate docs?

tests/testthat/test-multitables.R
```diff
@@ -87,6 +87,29 @@ with_mock_HTTP({
...
+                    '"multitable":"https://app.crunch.io/api/datasets/1/multitables/4de322/",',
+                    '"name":"copied_multitable"}}')
+    })
+    
+    test_that("multitable show method", {
+        with_POST("https://app.crunch.io/api/datasets/1/multitables/4de322/", {
+            mtable <- newMultitable(~ gender + mymrset, data=ds, name="Shared multitable")
+            expect_is(mtable, "Multitable")
+        })
+        expect_output(mtable, paste(paste0("Multitable ", dQuote("Shared multitable")),
```
nealrichardson: Where reasonable, I try to keep to an 80-char width, you know, in case punchcards come back. A good place to line break in tests like this is between the object being tested and the expected value, so in this case, between `mtable,` and `paste(paste0(...`

R/multitables.R
nealrichardson: TBH I have to google "roxygen markdown" to find their new vignette that would answer that question every time i have to do anything more than the two things I just named.

Another complication to your question is that `Multitable` is probably both a class name and a constructor function.

I also don't know that the class/constructor documentation is useful enough to link to right now. It's not expected that users will init Multitable objects; rather, they'll create them with `newMultitable/importMultitable` or load them by extracting from the multitable catalog.

PR #47, opened 2017-05-31
Added `changeCategoryID()` function

jonkeane:
Users can now change category IDs for both categorical variables as well as (categorical) array variables.


tests/testthat/test-categories.R
```diff
@@ -318,6 +318,44 @@ with_test_authentication({
...
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, 6, "not a numeric"),
+                         "The 'to' argument is not a numeric, please providee only the id number you want to change to")
+            ds$v3f <- df$v3
+            expect_error(ds$v3f <- changeCategoryID(ds$v3f, 2, 7),
+                         "The variable v3f doesn't have categories.")
+        })
+
+        test_that("Can changeCategoryID for array variables", {
+            ds_apidocs <- newDatasetFromFixture("apidocs")
+            ds_apidocs$petloca <- ds_apidocs$petloc
```
nealrichardson: why? I don't think this is necessary.

tests/testthat/test-categories.R
```diff
@@ -318,6 +318,44 @@ with_test_authentication({
             expect_equivalent(as.array(crtabs(~ v4e, data=ds)),
                 array(c(10, 10), dim=2L, dimnames=list(v4=c("B", "D"))))
         })
+
+        test_that("Can changeCategoryID", {
+            ds$v4f <- df$v4
```
nealrichardson: Not necessary.

nealrichardson: Oh wait, disregard me, I remember why.

tests/testthat/test-categories.R
```diff
@@ -318,6 +318,44 @@ with_test_authentication({
...
+                c(1, 6, -1))
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, 2, 7),
+                "No category with id 2")
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, 1, 6),
+                         "Id 6 is already a category, please provide a new category id.")
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, "not a numeric", 7),
+                         "The 'from' argument is not a numeric, please providee only the id number you want to change from")
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, 6, "not a numeric"),
+                         "The 'to' argument is not a numeric, please providee only the id number you want to change to")
+            ds$v3f <- df$v3
```
nealrichardson: But *this* one isn't necessary.

tests/testthat/test-categories.R
```diff
@@ -318,6 +318,44 @@ with_test_authentication({
...
+            expect_identical(names(categories(ds$v4f)),
+                c("B", "C", "No Data"))
+            expect_equal(ids(categories(ds$v4f)),
+                c(1, 2, -1))
+            ds$v4f <- changeCategoryID(ds$v4f, 2, 6)
+            expect_identical(names(categories(ds$v4f)),
+                c("B", "C", "No Data"))
+            expect_equal(ids(categories(ds$v4f)),
+                c(1, 6, -1))
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, 2, 7),
```
nealrichardson: Move these input validations to unit tests that run in the mock context above, similar to this pattern: https://github.com/Crunch-io/rcrunch/blob/master/tests/testthat/test-join.R#L44-L63

That way, we get some unit test coverage on this function and don't have as steep of a coverage drop.

tests/testthat/test-categories.R
```diff
@@ -318,6 +318,44 @@ with_test_authentication({
...
+                "No category with id 2")
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, 1, 6),
+                         "Id 6 is already a category, please provide a new category id.")
+            expect_error(ds$v4f <- changeCategoryID(ds$v4f, "not a numeric", 7),
+                         "The 'from' argument is not a numeric, please providee only the id number you want to change from")
```
nealrichardson: Spelling on "providee", and this can probably be said with fewer words.

tests/testthat/test-categories.R
jonkeane: It's not, that shouldn't be there, I meant to change it back to us `ds` from above. I started that yesterday and forgot I was using that there should be a categorical array in ds.

tests/testthat/test-categories.R
```diff
@@ -318,6 +318,44 @@ with_test_authentication({
...
+                         "The variable v3f doesn't have categories.")
+        })
+
+        test_that("Can changeCategoryID for array variables", {
+            ds_apidocs <- newDatasetFromFixture("apidocs")
+            ds_apidocs$petloca <- ds_apidocs$petloc
+            expect_identical(names(categories(ds_apidocs$petloca)),
+                             c("Cat", "Dog", "Bird", "Skipped", "Not Asked"))
+            expect_equal(ids(categories(ds_apidocs$petloca)),
+                         c(1, 2, 3, 8, 9))
```
nealrichardson: Also make an assertion about the data. compare `as.vector(ds_apidocs$petloc)` or a subset of that to expected values. Then also assert about the data after `changeCategoryID`

nealrichardson: The test for a categorical also should assert about the data.

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
...
+    
+    names(categories(variable))[pos.from] <- "__TO_DELETE__"
+    categories(variable) <- c(categories(variable), newcat)
+
+    ## Move data to that new id
+    if (is.Categorical(variable)) {
+        variable[variable == from] <- to
+    } else if (is.CategoricalArray(variable)) {
+        # If the variable is a categorical array, then lapply over the subvariables
+        # can't iterate over the variable directly because of how shojicatalogs
```
nealrichardson: Rephrase this "can't" as a TODO:?

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
...
+#' @param to the (new) id for the category
+#' @return \code{variable} with an the new category id
+#' @examples
+#' \dontrun{
+#' ds$country <- changeCategoryID(ds$country, 2, 6)
+#' ds$petloc <- changeCategoryID(ds$petloc, 2, 6)
+#' }
+#' @export 
+changeCategoryID <- function (variable, from, to) {
+    # check that variable is a category
```
nealrichardson: I'm all for code comments but these validation checks are self-explanatory. If you want, add a comment above all that says that you're doing input validation.

tests/testthat/test-categories.R
jonkeane: Or, on second thought: do you mean just line 348? You're right that should be removed. But using `ds_apidocs` here is ok, or is there a categorical array in `ds` that I could use to match  the previous tests?

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
...
+    if (!has.categories(variable)) {
+        stop("The variable ", name(variable), " doesn't have categories.")
```
nealrichardson: Use "halt" instead of "stop": https://github.com/nealrichardson/httpcache/blob/master/R/log.R#L25-L38

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
...
+    # check that from is a numeric
+    if (!is.numeric(from)) {
+        stop("The 'from' argument is not a numeric, please providee only the id number you want to change from.")
+    }
+    
+    # check that to is a numeric
+    if (!is.numeric(to)) {
+        stop("The 'to' argument is not a numeric, please providee only the id number you want to change to")
+    }
+
```
nealrichardson: Probably should also assert that to and from are length 1

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
     X@.Data <- lapply(X@.Data, FUN, ...)
     return(X)
 })
+
+#' Change the id of a category for a categorical variable
+#'
```
nealrichardson: This might need another sentence or two explaining. I can provide.

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
     X@.Data <- lapply(X@.Data, FUN, ...)
     return(X)
 })
+
+#' Change the id of a category for a categorical variable
+#' 
+#' @param variable the variable in a crunch dataset that will be changed
```
nealrichardson: Note that the variable must be categorical, categorical array, or multiple response.

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
...
+#' @param from the (old) id identifying the category you want to change
+#' @param to the (new) id for the category
+#' @return \code{variable} with an the new category id
+#' @examples
+#' \dontrun{
+#' ds$country <- changeCategoryID(ds$country, 2, 6)
+#' ds$petloc <- changeCategoryID(ds$petloc, 2, 6)
```
nealrichardson: What do these two examples show differently?

R/categories.R
```diff
@@ -239,3 +239,70 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
     X@.Data <- lapply(X@.Data, FUN, ...)
     return(X)
 })
+
+#' Change the id of a category for a categorical variable
+#' 
+#' @param variable the variable in a crunch dataset that will be changed
+#' @param from the (old) id identifying the category you want to change
+#' @param to the (new) id for the category
+#' @return \code{variable} with an the new category id
```
nealrichardson: roxygen supports markdown now, so you can call that `variable` in backticks rather than the \tex stuff. Also, how about for language saying "`variable` with category `from` and all associated data values mapped to id `to`"

tests/testthat/test-categories.R
nealrichardson: I mean you don't need a copy of `petloc` to operate on because you've created this dataset just for this test. The example above where I add a variable to an existing dataset to work with is atypical.

(Also, it turns out that `ds_apidocs$petloca <- ds_apidocs$petloc` is just a renaming.)

R/categories.R
jonkeane: I intended for them to be an example of a categorical and a categorical array. although, like you you said, without a description of the dataset it's not very clear. I can drop the second if we want.

R/categories.R
jonkeane: I didn't know that, that's super nice! Though there goes my one excuse of a reason why it was good to learn tex back in the day.

tests/testthat/test-categories.R
jonkeane: Done!

tests/testthat/test-categories.R
jonkeane: done

tests/testthat/test-categories.R
jonkeane: Done, the already existing id triggers an API call so I've kept it in the integration test, but other validations are now moved.

nealrichardson: It does, but the API call is a GET that is mocked right (for the variable entity to get the categories)? So it should work in the mock context.

tests/testthat/test-categories.R
```diff
@@ -230,6 +230,26 @@ with_mock_HTTP({
...
+        expect_error(ds$gender <- changeCategoryID(ds$gender, 1, "not a numeric"),
+                     "to should be a single numeric")
+        expect_error(ds$gender <- changeCategoryID(ds$gender, c(1,-1), "not a numeric"),
+                     "to should be a single numeric")
+        expect_error(ds$gender <- changeCategoryID(ds$gender, 8, 9),
+                     "No category with id 8")
+        expect_PATCH(changeCategoryID(ds$gender, 2, 6), 
+                     'https://app.crunch.io/api/datasets/1/variables/gender/',
+                     '{"categories":[{"id":1,"missing":false,"name":"Male","numeric_value":1},{"id":2,"missing":false,"name":"__TO_DELETE__","numeric_value":2},{"id":-1,"missing":true,"name":"No Data","numeric_value":null}]} (app.crunch.io/api/datasets/1/variables/gender-1390e4-PATCH.json)')
+    })
```
nealrichardson: fwiw you can omit the file path in parentheses; the assertion will still work.

jonkeane: Done

nealrichardson:
LGTM; i'll add the NEWS bullet

PR #48, opened 2017-06-02
Append not upsert

nealrichardson:


R/append-dataset.R
```diff
@@ -32,6 +32,10 @@ appendDataset <- function (dataset1, dataset2, autorollback=TRUE) {
     ## And filter the rows, if appropriate
     payload$filter <- zcl(activeFilter(dataset2))
 
+    ## Preventatively, delete the primary key on dataset1 so that this appends
+    ## and not "upsert"
+    crDELETE(shojiURL(dataset1, "fragments", "pk"))
```
nealrichardson: This should be `pk(dataset1) <- NULL`, and then handle `NULL` assignment in `pk<-`.

R/dataset-extract.R
```diff
@@ -71,8 +71,15 @@ setMethod("[", c("CrunchDataset", "logical", "missing"), function (x, i, j, ...,
...
-    w <- match(i, allnames)
+    allvars <- allVariables(x)
+    ## Handle "namekey", which should be deprecated
+    if (getOption("crunch.namekey.dataset", "alias") == "name") {
+        alt <- names(allvars)
+    } else {
+        alt <- aliases(allvars)
+    }
+    
+    w <- whichNameOrURL(allvars, i, alt)
```
nealrichardson: I now see this logic in two places, so let's factor it out.

R/dataset.R
```diff
@@ -20,6 +20,14 @@ getNrow <- function (dataset) {
     return(nrows)
 }
 
+setFragment <- function (x, name, variable) {
```
nealrichardson: I don't think this function should exist. "fragment" isn't a meaningful, standard enough concept in our API such that a generic "setFragment" would be useful, at least not at this time. Just do whatever requesting in the `pk<-` method.

R/dataset.R
```diff
@@ -88,6 +96,38 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+    if (length(pk)) {
+        return(x[[pk[[1]]]])
+    } else {
+        return(NULL)
+    }
+})
+#' @rdname pk
+#' @export
+setMethod("pk<-", "CrunchDataset", function (x, value) {
+    payload <- toJSON(structure(list(structure(list(self(value)))),
```
nealrichardson: If `value` is `NULL`, should `crDELETE`

inst/app.crunch.io/api/datasets/3.json
```diff
@@ -20,7 +20,8 @@
     },
     "fragments": {
         "table": "https://app.crunch.io/api/datasets/3/table/",
-        "preferences": "https://app.crunch.io/api/datasets/3/preferences/"
+        "preferences": "https://app.crunch.io/api/datasets/3/preferences/",
+        "pk": "https://app.crunch.io/api/datasets/3/pk/"
```
nealrichardson: Shouldn't this mockfile below be a shoji:entity with `"pk":[]`? Check what the API actually returns.

jonkeane: For the dataset mockfile, I'm seeing just the key:value without the list syntax. There *is* a list in the actual pk.json response (both in my mock as well as what the api returns.

tests/testthat/test-appending.R
```diff
@@ -55,6 +62,16 @@ with_test_authentication({
                 expect_identical(categories(out$v4), cats)
                 expect_equivalent(as.vector(out$v3), df$v3)
             })
+
+            test_that("append removes the primary key if there is one", {
+                expect_silent(pk(part2) <- part2$v3)
+                expect_equal(pk(refresh(part2)), part2$v3)
```
nealrichardson: Don't refresh. `pk<-` should return the dataset in a good state.

tests/testthat/test-appending.R
```diff
@@ -55,6 +62,16 @@ with_test_authentication({
...
+                out <- appendDataset(part2, part1)
+                expect_null(pk(out))
+                expect_identical(dim(out), c(nrow(df)*2L, ncol(df)))
+                expect_identical(getNrow(out), nrow(df)*2L)
```
nealrichardson: use `nrow` instead of `getNrow`, that's an internal function.

tests/testthat/test-dataset-entity.R
```diff
@@ -263,6 +277,15 @@ with_mock_HTTP({
...
             '{"dashboardUrl":"https://shiny.crunch.io/example/"}}}')
     })
+
+    test_that("Primary key methods", {
+        expect_null(pk(ds2))
+        expect_identical(pk(ds), ds$birthyr)
+        expect_POST(pk(ds) <- ds$textVar,
+            'https://app.crunch.io/api/datasets/1/pk/',
+            '{"pk":["https://app.crunch.io/api/datasets/1/variables/textVar/"]}'
+        )
```
nealrichardson: Also `expect_DELETE(pk(ds) <- NULL, 'https://app.crunch.io/api/datasets/1/pk/')`

tests/testthat/test-dataset-entity.R
```diff
@@ -350,6 +373,12 @@ with_test_authentication({
             expect_true(is.Numeric(ds$name))
         })
 
+        test_that("PK methods work", {
+            expect_null(pk(ds))
+            expect_silent(pk(ds) <- ds$name)
+            expect_equal(pk(refresh(ds)), ds$name)
```
nealrichardson: Also test NULL assignment (though the `appendDataset` test will exercise that code). And remove the `refresh` here too.

inst/app.crunch.io/api/datasets/3.json
nealrichardson: What i mean is that inst/app.crunch.io/api/datasets/3/pk.json appears empty, but it should be `
{"element": "shoji:entity", "self": "https://app.crunch.io/api/datasets/3/pk/", "description": "Allows a user to get and set the primary key for a dataset", "body": {"pk": []}}`

nealrichardson: Dataset 1's `pk.json` is not empty, but dataset 3 is.

jonkeane: Aaaah, I see, ok I've added that.

R/dataset.R
```diff
@@ -88,6 +88,42 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
     invisible(setEntitySlot(x, "notes", value))
 })
 
+
+#' Name, alias, and description for Crunch objects
+#'
+#' @param x a Dataset
+#' @param value For the setters, a single Variable to use as the primary key.
```
nealrichardson: Note that the variable must have all unique values. Also note that `NULL` is valid for assignment too. 

And "setter" should be singular.

R/dataset.R
```diff
@@ -88,6 +88,42 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
     invisible(setEntitySlot(x, "notes", value))
 })
 
+
+#' Name, alias, and description for Crunch objects
```
nealrichardson: Give this man page a title that matches this function, and it probably merits an additional sentence or two that explains what a `pk` is and what effect it has.

R/dataset.R
```diff
@@ -88,6 +88,42 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+        return(NULL)
+    }
+})
+#' @rdname pk
+#' @export
+setMethod("pk<-", "CrunchDataset", function (x, value) {
+    if (is.null(value)) {
+        crDELETE(shojiURL(x, "fragments", "pk"))
+    } else {
+        payload <- toJSON(structure(list(structure(list(self(value)))),
```
nealrichardson: You may be able to express this more concisely as `list(pk=I(self(value)))` because `I` ensures that you get an array in JSON even if the R vector is length 1.

R/dataset-extract.R
```diff
@@ -151,28 +150,28 @@ setMethod("[[", c("CrunchDataset", "ANY"), function (x, i, ..., drop=FALSE) {
...
-        return(callNextMethod(x, n, ..., drop=drop))
     }
+    return(out)
 })
 #' @rdname dataset-extract
 #' @export
 setMethod("$", "CrunchDataset", function (x, name) x[[name]])
+
+
+reference_to_var <- function(x, i) {
```
nealrichardson: Maybe call this `findVariablesInDataset` or something. It's quirkiness is specific to the CrunchDataset class.

R/dataset.R
```diff
@@ -88,6 +88,47 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
     invisible(setEntitySlot(x, "notes", value))
 })
 
+
+#' Get and set the primary key for a Crunch dataset
+#'
+#' A primary key is a variable in a dataset that has a unique value for every 
+#' row. A variable must be either numeric or text type and have no duplicate or
+#' missing values.
```
nealrichardson: What does a primary key do? Why might I want one?

R/dataset.R
```diff
@@ -88,6 +88,47 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+#'
+#' @param x a Dataset
+#' @param value For the setter, a single Variable to use as the primary key.
```
nealrichardson: ... "or `NULL` to remove the primary key."

PR #51, opened 2017-06-06
toVariable can now handle AsIs classes being passed to it.

### LINKED ISSUE #44: unable to find an inherited method for function 'toVariable' for signature '"AsIs"'
A user just reported this when calling `newDataset` on a `data.frame`. It's likely some quirk of how the data was imported from an SPSS file. Regardless, this should be easy to reproduce in a test with `toVariable(I(1:5))` and fixable by adding an "AsIs" method for `toVariable` that strips off the "AsIs" class and calls `toVariable` again.

jonkeane:
This resolves #44

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,5 @@
 ### crunch 1.17.1 (under development)
+* `toVariable()` now accepts (and then strips) arguments of class `AsIs` (#44)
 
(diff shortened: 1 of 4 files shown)


nealrichardson:
LGTM. Add a note to NEWS and merge it to master please.

PR #52, opened 2017-06-06
Multitable delete and update methods

### LINKED ISSUE #33: delete and update for multitables
It would be nice to be able to delete and update multitables rather than just `newMultitable`

jonkeane:
This resolves #33


tests/testthat/test-multitables.R
```diff
@@ -40,6 +40,12 @@ with_mock_crunch({
         expect_no_request(is.public(mults)[2] <- TRUE)
     })
 
+    test_that("Multitable delete", {
+        ## Note that this PATCHes the entity, not the catalog
```
nealrichardson: Invalid comment.

tests/testthat/test-multitables.R
```diff
@@ -232,6 +268,23 @@ with_test_authentication({
                                               "  selected_array(allpets)",
                                               "  q1"))
     })
+    
+    test_that("Can make a multitable with list methods", {
+        multitables(ds)[["new mt"]] <- ~ds$country+ds$q3
```
nealrichardson: Why do you need `ds$` on the right side? Let's find a way not to require that.

tests/testthat/test-multitables.R
```diff
@@ -277,6 +330,14 @@ with_test_authentication({
         m <- importMultitable(ds2, mult)
         expect_identical(name(m), "Yet another name")
     })
+    
+    test_that("Multitable delete", {
+        mults <- multitables(ds2)
+        delete(mults[["Yet another name"]])
```
nealrichardson: Should delete require confirmation? Deleting datasets and variables do. 

Can you delete by assigning NULL into the catalog? Looks like you added that method, so we should test it.

tests/testthat/test-multitables.R
```diff
@@ -116,6 +122,36 @@ with_mock_crunch({
...
+        expect_PATCH(multitables(ds)[["Shared multitable"]] <- ~ds$gender+ds$birthyr, 
+                    "https://app.crunch.io/api/datasets/1/multitables/4de322/",
+                    '{"element":"shoji:entity","body":',
+                    '{"template":[{"query":[{"variable":"https://app.crunch.io/api/datasets/1/variables/gender/"}],',
+                    '"variable":"https://app.crunch.io/api/datasets/1/variables/gender/"},',
+                    '{"query":[{"variable":"https://app.crunch.io/api/datasets/1/variables/birthyr/"}],',
+                    '"variable":"https://app.crunch.io/api/datasets/1/variables/birthyr/"}]',
+                    '}}'
+        )
+        expect_DELETE(multitables(ds)[["Shared multitable"]] <- NULL,
```
nealrichardson: Oh, here's the NULL assignment test. Move this to the "Multitable delete" test

tests/testthat/test-multitables.R
jonkeane: This could also be specified as `as.formula(~country+q3, ds)`. If we want the user to be able to specify the formula without the dataset reference, I'll need add in code that tracks down what the dataset that the multitable belongs to. This is probably the better solution because it lets us check that the variables the user is specifying actually exist in the dataset.

tests/testthat/test-multitables.R
jonkeane: Absolutely, I'll add the confirmation

R/multitables.R
```diff
@@ -26,6 +26,41 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
     getEntity(x, i, Multitable, ...)
 })
 
+#' @rdname catalog-extract
+#' @export
+setMethod("[[<-", c("MultitableCatalog", "character", "missing", "formula"),
+          function (x, i, j, value) {
+              stopifnot(length(i) == 1)
```
nealrichardson: You should be able to do 

    ds <- loadDataset(datasetReference(x))

and then you can do `formulaToQuery` with `ds`. `loadDataset` should all be hitting the query cache so it won't be expensive.

If a `datasetReference` method for `Multitable` doesn't exist, add one. See the couple of patterns that do exist in the codebase (for when there is a dataset_url in the entity or when there isn't and you have to do a dirty URL munge)

jonkeane: Done, I've left my changes to `formulaToQuery()` since it looked from the comments around it that was how it was supposed to work, although it's not used here anymore. I can remove them if you think we won't need them. Thoughts?

R/multitables.R
```diff
@@ -99,14 +155,20 @@ newMultitable <- function (formula, data, name, ...) {
...
+
+## Internal function to do the POSTing, both in [[ and in newMultitable
+.newMultitable <- function (template, multiable_url, name, ...) {
     payload <- wrapEntity(
         name=name,
         template=lapply(template$dimensions,
-            function (x) list(query=x, variable=findVariableReferences(x)))
+                        function (x) list(query=x, variable=findVariableReferences(x)))
     )
     ## TODO: remove "variable" from that--no longer required?
```
nealrichardson: Can you do this TODO (and also remove it from where I see it copied above)? Speaking of copying, since the same logic is in two places, might it merit being factored out to a function that is reused?

R/multitables.R
```diff
@@ -99,14 +155,20 @@ newMultitable <- function (formula, data, name, ...) {
         name <- formulaRHS(formula)
     }
 
+    u <- .newMultitable(template, shojiURL(data, "catalogs", "multitables"), name, ...)
+    
+    invisible(Multitable(crGET(u)))
+}
+
+## Internal function to do the POSTing, both in [[ and in newMultitable
+.newMultitable <- function (template, multiable_url, name, ...) {
```
nealrichardson: call the second arg `catalog_url`. I was confused when I first read it thinking that you were passing in a multitable entity url.

tests/testthat/test-multitables.R
```diff
@@ -116,6 +132,34 @@ with_mock_crunch({
...
                                    "  mymrset", sep="\n"))
     })
+    
+    test_that("multitable list methods", {
+        with_POST("https://app.crunch.io/api/datasets/1/multitables/4de322/", {
+            mtable <- newMultitable(~ gender + mymrset, data=ds,
+                                    name="Shared multitable")
+            expect_is(mtable, "Multitable")
+        })
+        expect_equal(multitables(ds)[["Shared multitable"]], mtable)
```
nealrichardson: I'm not sure that L137-142 are more than tautological

jonkeane: Absolutely, I should have deleted them earlier.

R/multitables.R
```diff
@@ -26,6 +26,60 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+                  crPATCH(urls(x)[match(i, names(x))], body=toJSON(payload))
+                  return(x)
+              } else {
+                  ## Creating a new filter
+                  f <- .newMultitable(template, self(x), name = i)
+                  return(refresh(x))
+              }
+          })
+
+makeMultitablePayload <- function (template, name) {
```
nealrichardson: a more concise way to do this is `function (template, ...)` and then pass `...` to `wrapEntity`. can then remove the `if (missing(name))` check.

R/multitables.R
```diff
@@ -26,6 +26,60 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+    }
+}
+
+#' @rdname catalog-extract
+#' @export
+setMethod("[[<-", c("MultitableCatalog", "character", "missing", "NULL"),
+          function (x, i, j, value) {
+              stopifnot(length(i) == 1)
+              if (i %in% names(x)) {
+                  if (is.null(value)) {
```
nealrichardson: why is `if (is.null(value))` required? doesn't method dispatch ensure that?

R/multitables.R
```diff
@@ -26,6 +26,60 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+                      delete(x[[i]])
+                      return(invisible(NULL))
+                  }}
+          })
+
+#' @rdname catalog-extract
+#' @export
+setMethod("[[<-", c("MultitableCatalog", "numeric", "missing", "NULL"),
```
nealrichardson: why do "numeric" and "character" need their own methods? looks like they do the same thing.

jonkeane: Without the numeric method I was getting an error when trying to delete with indexes `multitables(ds)[[1]] <- NULL`

R/multitables.R
nealrichardson: I'm generally not a fan of keeping dead code around, but it's tested and it's not big, so we can keep it if you want. We will probably be in this code again soon and can reevaluate then.

R/multitables.R
```diff
@@ -26,6 +26,60 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+                  ## Creating a new filter
+                  f <- .newMultitable(template, self(x), name = i)
+                  return(refresh(x))
+              }
+          })
+
+makeMultitablePayload <- function (template, name) {
+    template <- lapply(template$dimensions,
+                       function (x) list(query=x))
+
```
nealrichardson: Thanks for doing the TODO and removing the "variable" attribute of the template thing. Can you check the fixtures in inst and remove "variable" from them too accordingly?

R/multitables.R
nealrichardson: What was the error, a method dispatch one? Did you try with signature `c("MultitableCatalog", "ANY", "missing", "NULL")`? If so, ok. I have seen some multiple dispatch confusion before, like not knowing whether to pick that one vs. `c("ShojiCatalog", "numeric", "missing", "NULL")`.

jonkeane: With`c("MultitableCatalog", "ANY", "missing", "NULL")` we don't get an error, but line 60 needed to be modified to be true whenever there's a numeric.

R/multitables.R
```diff
@@ -26,6 +26,44 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+    return(wrapEntity(template = template, ...))
+}
+
+#' @rdname catalog-extract
+#' @export
+setMethod("[[<-", c("MultitableCatalog", "ANY", "missing", "NULL"),
+          function (x, i, j, value) {
+              stopifnot(length(i) == 1)
+              if (i %in% names(x) | is.numeric(i)) {
+                  if (is.null(value)) {
```
nealrichardson: Maybe I missed your answer in the various commits, but why is the is.null check needed here? Doesn't method dispatch assure you that value is NULL?

R/multitables.R
```diff
@@ -26,6 +26,44 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+                      delete(x[[i]])
+                      return(invisible(NULL))
```
nealrichardson: `return(invisible())` is redundant. it's one or the other.

tests/testthat/test-filters.R
```diff
@@ -98,6 +98,15 @@ with_mock_crunch({
             '{"variable":"https://app.crunch.io/api/datasets/1/variables/gender/"},',
             '{"value":1}]}}')
     })
+    
+    test_that("Alter  a filter by [[<-", {
+        expect_PATCH(filters(ds)[["Occasional Political Interest"]] <- ds$gender=="Female",
```
nealrichardson: Human linter asks: can we put spaces around operators like this `==` and the `~` and `+` in the formulas?

tests/testthat/test-multitables.R
```diff
@@ -277,6 +325,23 @@ with_test_authentication({
...
         expect_identical(name(m), "Yet another name")
     })
+    
+    mults <- multitables(ds2)
+    test_that("Multitable delete requires consent", {
+    expect_error(delete(mults[["Yet another name"]]),
+                 "Must confirm deleting multitable")
+    expect_error(mults[["Yet another name"]] <- NULL,
+                 "Must confirm deleting multitable")
+    })
```
nealrichardson: the consent check is covered in the unit tests so you can delete this one

R/multitables.R
jonkeane: Ah, it's totally not, sorry I missed it.

R/multitables.R
```diff
@@ -26,6 +26,43 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+                  payload <- makeMultitablePayload(template = template)
+                  crPATCH(urls(x)[match(i, names(x))], body=toJSON(payload))
+                  return(x)
+              } else {
+                  ## Creating a new filter
+                  f <- .newMultitable(template, self(x), name = i)
+                  return(refresh(x))
+              }
+          })
+
```
nealrichardson: There should probably also be a `setMethod("[[<-", c("MultitableCatalog", "numeric", "missing", "formula"), ...` ?

R/multitables.R
```diff
@@ -26,6 +26,43 @@ setMethod("[[", c("MultitableCatalog", "numeric"), function (x, i, ...) {
...
+    return(wrapEntity(template = template, ...))
+}
+
+#' @rdname catalog-extract
+#' @export
+setMethod("[[<-", c("MultitableCatalog", "ANY", "missing", "NULL"),
+          function (x, i, j, value) {
+              stopifnot(length(i) == 1)
+              if (i %in% names(x) | is.numeric(i)) {
```
nealrichardson: What if `i` isn't in `names(x)` or isn't numeric? Add tests for that.

Consider deferring the `i` validation to L61, where `x[[i]]` will handle it. Then you just have to determine how to handle invalid `i` in this method. `NULL` assignment into R lists seems to generally not error if it doesn't match an existing element in the list:

```
> d <- list()
> d[["a"]] <- NULL
> d[[25]] <- NULL
> d
list()
```

PR #55, opened 2017-06-09
Streaming updates

jonkeane:
Added the ability to append streamed data to a dataset.


R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+
+#' Stream data to a Crunch dataset
+#' 
+#' @param ds a CrunchDataset
+#' @param data a dataframe with data to send as a stream, The given data values
+#'  must be in the Crunch I/O format (for example, category ids instead of 
+#'  names or numeric_values)
+#' @keywords internal
+streamRows <- function (ds, data) {
+    payload <- by(data, 1:nrow(data), function(row) toJSON(row) )
```
nealrichardson: I know we're not really dealing with validation, but what if data has 0 rows? (Use `seq_len(nrow(data))` instead of `1:nrow`, it's more robust.)

Odd whitespace on this line too.

Also, is it required to JSONify by row? I thought you could send `{"var1":[1,2,3], "var2":["a", "b", "c"]}`.

R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+    payload <- paste0(payload, collapse = "\n") 
+    out <- crPOST(shojiURL(ds, "fragments", "stream"), body=payload)
```
nealrichardson: Return (invisibly) something?

R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+#' @examples
+#' \dontrun{
+#' # need examples!
+#' }
+NULL
+
+
+#' @rdname streaming
+#' @export
+setMethod("pendingMessages", "CrunchDataset", function (x) {
```
nealrichardson: I'm not sure if "pendingMessages" is intuitive to a user. I feel like it should have "stream" in the function name somewhere.

Also, matter of preference, and I think both patterns exist in the package, but in cases like this where you're adding a "method" for only one object type, sometimes I'll just make it a function and have the first line be `stopifnot(is.dataset(x))`. That may be out of laziness on my part, but it may be marginally more readable than having to parse the `setMethod` syntax. Definitely your call but I wanted to mention the alternative.

R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+#' @keywords internal
+streamRows <- function (ds, data) {
+    payload <- by(data, 1:nrow(data), function(row) toJSON(row) )
+    payload <- paste0(payload, collapse = "\n") 
+    out <- crPOST(shojiURL(ds, "fragments", "stream"), body=payload)
+}
+
+#' @rdname streaming
+#' @export
+setMethod("appendStreamedRows", "CrunchDataset", function (x, ...) {
```
nealrichardson: Maybe just call it `appendStream`? So your append functions are `appendDataset` and `appendStream`.

R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+    n_msg <- pendingMessages(x)
+    if (n_msg < 1) {
```
nealrichardson: I think `if (pendingMessages(x) < 1)` is more prose-like readable.

R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+        message("There's no pending stream data to be appended.")
+        return()
```
nealrichardson: Shouldn't you return `x`?

R/dataset-stream.R
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    # stream must be after all other arguments in case a user tries to pass 
+    # stream as an argument (like name)
+    body <- wrapEntity(type = "ldjson", ..., stream = NULL)
```
nealrichardson: Why accept `...`?

See also `addBatch`, which may do what these three lines do for you.

inst/app.crunch.io/api/datasets/1.json
```diff
@@ -29,7 +29,8 @@
         "table": "https://app.crunch.io/api/datasets/1/table/",
         "preferences": "https://app.crunch.io/api/datasets/1/preferences/",
         "settings": "https://app.crunch.io/api/datasets/1/settings/",
-        "pk": "https://app.crunch.io/api/datasets/1/pk/"
+        "pk": "https://app.crunch.io/api/datasets/1/pk/",
+		"stream": "https://app.crunch.io/api/datasets/1/stream/"
```
nealrichardson: Is it GitHub, or is this line oddly indented?

tests/testthat/test-dataset-stream.R
```diff
@@ -0,0 +1,64 @@
...
+    birthyr = c(0.577530, 0.577530),
+    gender = c(2, 1),# Female , Male
+    mymrset = c(list(2, 1, 1)),
+    textVar = c("a", "b"),
+    starttime = c("1955-12-28", "1955-12-29")
+)
+
+with_mock_crunch({
+    ds <- loadDataset("test ds")   ## has 2 messages waiting, 4 rows received
+    ds2 <- loadDataset("an archived dataset", kind="archived") ## Has no streams
```
nealrichardson: "Has no streams *mock*"?

tests/testthat/test-dataset-stream.R
```diff
@@ -0,0 +1,64 @@
...
+    ds3 <- loadDataset("ECON.sav") ## has 0 messages waiting, 0 rows received
+    test_that("pendingMessages gets pending messages", {
+        expect_equal(pendingMessages(ds), 2)
+        expect_GET(pendingMessages(ds2), 'https://app.crunch.io/api/datasets/2/stream')        
+    })
+    
+    test_that("pendingMessages gets pending messages", {
```
nealrichardson: Looks like inaccurate copy/paste.

tests/testthat/test-dataset-stream.R
```diff
@@ -0,0 +1,64 @@
...
+    v3 = c(28, 29),
+    v4 = c(1, 1),
+    v5 = c("1955-11-25T00:00:00.000000+00:00", "1955-11-26T00:00:00.000000+00:00"),
+    v6 = c(1, 1)
+)
+with_test_authentication({
+    ds <- newDataset(df)
+    test_that("streamRows streams rows", {
+        expect_equal(pendingMessages(ds), 0)
+        streamRows(ds, data=stream_rows)
```
nealrichardson: assign `ds <- streamRows(ds)` and then don't refresh on the next line.

tests/testthat/test-dataset-stream.R
```diff
@@ -0,0 +1,64 @@
...
+        expect_equal(pendingMessages(refresh(ds)), 1)
+    })
+
+    test_that("appendStreamedRows appends all pending rows", {
+        appendStreamedRows(ds)
```
nealrichardson: likewise assign the return of `appendStream` and don't refresh below. The function returns the refreshed dataset.

R/dataset-stream.R
jonkeane: I went back and forth myself, I'll bump them down to just functions for now. If we want to implement a stream from R where you can do something like `stream(ds) <- [some data frame]` then we can align all of these together, but for now it's probably more than is necessary.

R/dataset-stream.R
jonkeane: I was trying to allow the users to pass in names and descriptions if they would like. They are documented in the API, but we might not need to give tat kind of flexibility (and it would also ensure a user doesn't try to do `stream=1000`) Can you think of a reason a user would want to be able to name/describe these batches?

inst/app.crunch.io/api/datasets/1.json
jonkeane: This was some strange atom bug. In atom it looked totally fine, but anywhere else it was too tabbed. Should be fixed now.

R/dataset-stream.R
jonkeane: The API docs say it should be one row per line. I've tried with lists like`{"var1":[1,2,3], "var2":["a", "b", "c"]}`, and when I try to append data streamed I get an error about decoding: `Cannot decode untyped 'list' value to class 'datetime'. (for variable 000004)` So, I do think this has to be JSONified by row. I've fixed the other problems (including not posting anything if there are no rows since there's no reason to).

R/dataset-stream.R
```diff
@@ -0,0 +1,62 @@
...
+#'
+#' Crunch allows you to stream data to a dataset. Streaming data is useful for
+#' datasets which have frequent updates (see 
+#' [the Crunch documentation](http://docs.crunch.io/#streaming-rows) for more)
+#' information. Crunch automatically appends streamed data perioadically,
+#' however if you would like to trigger appending pending streamed data to a 
+#' dataset you can use `appendStream()`.
+#'
+#' @param ds a CrunchDataset
+#' @param ... Additional batch attributes to set when appending pending stream
```
nealrichardson: Nah, you can remove this. No one names batches. Easy to add if that need somehow arises, but until them let's keep it simple.

R/dataset-stream.R
```diff
@@ -0,0 +1,62 @@
...
+appendStream <- function (ds, ...) {
+    stopifnot(is.dataset(ds))
+    if (pendingStream(ds) < 1) {
+        message("There's no pending stream data to be appended.")
+        return(ds)
+    }
+
+    # stream must be after all other arguments in case a user tries to pass 
+    # stream as an argument (like name)
+    # body <- wrapEntity(type = "ldjson", ..., stream = NULL)
```
nealrichardson: Can delete your commented code.

tests/testthat/test-dataset-stream.R
```diff
@@ -0,0 +1,66 @@
...
+    test_that("pendingStream gets pending messages", {
+        expect_equal(pendingStream(ds), 2)
+        expect_GET(pendingStream(ds2), 'https://app.crunch.io/api/datasets/2/stream')        
+    })
+    
+    test_that("streamRows streams rows", {
+        expect_equal(streamRows(ds, data=data.frame()), ds)
+        expect_POST(streamRows(ds, data=mock_stream_rows),
+                    'https://app.crunch.io/api/datasets/1/stream/',
+                    '{"birthyr":0.5775,"gender":2,"mymrset.2":2,"mymrset.1":1,"mymrset.1.1":1,',
```
nealrichardson: `"mymrset.2"` isn't right. Array data should be streamed keyed by alias of the subvariables (as if it wasn't an array). So for this mock, it's `subvar1`, `subvar2`, and `subvar3`. Edit your `mock_stream_rows` to replace `mymrset` with those.

R/dataset-stream.R
```diff
@@ -0,0 +1,62 @@
...
+    }
+    invisible(refresh(ds))
+}
+
+#' Append data that has been streamed to a dataset
+#'
+#' Crunch allows you to stream data to a dataset. Streaming data is useful for
+#' datasets which have frequent updates (see 
+#' [the Crunch documentation](http://docs.crunch.io/#streaming-rows) for more)
+#' information. Crunch automatically appends streamed data perioadically,
```
nealrichardson: spelling, and "...; however, ..."

R/dataset-stream.R
```diff
@@ -0,0 +1,62 @@
...
+pendingStream <- function (ds) {
+    stopifnot(is.dataset(ds))
+    stream_cat <- ShojiEntity(crGET(shojiURL(ds, "fragments", "stream")))
+    stream_cat$pending_messages
+}
+
+#' Stream data to a Crunch dataset
+#' 
+#' @param ds a CrunchDataset
+#' @param data a dataframe with data to send as a stream, The given data values
```
nealrichardson: `data.frame`

R/dataset-stream.R
```diff
@@ -0,0 +1,62 @@
+#' Get the pending streams for a dataset
+#'
+#' Crunch allows you to stream data to a dataset. Streaming data is useful for
+#' datasets which have frequent updates (see 
+#' [the Crunch documentation](http://docs.crunch.io/#streaming-rows) for more)
+#' information. `pendingMessages()` retrieves the number of pending messages and `appendStreamedRows()` appends all pending streamed rows to the dataset.
```
nealrichardson: Update the function names. And if you do `[functionName()]`, it will link to the man page for that function.

jonkeane: I actually reduced this description a lot since most of it is already covered in the much more likely to be used function `appendStream()`, but added the linking nonetheless.

nealrichardson:
LGTM. Let's add a NEWS entry and merge this.

PR #56, opened 2017-06-15
Add the ability to make case variables

jonkeane:


R/case-variables.R
```diff
@@ -0,0 +1,57 @@
...
+                 missing=case@missing)
+        })))
+    new_cat_ids <- sapply(new_cat_type$value$categories, function(x) x$id)
+    new_cat <- list(column=I(new_cat_ids), type=new_cat_type)
+    
+    casevar <- list(...)
+    casevar$derivation <- zfunc("case", c(list(new_cat), lapply(cases, zcl)))
+
+    # remove a level of duplicate embedding
+    casevar$derivation$args <- casevar$derivation$args[[1]]
```
nealrichardson: [zfunc](https://github.com/Crunch-io/rcrunch/blob/master/R/zcl.R#L45) takes `...` so you can probably rewrite L50 such that this hack isn't necessary.

R/case-variables.R
```diff
@@ -0,0 +1,57 @@
...
+#' The `makeCaseVariable` function derives a variable using values from other 
+#' variables. These are evaluated in the order they are supplied in the list 
+#' as the `cases` argument (they proceed in an IF, ELSE IF, ELSE IF, ..., ELSE 
+#' fashion); the first one that matches selects the corresponding value from 
+#' the [`Case`] object.
+#' 
+#' @param cases A list of objects of class [`Case`]
+#' @param ... other properties to pass about the case variable (e.g. name)
+#' @export
+makeCaseVariable <- function (cases, ...) {
```
nealrichardson: Looking at the code, I'm not convinced that the `Case` class helps anything. (And I realize that I'm probably responsible for sending you down that path.)  An S4 class might make more sense if there were some input constraints you could enforce in the constructor, "Cases" don't really have meaning or validity outside of the totality of the case statement--i.e. you can't have a default "id" for a Case generically. So then `Case()` could be done as `list()` with some validation inside the variable building function.  

How about something like this: 

    makeCaseVariable <- function (..., cases, else, name) {
        dots <- list(...)
        if (missing(cases)) {
            is_expr <- function (x) inherits(x, "CrunchLogicalExpression")
            exprs <- Filter(is_expr, dots)
            dots <- Filter(Negate(is_expr), dots)
            cases <- mapply(function (expr, name) list(name=name, expression=expr), 
                expr=exprs, name=names(exprs))
        }
        # Else (and continuing), `cases` is a list of list objects with minimum 
        # "name" and "expression".
        # Now, fill in the potentially omitted attributes (id, missing, etc.)
        # and validate them
        cases <- ensureValidCases(cases)
        ...
    }

If you supply "cases", it works similar to how `combine` does (like `cases=list(list(name="Young men", expression=ds$gender == "Male" & ds$age < 34), ...)`. Alternatively, the function provides a shorter version where you could do something like 

    makeCaseVariable(
        `Young men`=ds$gender == "Male" & ds$age < 34, 
        `Old men`=ds$gender == "Male" & ds$age >=35,
        else="Women",
        name="Men by age group",
        description="More words about this variable"
    )

when you don't care about numeric values and ids and such.

R/case-variables.R
jonkeane: I've changed the way this is done, so it's less of a problem now, but I kept on having issues where there was an extra layer of embedding (I think because `zfunc` calls `list(...)`, so if you pass a list in you'll get a second level). When I tried just passing the cases in alone like `zfunc("case", new_cat, [list of expressions extracted from cases])` the case expressions were always embedded in their own list in the end. 

I could change zfunc to recognize when it's getting lists, but if it's just for this one (admittedly clunky) endpoint, it's probably not worth it, thoughts?

R/case-variables.R
```diff
@@ -0,0 +1,149 @@
...
+#' 
+#' @keywords internal
+ensureValidCases <- function(case) {
+    if (!is.list(case)) {
+        halt("A case must be a list")
+    }
+    wrong_case_names <- setdiff(names(case), c("id", "name", "expression",
+                                               "numeric_value", "missing"))
+    if (length(wrong_case_names) > 0) {
+        halt("each case must have at most an id, name, expression, numeric_value, ",
```
nealrichardson: You could `serialPaste(wrong_case_names)` in the error message so the user can see what they provided wrong. I've found that helpful in the past.

R/case-variables.R
```diff
@@ -0,0 +1,149 @@
...
+                                               "numeric_value", "missing"))
+    if (length(wrong_case_names) > 0) {
+        halt("each case must have at most an id, name, expression, numeric_value, ", 
+             "and missing element")
+    }
+    if (!is.null(case$id) && !is.integer(case$id)) {
+        halt("a case's id must be an integer")
+    }
+    if (!is.character(case$name)) {
+        halt("a case's name must be a character and must not be NULL")
```
nealrichardson: Isn't "not be NULL" redundant?

jonkeane: True, I was trying to emphasize that the name is mandatory, but this is probably not the place for that (I have that in the docs already, so don't need to repeat myself).

R/case-variables.R
```diff
@@ -0,0 +1,149 @@
...
+    if (missing(cases) & length(exprs) > 0) {
+        # save any member of dots that isn't an expression for use later.
+        dots <- Filter(Negate(is_expr), dots)
+        cases <- mapply(function (expr, name) list(name=name, expression=expr),
+                        expr=exprs, name=names(exprs),
+                        SIMPLIFY = FALSE, USE.NAMES = FALSE)
+    }
+    
+    # make ids if the cases don't have ids.
+    need_ids <- sapply(cases, function(x) is.null(x$id))
```
nealrichardson: hadley recommends `vapply(cases, function (x) is.null(x$id), logical(1))` in package code, rather than sapply, which can return different shapes depending on the input.

R/case-variables.R
```diff
@@ -0,0 +1,149 @@
...
+    cases[need_ids] <- lapply(seq_along(cases[need_ids]), function(i) {
```
nealrichardson: Won't this break if I supply

```
cases=list(
    list(name="case 1", id=1, expression=ds$x == 1), 
    list(name="case 2", expression=ds$x == 2)
)

File: R/case-variables.R [hunk 7]
```diff
@@ -0,0 +1,149 @@
...
+        cases[need_ids][[i]]$id <- i
+        return(cases[need_ids][[i]])
+    })
+    
+    # NULLify numeric_value and FALSE missing if they are empty
+    cases <- lapply(cases, function(x) {
```
nealrichardson: You could do this logic inside `ensureValidCases`, if you didn't find that too much for the function to do.

R/case-variables.R
```diff
@@ -0,0 +1,149 @@
...
+        }
+        if(!is.character(else_case$name)) {
+            halt("else_cases must have a (character) name")
+        }
+        if(!is.integer(else_case$id)) {
+            halt("id must be an integer")
+        }
+        cases <- append(cases, list(else_case))
+    }
+
```
nealrichardson: I might be missing it, but I think you're missing the check that `id` is unique across case definitions.

R/case-variables.R
jonkeane: I was on the fence about this, but given how much `makeCaseVariable` has grown I'll go ahead and move it.

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+    }
+    
+    # create the new categorical variable
+    new_cat_type <- list(
+        value=list(class="categorical",
+        categories=lapply(cases, function (case) case[c("id", "name", "numeric_value", "missing")])))
+    new_cat_ids <- sapply(new_cat_type$value$categories, function(x) x$id)
+    new_cat <- list(column=I(new_cat_ids), type=new_cat_type)
+    
+    casevar <- dots
```
nealrichardson: No reason you can't call `dots` `casevar` from the beginning.

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+            halt("id must be an integer")
+        }
+        cases <- append(cases, list(else_case))
+    }
+    
+    # create the new categorical variable
+    new_cat_type <- list(
+        value=list(class="categorical",
+        categories=lapply(cases, function (case) case[c("id", "name", "numeric_value", "missing")])))
+    new_cat_ids <- sapply(new_cat_type$value$categories, function(x) x$id)
```
nealrichardson: vapply

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+    new_cat <- list(column=I(new_cat_ids), type=new_cat_type)
+    
+    casevar <- dots
+    casevar$name <- name
+    casevar$derivation <- zfunc("case", new_cat)
+
+    # add case_expressions, remove nulls (should only be from the else case)
+    case_exprs <- lapply(cases, function(x) zcl(x$expression))
+    case_exprs <- Filter(Negate(is.null), case_exprs)
+    casevar$derivation$args <- append(casevar$derivation$args, case_exprs)
```
nealrichardson: Why `append` instead of `c`?

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+#' properties to pass about the case variable (i.e. alias, description)
+#' @param cases a list of lists with each case condition to use each must 
+#' include at least a `name` and an `expression` element.
+#' @param else_case a single list that has at least a `name` to serve as the 
+#' case when no others match (if not specified Crunch will use the system 
+#' default "No data")
+#' @param name a character to use as the name of the case variable to create
+#' @export
+makeCaseVariable <- function (..., cases, else_case, name) {
+    dots <- list(...)
```
nealrichardson: You could probably do `casevar <- list(..., name=name)` and save yourself two lines later (and get your input validation about name being required done in the process).

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+    if (missing(cases) & length(exprs) > 0) {
+        # save any member of dots that isn't an expression for use later.
+        dots <- Filter(Negate(is_expr), dots)
+        cases <- mapply(function (expr, name) list(name=name, expression=expr),
+                        expr=exprs, name=names(exprs),
+                        SIMPLIFY = FALSE, USE.NAMES = FALSE)
+    }
+    
+    # make ids if the cases don't have ids.
+    need_ids <- vapply(cases, function (x) is.null(x$id), logical(1))
```
nealrichardson: I think the intent of L96-L134 would be clearer if it were in a single function `ensureValidCases` (and so you'd rename the function that currently exists to `ensureValidCase`, which is more correct anyway).

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+    all_ids <- vapply(cases, function (x) x$id, integer(1))
+    if (anyDuplicated(all_ids) > 0) {
+        halt("there are duplicate ids provided: ", serialPaste(all_ids))
+    }
+    
+    cases <- lapply(cases, ensureValidCases)
+
+    if (!missing(else_case)) {
+        # assign API default values if not given
+        if(is.null(else_case$id)) {
```
nealrichardson: This series of `if`s could be done more concisely with `modifyList` (in `utils`, already imported elsewhere)

R/case-variables.R
nealrichardson: and see the `vget` function in misc.R. You could do this as `vapply(cases, vget("id"), numeric(1))`

R/case-variables.R
```diff
@@ -0,0 +1,154 @@
...
+        dots <- Filter(Negate(is_expr), dots)
+        cases <- mapply(function (expr, name) list(name=name, expression=expr),
+                        expr=exprs, name=names(exprs),
+                        SIMPLIFY = FALSE, USE.NAMES = FALSE)
+    }
+    
+    # make ids if the cases don't have ids.
+    need_ids <- vapply(cases, function (x) is.null(x$id), logical(1))
+    supplied_ids <- vapply(cases[!need_ids], function (x) x$id, integer(1))
+    new_ids <- seq_along(cases) # generate too many ids
```
nealrichardson: Ask Jj to review this logic. He's done this kind of thing a few times on the backend, and dealt with lots of bugs/corner cases along the way.

R/case-variables.R
jonkeane: I think that would be problematic in the case where dots get filtered to be only the non-expression elements? (When we call `makeCaseVariable(`Cats`=ds$q1 == "Cat",name="name")`) Let me look into it more though.

R/case-variables.R
jonkeane: This was a vestige from testing, it should have been `c`.

R/case-variables.R
```diff
@@ -0,0 +1,163 @@
...
+        cases[need_ids][[i]]$id <- new_ids[i]
+        return(cases[need_ids][[i]])
+    })
+    
+    # check there are no duplicate ids and they are less than 2^15-1
+    all_ids <- vapply(cases, function (x) x$id, integer(1))
+    if (anyDuplicated(all_ids) > 0) {
+        halt("there are duplicate ids provided: ", serialPaste(all_ids))
+    }
+    if (any(all_ids > 2^15-1)) {
```
nealrichardson: Just use `32767` here, it's more readable

R/case-variables.R
```diff
@@ -0,0 +1,169 @@
...
+
+## Check for "integers" but don't require true integer storage
+is.whole <- function (x) is.numeric(x) && floor(x) == x
+
+ensureValidCases <- function(cases, else_case) {
+    cases <- lapply(cases, ensureValidCase)
+
+    if (!missing(else_case)) {
+        # assign API default values if not given
+        defaults <- list(id=NULL, name=NULL, numeric_value=NULL, missing=FALSE)
```
nealrichardson: Why not do this same treatment in `ensureValidCase`?

R/case-variables.R
```diff
@@ -0,0 +1,169 @@
...
+        else_case <- modifyList(defaults, else_case)
+
+        if(!is.null(else_case$expression)) {
+            halt("else_cases should not have any conditions expression")
+        }
+        if(!is.character(else_case$name)) {
+            halt("else_cases must have a (character) name")
+        }
+        if(!is.null(else_case$id) & !is.whole(else_case$id)) {
```
nealrichardson: This seems to duplicate `ensureValidCase`: why not run the "else_case" through there?

R/case-variables.R
```diff
@@ -0,0 +1,169 @@
...
+            halt("id must be an integer")
+        }
+        cases <- c(cases, list(else_case))
+    }
+
+    # make ids if the cases don't have ids.
```
nealrichardson: Can you factor this id-filling logic to a separate function? Not that we'll likely reuse it, but just to make this function more readable.

R/case-variables.R
```diff
@@ -0,0 +1,169 @@
...
+    supplied_ids <- vapply(cases[!need_ids], vget("id"), numeric(1))
+    new_ids <- seq_along(cases) # generate too many ids
+    new_ids <- setdiff(new_ids, supplied_ids) # discard ids where there is overlap
+    cases[need_ids] <- lapply(seq_along(cases[need_ids]), function(i) {
+        cases[need_ids][[i]]$id <- new_ids[i]
+        return(cases[need_ids][[i]])
+    })
+
+    # check there are no duplicate ids and they are less than 2^15-1
+    all_ids <- vapply(cases, vget("id"), numeric(1))
```
nealrichardson: Could also include the `id` validation in that factored-out function. Then `ensureValidCases` reads like:

* Deal with "else", if exists
* Validate each case's fields
* Fill in ids and make sure they're valid as a set

R/case-variables.R
```diff
@@ -0,0 +1,169 @@
...
+    if (anyDuplicated(all_ids) > 0) {
+        halt("there are duplicate ids provided: ", serialPaste(all_ids))
+    }
+    if (any(all_ids > 32767L)) {
+        halt("id must be less than 32,768, this might be a result of too many cases being used.")
+    }
+    if (any(all_ids < 1L)) {
+        halt("id must not be less than 1")
+    }
+
```
nealrichardson: Is there any validation missing? Two things I see:

1. I could supply duplicate category/case "names"
2. "missing" could be `NA` and pass validation

R/case-variables.R
```diff
@@ -0,0 +1,169 @@
...
+#'
+#' 1. You can also use the `cases` argument, which is useful when you want to
+#' prespecify ids, or set numeric_values:
+#' `makeCaseVaraible(cases=list(list(expression=ds$v1 == 1, name="case1"), list(expression=ds$v2 == 2, name="case2")), name="new case")`
+#'
+#' @param ... a sequence of named expressions to use as cases as well as other
+#' properties to pass about the case variable (i.e. alias, description)
+#' @param cases a list of lists with each case condition to use each must
+#' include at least a `name` and an `expression` element.
+#' @param else_case a single list that has at least a `name` to serve as the
```
nealrichardson: Did you try my idea of allowing an "expression" be `"else"` as a magic value, so that you can just define "cases" and not have a special "else_case" argument?

R/interact.R
```diff
@@ -0,0 +1,68 @@
...
+#' @param ... A sequence of categorical variables to make an interaction of
+#' @param name a character to use as the name for the interaction variable
+#' 
+#' @return A [`VariableDefinition`] that will create the new
+#' interaction variable. 
+#' 
+#' @export
+interactVariables <- function(..., name) {
+    dots <- list(...)
+    is_expr <- function (x) inherits(x, "CrunchVariable")
```
nealrichardson: `is_expr` isn't right here (looks like stale copy/paste). Looks like what you're trying to say is `is.variable` (which exists)

R/case-variables.R
jonkeane: I've fixed both of these, as well as adding a check in case someone tries to supply multiple cases in a single list (eg `cases=list(name=c("A", "B"), expression=c(ds$v1 == "A", ds$v2 == "B"))`, unlikely, but the error that is produced when that is done isn't super transparent.). I've also added a check if any of the expressions are duplicates. This wouldn't *break* anything since whichever is first will be matched will be used, I could imagine someone accidentally doing this and being confused by the result.

R/interact.R
```diff
@@ -0,0 +1,66 @@
+#' Make case statments that define the interaction of two or more other variables
+#'
+#' @param ... A sequence of categorical variables to make an interaction of
+#' @param drop should unused levels be dropped? (currently not implemented, so `FALSE`)
+#' @param sep a character to separate the values of the individual variables (default: `:`)
+#'
+#' @return A list of cases that can be passed to [`makeCaseVariable`] to make
+#' the interaction variable.
+#'
+#' @export
```
nealrichardson: This seems like an internal function. Why is it exported?

R/interact.R
```diff
@@ -0,0 +1,66 @@
...
+makeInteractions <- function(..., drop = FALSE, sep = ":") {
```
nealrichardson: I'd drop `drop` since it's not implemented.

R/interact.R
```diff
@@ -0,0 +1,66 @@
...
+
+    if (length(vars) < 2) {
+        halt("must supply more than one variable to make an interaction")
+    }
+
+    case_args <- c(vars, drop=other_dots[["drop"]], sep=other_dots[["sep"]])
+    cases <- do.call(makeInteractions, case_args)
+
+    int_args <- c(cases=list(cases), name=name, other_dots)
+    interactionVar <- do.call(makeCaseVariable, int_args)
```
nealrichardson: So it looks like if I pass in "sep" as an arg here, that gets passed to `makeCaseVariable` too, which means it would try to add that as some variable metadata field in the VariableDefinition? That's not right. Why not make "sep" a function argument?

R/interact.R
```diff
@@ -0,0 +1,66 @@
...
+#' Make a variable that is the interaction of two or more other variables
+#'
+#' @param ... A sequence of categorical variables to make an interaction of
+#' @param name a character to use as the name for the interaction variable
+#'
+#' @return A [`VariableDefinition`] that will create the new
+#' interaction variable.
+#'
+#' @export
+interactVariables <- function(..., name) {
```
nealrichardson: Did you try the method dispatch on `interact`?

R/interact.R
```diff
@@ -0,0 +1,66 @@
...
+}
+
+#' Make a variable that is the interaction of two or more other variables
+#'
+#' @param ... A sequence of categorical variables to make an interaction of
+#' @param name a character to use as the name for the interaction variable
+#'
+#' @return A [`VariableDefinition`] that will create the new
+#' interaction variable.
+#'
```
nealrichardson: This man page needs some more explanation I think, as well as an example.

R/interact.R
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    if (drop) halt(dQuote("drop"), "has not been implemented yet")
+    cats <- lapply(vars, categories)
+    all_combos <- expand.grid(cats)
+
+    cases <- by(all_combos, seq(nrow(all_combos)), makeCase, vars=vars, sep=sep)
+
+    # collpase and remove names
+    cases <- c(cases)
```
nealrichardson: This line is a no-op, right?

PR #60, opened 2017-06-23
146889237: added the ability to get and set digits for variables.

jonkeane:


R/variable.R
```diff
@@ -43,6 +43,27 @@ setMethod("notes", "CrunchVariable", function (x) tuple(x)$notes)
 setMethod("notes<-", "CrunchVariable",
     function (x, value) setTupleSlot(x, "notes", value %||% ""))
 
+#' @rdname describe
+#' @export
+setMethod("digits", "CrunchVariable", function (x) {
+    var_entity <- entity(x)
+    return(var_entity@body$format$data$digits %||% NULL)
```
nealrichardson: `%||% NULL` is superfluous

Issue #61, opened 2017-06-26
Add vignette for 🗺ing


malecki:
There are a lot of things that might be nice, but at a minimum I need to be able to see for a given variable if it has a geodatum associated with it and be able to fetch the {geo,topo}json without manually doing `crGET`.

Niceties could be downloading the object, checking intersection of category `names` to values of the specified `feature_key`, and setting the `feature_key` to a different property of each feature through https://github.com/ropensci/geojsonio.

needs a bit of r api design work. the geodata catalog (considered read-only, maintained by us) lives at https://app.crunch.io/api/geodata/ , and the variable’s association is defined [on its view](http://docs.crunch.io/#associating-variables-with-geodata):
```
{"view": { "geodata": [
        {"geodatum": "<uri>",
         "feature_key": "properties.name"}
    ]}
}
```

malecki:
I would also like to make a vignette showing maps-with-crunch are easy and fun

PR #62, opened 2017-06-29
First step of geo features

jonkeane:
Accomplishes many of the goals in #61, but it definitely needs a bit more user-input checking before it's good to go, possibly other tweaks and changes.


nealrichardson:
General thoughts:

What is the objective of this feature?
* I would expect that the first things we'd want to do would be to let a user view and modify geo settings on a variable. It appears that the PR does this.
* But then how does the geo get set on the variable in the first place? Don't we need to expose the catalog of available geo shape files or whatever? 
* Then perhaps add utilities to figure out which fields map onto the categories of the variable and handle that complexity?

Why would I want to `fetchGeoFile`? 
* I don't think we should install geojsonio on Travis. I can't exactly tell what the fetchGeoFile test is testing, but it looks like the it is just checking the file extension behavior. That seems pointless.
* If we think that function is worth testing as a e2e test, fine. Let's write that instead and run it locally/on Jenkins. 

Do we really need/want geojsonio? If I don't need to fetchGeoFile, then I have no need for geojsonio.
* Is there no other alternative to geojsonio that doesn't require manual system dependency installation? I've worked with spatial data in R before and never messed with this.
* README and Makefile would need updating with instructions for installing geojsonio

R/as-data-frame.R
```diff
@@ -1,34 +1,64 @@
...
+CrunchDataFrame <- function (dataset, order) {
     ## S3 constructor method for CrunchDataFrame. terms.formula doesn't seem
     ## to like S4 subclass of environment
     stopifnot(is.dataset(dataset))
     out <- new.env()
     out$.crunchDataset <- dataset
+    out$.names <- c()
+    # set the order of the dataset based on order function, if no order is 
+    # given return all rows in the dataset in the order they appear
+    if (is.null(order)) {
```
nealrichardson: Unless I'm missing something, both sides of this `if` are identical.

R/as-data-frame.R
```diff
@@ -1,34 +1,64 @@
-CrunchDataFrame <- function (dataset) {
+CrunchDataFrame <- function (dataset, order) {
     ## S3 constructor method for CrunchDataFrame. terms.formula doesn't seem
     ## to like S4 subclass of environment
     stopifnot(is.dataset(dataset))
     out <- new.env()
     out$.crunchDataset <- dataset
+    out$.names <- c()
```
nealrichardson: Why not make this `out$.names <- aliases(allVariables(dataset))` and avoid the `.names <- c(.names, .a)` below?

R/as-data-frame.R
```diff
@@ -1,34 +1,64 @@
-CrunchDataFrame <- function (dataset) {
+CrunchDataFrame <- function (dataset, order) {
```
nealrichardson: 1) I'd suggest a different name for "order"; I didn't realize it was for permuting/selecting row indices until I read below; also, it's the name of a function.

2) It appears below that you assume `NULL` is the default, but none is specified here.

R/as-data-frame.R
```diff
@@ -1,34 +1,64 @@
...
-                list(v=a)))
+        for (.a in aliases(allVariables(dataset))) {
+            eval(substitute(delayedAssign(v, {
+                if (is.null(.order)) {
+                    as.vector(.crunchDataset[[v]])
+                } else {
+                    # as.numeric(factor(.order)) gets the dense ranking of the 
+                    # ordering so that the non-duplicated results returned in 
+                    # crunch order can be used in the order specified in .order.
+                    as.vector(.crunchDataset[[v]][.order])[as.numeric(factor(.order))]
```
nealrichardson: This is confusing to me, both the code and the comment.

R/as-data-frame.R
```diff
@@ -1,34 +1,64 @@
...
+names.CrunchDataFrame <- function (x) x$.names
 
 #' as.data.frame method for CrunchDataset
 #'
 #' This method is defined principally so that you can use a CrunchDataset as
-#' a \code{data} argument to other R functions (such as
-#' \code{\link[stats]{lm}}). Unless you give it the \code{force==TRUE}
-#' argument, this function does not in fact return a \code{data.frame}: it
+#' a `data` argument to other R functions (such as
+#' `\link[stats]{lm}`). Unless you give it the `force==TRUE`
```
nealrichardson: `[stats::lm()]`

R/as-data-frame.R
```diff
@@ -73,3 +104,96 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
     names(out) <- names(x)
     return(structure(out, class="data.frame", row.names=c(NA, -nrow(x))))
 }
+
```
nealrichardson: Seems like there should also be a `[` method for CrunchDataFrame that lets you set `order`, but maybe that's too much scope creep for this PR.

R/expressions.R
```diff
@@ -166,10 +166,20 @@ setMethod("!", "CrunchExpr", function (x) zfuncExpr("not", x))
...
         identical(as.numeric(head(table, 1):tail(table, 1)), as.numeric(table))) {
 
+        # set beg(ining) and end appropriately in case we have been given a rev
+        # sequence (eg 20:1) ZCL returns nothing if asked for between 20 and 1
+        if (head(table, 1) < tail(table, 1)) {
+            beg <- head(table, 1)
+            end <- tail(table, 1)
+        } else {
+            beg <- tail(table, 1)
+            end <- head(table, 1)
```
nealrichardson: But won't this return out-of-order results?

R/as-data-frame.R
jonkeane: Absolutely, I should have dropped the if/else. Holdover from when there was something slightly different.

R/as-data-frame.R
jonkeane: I agree, I wasn't fully happy with `order` anyway and 1 is a very valid thing to ignore. I've changed it to `selection` which at a minimum avoids 1, but might still not be the best name (open to other suggestions!)

R/as-data-frame.R
jonkeane: This is probably a hack, but was the only way I could find to do what I needed. (Looking back, and digging through some source of [`dense_rank`](https://github.com/tidyverse/dplyr/blob/master/R/rank.R#L72-L73) we really probably just need `match(r, sort(unique(r)))` here instead which is much clearer, at least to me.) When we send subsets by rownumber to the server the server actually returns them ordered like they were in the dataset not like they were requested. But what is returned is only the rows that were selected, so we can't use `.order` here again to match the ordering of the rows that are returned, because they frequently will not be the correct indices for what is returned.

But ultimately what the `[as.numeric(factor(.order))]`|`[match(.order, sort(unique(.order)))]` is doing is correcting the `.order` indices to match what the API returns. So if we request rows `c(12, 12, 1, 2, 3, 4, 4, 5, 5, 5, 6, 6, 6, NA)` we will order the responses we get back as `c(7, 7, 1, 2, 3, 4, 4, 5, 5, 5, 6, 6, 6, NA)`

We might consider adding this on to the `[` method of variables so that the r-client always overrides the API behavior, and returns what an R user would expect in those contexts.

R/expressions.R
jonkeane: This is the same as the discussion above: the API returns selected rows in the order they appear in Crunch always, so we were always getting out of order results if the row numbers weren't ranges (eg if we `ds[c(12, 10, 3)]` we would get back `ds[c(3, 10, 12)]`.

R/as-data-frame.R
jonkeane: Absolutely, I was planning on looking into making the current promises closures and having the `$`, `[`, and `[[` methods evaluate them on the fly, but I think for the purposes of keeping this PR simply huge rather than massive, I'm going to defer those if possible.

R/as-data-frame.R
```diff
@@ -1,34 +1,62 @@
...
+                list(v=.a)))
         }
     })
+    
     class(out) <- "CrunchDataFrame"
     return(out)
 }
 
 setOldClass("CrunchDataFrame")
+setClass(Class = "CrunchDataFrame", contains = "data.frame")
```
nealrichardson: Is this necessary? CrunchDataFrame is S3, and this makes it S4.

R/as-data-frame.R
```diff
@@ -39,19 +67,20 @@ names.CrunchDataFrame <- function (x) names(x$.crunchDataset)
...
 #' @name dataset-to-R
 NULL
 
 #' @rdname dataset-to-R
 #' @export
 as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
-                                        force=FALSE, ...) {
-    out <- CrunchDataFrame(x)
+                                        force=FALSE, selection = NULL, ...) {
+    out <- CrunchDataFrame(x, selection = selection)
```
nealrichardson: maybe `row.order` to match `row.names` in the generic?

R/as-data-frame.R
```diff
@@ -73,3 +102,96 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
     names(out) <- names(x)
     return(structure(out, class="data.frame", row.names=c(NA, -nrow(x))))
 }
+
+
+#' Merge a CrunchDataFrame
+#' 
+#' Merging a CrunchDataFrame with a local dataframe is experiemental and might
```
nealrichardson: This disclaimer should go at the bottom, not at the top.

R/as-data-frame.R
```diff
@@ -73,3 +102,96 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' Merge a CrunchDataFrame
+#' 
+#' Merging a CrunchDataFrame with a local dataframe is experiemental and might 
+#' result in unexpected results. One known issue is that using `merge` on a 
+#' CrunchDataFrame will change the both the CrunchDataFrame used as input as 
+#' well as create a new CrunchDataFrame. 
+#' 
+#' `merge`ing a CrunchDataFrame with a local dataframe is useful in situations 
+#' where you have new information in your local R session that you want to 
+#' connect with Crunch data. For example, this is especially useful for making
```
nealrichardson: repetition of "useful"

R/as-data-frame.R
```diff
@@ -73,3 +102,96 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' plots with crunch and non-crunch data. It produces a hybrid CrunchDataFrame
```
nealrichardson: Crunch is capitalized

R/as-data-frame.R
```diff
@@ -73,3 +102,96 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' that has the local data attached to it, but like normal CrunchDataFrames 
+#' it is still judicious about downloading data from the server only when it 
+#' is needed.
+#' 
+#' @param x a CrunchDataFrame
+#' @param y a standard data.frame
+#' @param by.x name of the variable to match
+#' @param by.y name of the variable to match
```
nealrichardson: Doesn't merge have a "by" argument for when by.x and by.y are the same? cf. https://github.com/Crunch-io/rcrunch/blob/master/R/join-dataset.R#L62

R/as-data-frame.R
```diff
@@ -73,3 +102,96 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' @param ... ignored for now
+#' 
+#' @return a CrunchDataFrame with columns from both `x` and `y`
+#' 
+#' @export
+merge.CrunchDataFrame  <- function (x, y, by.x, by.y, sort = "x", ...) {
+    if (missing(by.x) | missing(by.y)) {
+        halt("Must supply both a by.x and a by.y to match by.")
+    }
+    if (!sort %in% c("x", "y")) {
```
nealrichardson: make `sort = c("x", "y")` in the signature and then do `sort <- match.arg(sort)` here. Gives you free documentation and validation.

R/geo.R
```diff
@@ -0,0 +1,217 @@
...
+#' @aliases geo geo<- CrunchGeography
+NULL
+
+#' @rdname geo
+#' @export
+setMethod("geo", "CrunchVariable", function (x) {
+    var_geodata <- entity(x)@body$view$geodata[[1]]
+    if (is.null(var_geodata)) {
+        # if there's no geodata, return null.
+        return()
```
nealrichardson: For explicitness, can you actually `return(NULL)`?

R/geo.R
```diff
@@ -0,0 +1,217 @@
...
+        match_field = var_geodata$match_field
+    )
+
+    return(geo_object)
+})
+#' @rdname geo
+#' @export
+setMethod("geo<-", c("CrunchVariable", "CrunchGeography"),
+          function (x, value) {
+              # if geodatum is of class CrunchGeodata, extact url
```
nealrichardson: Why would this be the case? Sounds like the object is weakly typed.

R/geo.R
```diff
@@ -0,0 +1,217 @@
...
+              if (is.Geodata(value$geodatum)) {
+                  value$geodatum <- self(value$geodatum)
+              } else {
+                  value$geodatum <- value$geodatum
+              }
+
+              geodata <- list(geodata = list(value))
+
+              dropCache(cubeURL(x))
```
nealrichardson: Doesn't matter practically, but I think it makes more sense to dropCache after you make the request rather than before.

R/geo.R
```diff
@@ -0,0 +1,217 @@
...
+              ent <- setEntitySlot(entity(x), "view", geodata)
+              return(x)
+          })
+
+#' Add geodata metadata to a crunch variable
+#' 
+#' @param variable a character with the variable alias to use for matching.
```
nealrichardson: Why is `variable` not actually a variable?

R/geo.R
```diff
@@ -0,0 +1,217 @@
...
+#' 
+#' @param features a vector of features to match (usually from a subset of the
+#' output `[availableFeatures]`) with a single property for a single geodatum.
+#' @param categories a vector of categories to match
+#' 
+#' @return the Jaccard index for the values of the property given in 
+#' feat_df and the vector of categories
+#' 
+#' @export
+scoreCat2Feat <- function (features, categories) {
```
nealrichardson: The norm in the package currently is to use "To" instead of "2" in functions that convert (like `toVariable`, `catalogToDataFrame`, etc.)

R/as-data-frame.R
jonkeane: I needed to do this in order to replace the data.frame in the `Spatial[X]DataFrame` objects. If `CrunchDataFrame` doesn't explicitly inherit `data.frame` that assignment errors. Another option would be to create `CrunchSpatial[X]DataFrame` classes that accept `CrunchDataFrame`s I'm not sure if us extending each of those or us extending `data.frame` is likely to have worse side effects.

R/geo.R
jonkeane: The way I've designed it the class `CrunchGeography` is weakly typed, it should have an object of `GeoData` or a string that is the url to that geodatum in the `geodatum` slot. I don't (yet) have any checks to ensure that's the case, but that's what this is was for. This is helpful for manually specifying geographies (see below) since the url to the geodatum is sufficient for creating the geography view.
```
states_geography <- list(geodatum='https://app.crunch.io/api/geodata/8f9f5fed101042c4815d2dd1fd248cec/',
                         feature_key='properties.name', # parse and list fields?
                         match_field='name') # list possible fields?
states_geography <- new('CrunchGeography', states_geography)

geo(so_data_combined$Country) <- states_geography
```
My plan is to add on an initialization method for `CrunchGeography` that takes the url and makes the appropriate `GeoData` object from it, and then we can remove this totally.

R/geo.R
jonkeane: My plan was to allow either a string as variable alias or the variable itself. Though on reflection and thinking about the rest of the package, we probably just want this to be the variable itself (since that's the much more common (only?) other pattern of interaction), and we can proceed that way.

R/as-data-frame.R
nealrichardson: Just seems wrong to define `CrunchDataFrame` as inheriting from environment in an S3 way and then re-defining it to inherit from `data.frame` as S4. I can't see how that would lead to good things, and it sure is confusing. 

Can you point me to the code for `SpatialDataFrame`? I'm not familiar with it.

tests/testthat/test-as-data-frame.R
```diff
@@ -122,6 +122,140 @@ with_mock_crunch({
         expect_identical(.crunchPageSize(ds$starttime), 100000L)
         expect_identical(.crunchPageSize(2016 - ds$birthyr), 50000L)
     })
+    
+    test_that("can manipulate the order of a crunchDataFrame", {
+        ds_df <- as.data.frame(ds)
+        gndr <- ds_df$v1
+        new_order <- c(4,3,2,1)
```
nealrichardson: Perhaps shuffle this so it's not just reversed, just to make it clear that it takes this order, whatever it is.

tests/testthat/test-as-data-frame.R
```diff
@@ -122,6 +122,140 @@ with_mock_crunch({
...
+                                         local_df,
+                                         by.x = "gender",
+                                         by.y = "gender"))
+        expect_identical(nrow(merged_df), nrow(ds_df))
+        expect_identical(merged_df$new,
+                         factor(c("new", "new", NA, "new", "new", "new", "new",
+                                  NA, NA, "new", "new", "new", "new", NA, NA,
+                                  NA, "new", "new", "new", NA, "new", "new",
+                                  "new", NA, "new")))
+        expect_error(merged_df <- merge(ds_df,
```
nealrichardson: I personally like to factor out the input validation tests to their own `test_that`. And keep in mind that the more descriptive the label in the `test_that`, the easier it is to read test output and understand what failed without having to look up line numbers etc.

tests/testthat/test-as-data-frame.R
```diff
@@ -122,6 +122,140 @@ with_mock_crunch({
...
+                                        sort = "not_an_input"),
+                     "Must supply both a by.x and a by.y to match by.")
+        expect_error(merged_df <- merge(ds_df,
+                                        data.frame(),
+                                        by.y = "gender",
+                                        sort = "not_an_input"),
+                     "Must supply both a by.x and a by.y to match by.")
+    })
+ 
+    test_that("Can't assign too many rows into a crunchdataframe", {
```
nealrichardson: CrunchDataFrame

R/as-data-frame.R
jonkeane: Sure, it comes from `sp` here is the code for `SpatialPolygonsDataFrame` https://github.com/edzer/sp/blob/master/R/Class-SpatialPolygonsDataFrame.R there are numerous others as well: https://github.com/edzer/sp/tree/master/R We likely will only use polygons, but there's no reason we might not in the future use points or other shapes, so I'm trying to keep that option open.

R/as-data-frame.R
jonkeane: I tried for a while to unify the functions used for joining datasets, but in the end I found that having the helper function locate the by variables returning the values of the variable was working against me more than it was helping. I've since abstracted a bit, and put the checking and processing the `by`s for either `data.frame`s or `CrunchDataFrame`s into a separate function similar to how join works, but separate to not impact the join logic. Let me know if this is a big problem, and I can see if there is a way to unify them (possibly involving changing the way join works and what it expects from that helper function.)

tests/testthat/test-as-data-frame.R
```diff
@@ -122,6 +122,158 @@ with_mock_crunch({
         expect_identical(.crunchPageSize(ds$starttime), 100000L)
         expect_identical(.crunchPageSize(2016 - ds$birthyr), 50000L)
     })
+    
+    test_that("can manipulate the order of a crunchDataFrame", {
```
nealrichardson: "row order"

tests/testthat/test-as-data-frame.R
```diff
@@ -122,6 +122,158 @@ with_mock_crunch({
...
+        ds_df <- as.data.frame(ds)
+        gndr <- ds_df$v1
+        new_order <- c(4,3,1,2)
```
nealrichardson: Note that this is not just a permutation but also a subset of rows

tests/testthat/test-geo.R
```diff
@@ -0,0 +1,89 @@
...
+    ds <- loadDataset("test ds")
+    geo_data <- geo(ds$location)
+    test_that("geo getter", {
+        expect_equal(geo_data$feature_key, "properties.location")
+        expect_equal(geo_data$match_field, "name")
+        expect_equal(geo_data$geodatum, "https://app.crunch.io/api/geodata/8684c65ff11c4cc3b945c0cf1c9b2a7f/")
+    })
+    
+    test_that("geo error", {
+        # if there is no geography on a variable, geo() returns null.
```
nealrichardson: Make this and the other comments like this be the message in `test_that` so that they show up in the test output

PR #65, opened 2017-07-05
Conditional transformation

jonkeane:


R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+#'
+#' `conditionalTransform` is similar to `makeCaseVariable` however `conditionalTransform` can use other crunch variables as a source of a variable, wehreas, whereas `makeCaseVariable` can only use characters. This additional power comes at a cost: `makeCaseVariable` can be executed entirely on Crunch servers, so no data needs to be downloaded or uploaded to/from the local R session. `conditionalTransform` on the other hand will download the data necesary to construct the new variable.
+#'
+#' @param ... a list of cases to evaluate
+#' @param default_value a default value to use if none of the conditions are true (default: `NA`)
```
nealrichardson: For consistency with `makeCaseVariable`, this should be called "else" right?

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+#' @param type a character that is either "categorical", "text", "numeric" what type of output should be returned?
```
nealrichardson: 1. Why is this necessary? How do the cases specified not already tell you what type the output will be?
2. As is, this function doesn't return a Crunch object anyway (it's an R vector and not a VarDef), so a type specification isn't quite accurate anyway.

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+#' @param categories if `type="categorical"`, these are all of the categories that should be in the resulting variable
```
nealrichardson: Why is this argument necessary?

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+#'
+#' @return either a factor (if `type="categorical"`) or character (if `type="text"`)
```
nealrichardson: If the point of this function is to derive a new variable on Crunch, but we don't yet have the means to derive it on the server so we're pulling data, perhaps it should return a VarDef like any other derivation. So, `VariableDefinition(result, ...)` or similar. Which means you'll need the same dance as `makeCaseVariable` to separate the formulas from the variable metadata from `...`.

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
+#' Conditional transformation
+#'
+#' Create a new variable that has values when specific conditions are met. Conditions are specified using a series of formulas: the righthand side is the condition that must be true (a `CrunchLogicalExpr`) and the lefthand side is where to get the value if the condition on the righthand side is true. This is commonly a crunch variable. Imagine that we have two sets of questions, one about what pets one has (Pet1, Pet2, Pet3) and then a second that asks about ones opinion of those pets that one gave in the Pet* series of questions (Opinion1, Opinion2, opinion3). What we ultimately want is what people's opinion of Cats or Dogs is, but because the value of Opinion1, Opinion2, and Opinion3 is dependent on the answer to Pet1, Pet2, and Pet3 respectively, we can't easily get the overall opinions about Cats. `conditionalTransform`  allows you to specify conditions like the following: if variable Pet1 is equal to 'Cat' then use the value from variable Opinion1, if variable Pet2 is equal to 'Cat' then use the value from variable Opinion2, etc. See examples below for how to write this.
```
nealrichardson: Line break please

nealrichardson: Other prose comments:

* Crunch should be capitalized
* Give it a spell-check
* Perhaps save the "Imagine" discussion for the examples, and add more to the examples than just creating the variables (i.e. derive the variable, then show what it made)

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+    if (!type %in% c("categorical", "text", "numeric")){
+        halt("type must be either ", dQuote("categorical"), dQuote("text"), " or ", dQuote("numeric"))
+    }
+    if (type != "categorical" & !is.null(categories)){
+        halt("type is not ", dQuote("categorical"), " ignoring ", dQuote("categories"))
+    }
+
+    cases <- vector("list", n)
+    values <- vector("list", n)
+    for (i in seq_len(n)) {
```
nealrichardson: Why are the existing formula evaluation functions not adequate here? Perhaps those should be refactored so that all of the formula evaluation is consolidated.

Also, looking at this reminds me: perhaps this function, and for that matter, `makeCaseVariable` too, should take a `data` argument in which the formula should be evaluated. `function(formula, data)` is the usual way of dealing with formulas. That way the `$` can be dropped from all of those case definitions.

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+            # if var isn't a crunchvariable, just return var
+            return(var)
+        }
+    }, ind = case_indices, var = values)
+
+    # setup NAs for as default
+    result <- rep(default_value, n_rows)
+
+    for (i in seq_along(case_indices)) {
+        if (type == "numeric") trans <- as.numeric
```
nealrichardson: (1) I don't think this 'trans' is necessary; R type coercion would just handle it fine.

```
> v <- 1:5
> v[3] <- "a"
> v
[1] "1" "2" "a" "4" "5"
```

(2) Please don't if... else without curly braces

tests/testthat/test-conditional-transform.R
```diff
@@ -0,0 +1,58 @@
...
+    test_that("conditionalTransform", {
+        expect_error(conditionalTransform(ds$gender),
+                     'The case provided is not a formula: ds\\$gender')
+        expect_error(conditionalTransform("bar"~"foo"),
+                     'The LHS provided is not a CrunchLogicalExpr: "bar"')
+        expect_error(conditionalTransform(ds$gender~"foo", type="unknown"),
+                     "type must be either ", dQuote("categorical"), " or ", dQuote("text"))
+        expect_error(conditionalTransform(ds$gender~"foo", type="text", categories=c("foo", "bar")),
+                     "type is not ", dQuote("categorical"), " ignoring ", dQuote("categories"))
+        # TODO: Figure out why the mock crunch expression response is not working
```
nealrichardson: Could you elaborate on "not working"?

R/conditional-transform.R
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    # check categories, return factor type="categorical"
+    if (type == "categorical") {
+        uni_results <- unique(result)
+        results_not_categories <- !uni_results %in% categories
+        # if categories are supplied and there are any
+        if (is.null(categories)) {
+            result <- factor(result)
+        } else {
```
nealrichardson: If specifying categories is just about validation, then I'm more against having it as an argument at all. If we're returning a VarDef and we want to let the user specify a particular category order or something, then I could see some value, perhaps. 

Given the current code, you could move L85-86 inside the "else" and only evaluate them if the validation is happening.

R/variable-to-R.R
```diff
@@ -53,6 +53,11 @@ parse_column <- list(
     datetime=function (col, variable, mode) {
         out <- columnParser("text")(col)
         return(from8601(out))
+    },
+    boolean=function (col, variable, mode) {
```
jonkeane: I've made the test on line 172 of test-expressions a bit more explicit (reference the dataframe that created the dataset, and ensure those are equal instead of just grabbing the vector). Is that what you're looking for here, or something else?

inst/app.crunch.io/api/datasets/1/table-ec73c0.json
```diff
@@ -0,0 +1,2 @@
+{"self":
+   "http%3A%2F%2Flocal.crunch.io%3A8080%2Fapi%2Fdatasets%2F1%2Ftable%2F%3Fquery%3D%7B%22out%22%3A%7B%22function%22%3A%22%3D%3D%22%2C%22args%22%3A%5B%7B%22variable%22%3A%22http%3A%2F%2Flocal.crunch.io%3A8080%2Fapi%2Fdatasets%2F1%2Fvariables%2Fgender%2F%22%7D%2C%7B%22value%22%3A1%7D%5D%7D%7D%26filter%3D%7B%7D%26offset%3D0%26limit%3D50000", "element": "crunch:table", "data": {"out": [false, false, false, false, false, false, true, false, false, false, true, false, true, false, false, false, false, false, false, false, true, true, true, false, true]}, "description": "A Crunch Table of data for this dataset.", "metadata": {"out": {"alias": null, "type": "boolean", "name": null, "missing_reasons": {}, "description": ""}}}
```
jonkeane: The server's response is fine, this was me overly aggressively unescaping/reescaping.

R/conditional-transform.R
jonkeane: The first reason was that I was trying to match the feature request. The second was, with the transformations explicit it lets users specify different variable types to use as sources for the transform the use case I imagine most frequently wanting to use both a categorical and a text variable for the source of one of these conditionals, and then being able to specify that it should result in either a text or categorical. This might be over-engineering, but that was what I was aiming for.

R/conditional-transform.R
jonkeane: (1) the issue I ran into is when we were trying to store elements from factors, I needed to `as.character` them in order to get the level instead of the level id:
```
> foo <- factor(c("A", "B"))
> bar <- c("c","d")
> baz <- rep(NA, 2)
> baz[1] <- foo[1]
> baz[2] <- bar[2]
> baz
[1] "1" "d"
> baz <- rep(NA, 2)
> baz[1] <- as.character(foo[1])
> baz[2] <- bar[2]
> baz
[1] "A" "d"
```
We could just limit this transformation to categoricals that get used, and use coercion otherwise. Although if we keep the type specifications we should probably at least check the types at the end if not converting them specifically.

(2) done

R/conditional-transform.R
jonkeane: I've changed this to `else_condition` as `else` is reserved.

R/conditional-transform.R
rdrivers: We want this to create a derived variable in Crunch, not an R vector. The type could be categorical, numeric, or text, I think.

tests/testthat/test-conditional-transform.R
jonkeane: I was able to fix this with refactoring evaluation.

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
+#' Conditional transformation
+#'
+#' Create a new variable that has values when specific conditions are met. 
+#' Conditions are specified using a series of formulas: the righthand side is 
+#' the condition that must be true (a `CrunchLogicalExpr`) and the lefthand 
+#' side is where to get the value if the condition on the righthand side is 
+#' true. This is commonly a Crunch variable.
```
nealrichardson: ... but may be a string or numeric value, depending on the type of variable you're constructing.

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#'
+#' `conditionalTransform` is similar to `makeCaseVariable` however
```
nealrichardson: ; however,

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#'
+#' Create a new variable that has values when specific conditions are met. 
+#' Conditions are specified using a series of formulas: the righthand side is 
+#' the condition that must be true (a `CrunchLogicalExpr`) and the lefthand 
+#' side is where to get the value if the condition on the righthand side is 
+#' true. This is commonly a Crunch variable.
+#'
+#' `conditionalTransform` is similar to `makeCaseVariable` however 
+#' `conditionalTransform` can use other Crunch variables as a source of a 
+#' variable, whereas, whereas `makeCaseVariable` can only use characters. This
```
nealrichardson: whereas whereas

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#' # dependent on the answer to Pet1, Pet2, and Pet3 respectively, we can't 
+#' # easily get the overall opinions about Cats. `conditionalTransform` allows 
+#' # you to specify conditions like the following: if variable Pet1 is equal to
+#' # 'Cat' then use the value from variable Opinion1, if variable Pet2 is 
+#' # equal to 'Cat' then use the value from variable Opinion2, etc.
+#' 
+#' ds$cat_opinion <- conditionalTransform(pet1 == 'Cat' ~ Opinion1,
+#'                                        pet2 == 'Cat' ~ Opinion2,
+#'                                        pet3 == 'Cat' ~ Opinion3, 
+#'                                        data = ds)
```
nealrichardson: Include variable metadata here, like `name="Opinion of Cats"`

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#' as.vector(ds$cat_opinion)
```
nealrichardson: Rather than as.vector, just show (print) the variable.

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#'
+#' @param ... a list of cases to evaluate as well as other
+#' properties to pass about the case variable (i.e. alias, description)
+#' @param data a Crunch dataset object to use
+#' @param else_condition a default value to use if none of the conditions are 
+#' true (default: `NA`)
+#' @param type a character that is either "categorical", "text", "numeric" what
+#'  type of output should be returned? The source variables will be converted 
+#'  to this type if necessary
+#' @param categories if `type="categorical"`, these are all of the categories
```
nealrichardson: What type of object is `categories`? In the example below, it appears to be character (category names, I presume), but that isn't the same thing as `Categories`. Fine to accept character in order to specify order of categories, but if this argument is going to exist, it should (also) accept full category definitions so that you can specify more category metadata (like missingness, numeric values, etc.)

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#'                                        pet2 == 'Dog' ~ Opinion2,
+#'                                        pet3 == 'Dog' ~ Opinion3,
+#'                                        data = ds)
+#' as.vector(ds$dog_opinion)
+#' # [1] "Strongly Agree" "Disagree" "too early to tell" "Strongly Disagree"
+#' # Levels: Agree  Disagree  Strongly Agree  Strongly Disagree  too early to tell
+#'                                     
+#' # Further, we can also use conditional transform with numeric source variables.
+#' 
+#' ds$days_with_dog <- conditionalTransform(pet1 == 'Dog' ~ days_having_pet,
```
nealrichardson: These are all drawing from the same `days_having_pet` variable

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#'                                          pet2 == 'Dog' ~ days_having_pet,
+#'                                          pet3 == 'Dog' ~ days_having_pet,
+#'                                          data = ds, type = "numeric")
+#' as.vector(ds$days_with_dog)
```
nealrichardson: Yeah, in all of these, use some method other than `as.vector` to show what was created. We're trying to encourage our users not to pull data.

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#' # [1] 300 57 70 5 ...
+#' 
+#' # conditionalTransform will evaluate is.na() or NA as conditoins or sources
```
nealrichardson: Spelling, and I would make this the more general point that your conditions can be more complex expressions, anything that is "logical"

R/conditional-transform.R
```diff
@@ -0,0 +1,215 @@
...
+#'                                        pet2 == 'Cat' ~ Opinion2,
+#'                                        pet3 == 'Cat' ~ Opinion3, 
+#'                                        is.na(pet1) & is.na(pet2) & is.na(pet3) ~ "None",
+#'                                        data = ds)
+#' as.vector(ds$cat_opinion)
+#' # [1] "Strongly Agree" "Disagree" "Agree" "Strongly Disagree" "None" ...
+#' # Levels: Agree  Disagree None Strongly Agree  Strongly Disagree
+#' 
+#' 
+#' # Finally, when constructing conditionalTransforms, it is frequently helpful
```
nealrichardson: This is reading to me more like a vignette than a man page. That's fine, it's easy to link a man page to a vignette.

R/conditional-transform.R
```diff
@@ -0,0 +1,150 @@
...
+    if (n == 0) {
+        halt("no conditions have been supplied; please supply formulas as conditions.")
+    }
+    if (!type %in% c("categorical", "text", "numeric")){
+        halt("type must be either ", dQuote("categorical"), ", ", dQuote("text"), ", or ", dQuote("numeric"))
+    }
+    if (type != "categorical" & !is.null(categories)){
+        halt("type is not ", dQuote("categorical"), " ignoring ", dQuote("categories"))
+    }
+
```
nealrichardson: IMO this would be clearer to read and grok if here you assigned `var_def$values <- computeStuff(formulas)` and push L60-105 into 

```
computeStuff <- function (formulas, data) {
    ...
    return(result)
}
```

(obviously, with a proper function name)

R/conditional-transform.R
```diff
@@ -0,0 +1,150 @@
...
+    }
+ 
+    if (type == "categorical") {
+        # if categories are supplied and there are any
+        if (missing(categories)) {
+            result <- factor(result)
+            # if categories aren't a Categories object,
+            # make categories from names
+            categories <- Categories(data = categoriesFromLevels(levels(result)))
+            categories <- c(categories, Category(data=.no.data))
```
nealrichardson: I'd factor out this "No data" category adding since you're doing it in every case. And look to how you can encapsulate this logic better in the few other existing parts of the code where this behavior happens (search for `.no.data`).

inst/app.crunch.io/api/datasets/1/variables/textVar/values-55e7eb.json
```diff
@@ -0,0 +1,6 @@
+// Query: filter=%7B%22function%22%3A%22in%22%2C%22args%22%3A%5B%7B%22function%22%3A%22row%22%2C%22args%22%3A%5B%5D%7D%2C%7B%22column%22%3A%5B6%2C10%2C12%2C20%2C21%2C22%2C24%5D%7D%5D%7D&offset=0&limit=5000
```
nealrichardson: Good to put the query here in comment, but it's more readable if you URL decode it

R/case-variables.R
```diff
@@ -100,6 +112,14 @@ is_else_case <- function (case) {
     identical(case[['expression']], magic_else_string)
 }
 
+evalWithData <- function (to_eval, dat = NULL, eval_env) {
```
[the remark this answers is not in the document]
jonkeane: Good catch!

R/conditional-transform.R
```diff
@@ -0,0 +1,150 @@
...
+    }
+    if (type != "categorical" & !is.null(categories)){
+        halt("type is not ", dQuote("categorical"), " ignoring ", dQuote("categories"))
+    }
+    
+    cases <- vector("list", n)
+    values <- vector("list", n)
+    for (i in seq_len(n)) {
+        formula <- formulas[[i]]
+        if (!inherits(formula, "formula") || length(formula) != 3) {
```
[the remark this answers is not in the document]
jonkeane: Yup, this was a holdover from a previous way of testing I've removed the `!inherits(formula, "formula") || ` bit. I don't think there's a situation where `length(formula) != 3` but because we are referencing the formula by indices, I want to keep it just in case there is some weird edge case formula we aren't expecting.

R/conditional-transform.R
```diff
@@ -0,0 +1,150 @@
...
+            # grab the variable contents at inds
+            return(as.vector(var[ind]))
+        } else {
+            # if var isn't a crunch variable or expression, just return var
+            return(var)
+        }
+    }, ind = case_indices, var = values)
+    
+    # setup NAs for as default
+    n_rows <- nrow(CrunchDataset(crGET(datasetReference(cases[[1]]))))
```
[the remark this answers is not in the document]
jonkeane: We could do this if we require that people use the data argument, but I was trying to allow users to be able to specify fully specified columns if they want (eg `conditionalTransform(ds$v3 < 1 ~ 'foo', ds$v4 == "B" ~ 'bar', name='new')`).

Thought thinking about this again, we should at least check that data is there, and try to grab it from there (to save requests, but also help avoid relying on the first case...) I've added a bit of logic to do the right thing.

R/conditional-transform.R
```diff
@@ -0,0 +1,150 @@
...
+                     ") that were not specified in categories")
+            }
+            result <- factor(result, levels = names(categories))
+        }
+    }
+
+    var_def$type <- type
+    if (type == "categorical") {
+        # make a category list to send with VariableDefinition and then store that and convert values to ids values
+        category_list <- listifyCategories(categories)
```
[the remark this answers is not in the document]
jonkeane: I needed someway to go from a `Categories` object to a plain list that would then get parsed into json for the variable definition. I tried a bunch of different things, but I didn't find anything already in the package that would do this. It might be better to have this as the `as.list` method (I didn't because `Categories` objects already inherit from lists through s4 inheritance magic and that seems a bit strange to do).

R/conditional-transform.R
jonkeane: Absolutely, I've added an `ensureNoDataCategory()` function that adds the No Data category if there's not one already. I've put it in the few places it would fit elsewhere in the package as well.

R/conditional-transform.R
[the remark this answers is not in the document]
nealrichardson: `Categories` already have a JSON serializing method (https://github.com/Crunch-io/rcrunch/blob/master/R/json.R), so you don't need to do anything special to make them turn into JSON correctly. (Note that the `categories<-` method just PATCHes with the `Categories` object.)

R/conditional-transform.R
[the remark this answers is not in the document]
nealrichardson: Due to caching, you won't actually save requests by checking if `data` is defined.

R/conditional-transform.R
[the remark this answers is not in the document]
jonkeane: It turns out that calling `toJSON()` on a `VariableDefinition` doesn't trigger the list-based `lapplying` of `jsonprep()`, which was leading to errors. I've added that, and now can fully do away with this listifying.

tests/testthat/test-conditional-transform.R
```diff
@@ -0,0 +1,199 @@
...
+                     paste0("There is more than one dataset referenced. Please ",
+                            "supply only one."))
+    })
+
+    test_that("conditionalTransform works with categories", {
+        expect_silent(new_var <- conditionalTransform(gender == "Male" ~ textVar, data = ds))
+        expect_equal(new_var$values, c(-1, -1, -1, -1, -1, -1, 2, -1, -1, -1,
+                                       3, -1, 4, -1, -1, -1, -1, -1, -1, -1, 1,
+                                       6, 3, -1, 5))
+        expect_equal(new_var$type, "categorical")
```
nealrichardson: This seems odd to me. Your expression says "i want a variable where the values are `textVar` where gender is Male", but your resulting variable is categorical and not text?

tests/testthat/test-case-variables.R
```diff
@@ -35,8 +35,8 @@ with_mock_crunch({
                                           missing="nope")),
                      "a case's missing must be a logical")
         expect_error(ensureValidCase(list(not_right="not")),
-                     paste("each case must have at most an id, name, expression,",
-                     "numeric_value, and missing element. The errant",
+                     paste0("each case must have at most an id, name, expression, ",
```
nealrichardson: Why all of these `paste` to `paste0` changes?

R/json.R
```diff
@@ -8,6 +8,10 @@ setMethod("jsonprep", "list", function (x, ...) lapply(x, jsonprep, ...))
 
 #' @rdname tojson-crunch
 #' @export
+setMethod("jsonprep", "VariableDefinition", function (x, ...) lapply(x, jsonprep, ...))
```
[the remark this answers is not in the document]
jonkeane: I added this when we added the ability to specify categories with a `Categories` object. If we leave this as is, those categories don't (recursively) get turned into json. 

When creating a categorical variable from an R factor, we the way we set categories was actually bypassing those `Categories` objects all together https://github.com/Crunch-io/rcrunch/blob/master/R/R-to-variable.R#L89 (tangentially, I wonder if that should be changed to use `Categories` objects so we get some of that validation built in to that for free? I'll ticket accordingly if we do, so as not to scope creep this PR.

This is related to the Categories to list function that we decided against here: https://github.com/Crunch-io/rcrunch/pull/65#discussion_r138035809 But maybe I'm missing a better way to make sure that `Categories` inside of a variable definition gets json-prepped and isn't left alone?

tests/testthat/test-case-variables.R
jonkeane: I feel like we more frequently used `paste0` in tests (I just searched and it turns out it's almost exactly half and half). I'll remove the changes here.

jonkeane:
I've made most of these changes (a few I had comments replying to). Type coercion now (more or less) follows the way that R does, with the notable exception that factors have to be treated specially in order to get them into characters (and not just get ids as characters). I've added discussion to the man page that describes that behavior. 

In doing that, I was able to factor out and test separately the coercion process (addressing part of the second concern you had), I've also added a not in the tests to add more mocks for other variable's responses and test `makeConditionalValues()` more thoroughly, which hopefully will address the combinatorial unit testing concerns. But I think for now, deferring those to get this out and get back to nets might be a wise move (I can create an issue that explains what's going on so that future us can jump back in to it).

nealrichardson:
Can you please add a NEWS entry, and then let's merge this.

PR #68, opened 2017-07-13
Dichotomize derive array 148685747

jonkeane:
deriveArray() now can accept (and will use!) a selections argument. When selections is provided, a multiple response variable will be derived instead of a categorical array. Now with more CI goodness.

### DIFF
--- a/R/make-array.R
+++ b/R/make-array.R
@@ -5,13 +5,14 @@
 #' a Dataset)
 #' @param name character, the name that the new Categorical Array variable
 #' should have. Required.
-#' @param selections character, for \code{makeMR}, the names of the
+#' @param selections character, for `makeMR` and `deriveArray` the names of the
 #' categories to mark as the dichotomous selections. Required for
-#' \code{makeMR}; ignored in \code{makeArray}.
+#' `makeMR`; optional for `deriveArray`; ignored in `makeArray`.
 #' @param ... Optional additional attributes to set on the new variable.
 #' @return A VariableDefinition that when added to a Dataset will create the
-#' categorical-array or multiple-response variable. \code{deriveArray} will
-#' make a derived array expression, while \code{makeArray} and \code{makeMR}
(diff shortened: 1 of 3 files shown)


R/make-array.R
```diff
@@ -90,6 +91,16 @@ deriveArray <- function (subvariables, name, ...) {
...
         list(value=I(subvarids))))
 
+    if (!missing(selections)) {
+        # if there are selections, wrap the array function inside of a
+        # select_categories function
+        if(!is.list(selections)){
+            # if selections is not already a list convert it to one.
+            selections <- as.list(selections)
+        }
+        derivation <- zfunc("select_categories", derivation, list(value=selections))
```
nealrichardson: I don't think the as.list is necessary, nor the `list` here. IIRC `zfunc("select_categories", derivation, value=I(selections))` is what you want, and then you could drop the if is.list check.

tests/testthat/test-derive-array.R
```diff
@@ -29,6 +29,16 @@ with_test_authentication({
         expect_identical(as.vector(ds$derivedarray[[1]]), q1.values)
     })
 
+    mr_derivation <- deriveArray(list(ds$q1, ds$petloc$petloc_home), selections=c("Dog"), name="Derived pets MR") # cat dog has id 2
```
nealrichardson: `identical(c("Dog"), "Dog")`

tests/testthat/test-derive-array.R
```diff
@@ -29,6 +29,16 @@ with_test_authentication({
...
+    ds$derivedmr <- mr_derivation
```
nealrichardson: If you're not going to assert anything about the shape of `mr_derivation`, just do the assignment into `ds` on the same line.

R/make-array.R
nealrichardson: oh right, forgot, we do need that `list`, ZCL likes extra nesting.

PR #81, opened 2017-07-14
Ordering

persephonet:
Added a few functions for ordering


nealrichardson:
Thanks @persephonet !

PR #82, opened 2017-07-18
Feature cleanup

jonkeane:
add fortification for ggplotting, and add a method to reorder categories by name


R/categories.R
```diff
@@ -66,6 +66,13 @@ setMethod("[", c("Categories", "ANY"), function (x, i, ...) {
 
 #' @rdname Categories
 #' @export
+setMethod("[", c("Categories", "character"), function (x, i, ...) {
+    indeces <- match(i, names(x))
```
nealrichardson: Spelling

tests/testthat/test-categories.R
```diff
@@ -34,6 +34,7 @@ with_mock_crunch({
 
     test_that("category slicers", {
         expect_true(is.categories(cats[1]))
+        expect_equal(cats[c("Female","Male")], cats[c(2,1)])
```
nealrichardson: Spaces.

Also, can you add some testing around the error handling (e.g. what happens if the strings you have don't match any category names)?

PR #85, opened 2017-07-18
Web url

nealrichardson:


R/dataset.R
```diff
@@ -329,12 +329,39 @@ setMethod("allVariables<-", c("CrunchDataset", "VariableCatalog"),
...
 
-webURL <- function (x) {
+APIToWebURL <- function (x) {
     ## URL to view this dataset in the web app
     stopifnot(is.dataset(x))
     return(paste0(absoluteURL("/", getOption("crunch.api")), "dataset/", id(x)))
 }
 
+webToAPIURL <- function (url) {
+    id <- sub("^https.*?/dataset/([0-9a-f]+)/.*$", "\\1", url)
```
jonkeane: Would `datasetReference` work here?

nealrichardson: It wouldn't as-is ("datasets" vs. "dataset", subtle difference in the whaam URL): https://github.com/Crunch-io/rcrunch/blob/master/R/variable.R#L169

Could make the regexp in that line match on either if you think that's cleaner.

jonkeane: Ah, yeah I think having `datasetReference` match either won't break anything, and it will keep our parsing all in one place. There's a bug I fixed in the conditional branch (https://github.com/Crunch-io/rcrunch/pull/65/files#diff-3647057bf9eebd954d358f8bc6e4beedR168) that bit Brian using logical expressions. 

It's not a big deal if we have them separate, I just remembered running into a problem parsing urls for ids earlier.

nealrichardson: Oh actually `datasetReference` returns a URL and I need the id here.

jonkeane: Ah right of course. I think we're good. If we run into it again / elsewhere I'll see about generalizing a bit, but for one line here I don't think it's worth holding it up.

PR #86, opened 2017-07-24
As.data.frame.variable catalog

### LINKED ISSUE #75: as.data.frame variables(ds)
When `variables(ds)` is printed to the screen, it prints the aliases, names, and types for all the variables. It would be great if that could be made into an R object. It is not much harder to do `data.frame(alias=aliases(variables(ds)), name=name(variables(ds)), type=type(variables(ds)))` but it would make my code much cleaner.

(Also, if `variables(ds)` also printed descriptions, that would be nice.)

GShotwell:
This closes #75  by adding an `as.data.frame()` method for variableCatalogs.


R/as-data-frame.R
```diff
@@ -73,3 +73,54 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' vars <- variables(ds)
+#' var_df <- as.data.frame(vars, fields = "all")
+#' }
+#' 
+#' @rdname VariableCatalog-to-Data-Frame
+#' @importFrom purrr map_df map_lgl
+#' @importFrom dplyr bind_cols
+#' @export
+as.data.frame.VariableCatalog <- function(varCat, 
+                                          fields = c("alias", "name", "type")) {
```
jonkeane: It looks like there was also a warning about the method signature consistency during `R CMD check`. It might be as simple as adding `...` as an argument here. We might also want or possibly need to change `varCat` to `x` to match the arguments of the `as.data.frame` generic.

R/as-data-frame.R
```diff
@@ -73,3 +73,54 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+  ## rectangular. 
+  vectors$description[map_lgl(vectors$description, is.null)] <-  "NULL"
+  vectors <- map_df(vectors, unlist)
+  out <- bind_cols(vectors, sub_variables)
+
+ if (all(fields == "all")) {
+   return(out)
+ }
+ if (!all(fields %in% names(out))) {
+   stop("Field name not present in variable catalog, use fields = 'all' to see available fields.")
```
jonkeane: `stop` was my first inclination as well, but we have a `halt` function from `httpcache` that we've been using. `halt` adds some logging as well as by default hides the call from users.

tests/testthat/test-variable-catalog.R
```diff
@@ -99,6 +98,16 @@ with_mock_crunch({
...
+    test_that("Variable Catagory as.data.frame method", {
+      expect_equal(as.data.frame(varcat[1:3]),
+                    data.frame(
+                      alias=c("birthyr", "gender", "mymrset"),
+                      name=c("Birth Year", "Gender", "mymrset"),
+                      type=c("numeric", "categorical", "multiple_response"),
+                      stringsAsFactors = FALSE
+                    ))
+         })
 })
```
jonkeane: It would be nice to have a test for the failure states if a user passes fields that aren't in the catalog, etc.

R/as-data-frame.R
```diff
@@ -1,18 +1,18 @@
...
-    with(out, {
-        ## Note the difference from as.environment: wrapped in as.vector
-        for (a in aliases(allVariables(dataset))) {
-            eval(substitute(delayedAssign(v, as.vector(.crunchDataset[[v]])),
-                list(v=a)))
-        }
-    })
-    class(out) <- "CrunchDataFrame"
-    return(out)
+  ## S3 constructor method for CrunchDataFrame. terms.formula doesn't seem
```
nealrichardson: Please undo your editor's linting. In this package, tabs are 4 spaces.

DESCRIPTION
```diff
@@ -17,11 +17,14 @@ License: LGPL (>= 3)
 Depends:
     R (>= 3.0.0)
 Imports:
+    curl,
+    dplyr,
```
nealrichardson: Adding three new package dependencies for a single function is unnecessary. Please find a way to accomplish the data manipulations using base R methods.

R/as-data-frame.R
```diff
@@ -50,26 +50,72 @@ NULL
...
+#' }
+#' 
+#' @rdname VariableCatalog-to-Data-Frame
+#' @export
+as.data.frame.VariableCatalog <- function(x, 
+                                          row.names = NULL, 
+                                          optional = FALSE, 
+                                          ...,
+                                          fields = c("alias", "name", "type")) {
+  out <- catalogToDataFrame(x, rownames = NA)
```
nealrichardson: Why not just pass "fields" to `catalogToDataFrame` (the arg there is called "keys" but you could rename it to match)?

R/shoji-catalog.R
```diff
@@ -240,28 +240,61 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
 #' @return a \code{data.frame} view of the catalog
+#' @importFrom purrr map map_df map_lgl
+#' @importFrom tibble as_tibble
 #' @export
-catalogToDataFrame <- function (x, keys=TRUE, rownames, ...) {
+catalogToDataFrame <- function(x, keys=TRUE, 
+                                rownames = "default", 
+                                list_columns = c("subvariables", "subvariables_catalog"),
+                                ...) {
+    index <- map(index(x), function(a) a[keys])
```
nealrichardson: map -> lapply

R/shoji-catalog.R
```diff
@@ -240,28 +240,61 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+#' in a list-column. 
 #' @param ... additional arguments passed to \code{data.frame}
 #' @return a \code{data.frame} view of the catalog
+#' @importFrom purrr map map_df map_lgl
+#' @importFrom tibble as_tibble
 #' @export
-catalogToDataFrame <- function (x, keys=TRUE, rownames, ...) {
+catalogToDataFrame <- function(x, keys=TRUE, 
+                                rownames = "default", 
+                                list_columns = c("subvariables", "subvariables_catalog"),
```
nealrichardson: Could you infer `list_columns` from the data itself?

R/shoji-catalog.R
```diff
@@ -240,28 +240,61 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+                                ...) {
+    index <- map(index(x), function(a) a[keys])
+    
+    entry_to_df <- function(l, list_col_names = list_columns){
+      l[map_lgl(l, is.null)] <- NA
+      vect_col <- l[!(names(l) %in% list_col_names)]
+      entry_df <- tibble::as_tibble(vect_col)
```
nealrichardson: `as.data.frame(... stringsAsFactors=FALSE)`

tests/testthat/test-variable-catalog.R
```diff
@@ -99,12 +98,23 @@ with_mock_crunch({
...
+    test_that("Variable Catagory as.data.frame method", {
+      expect_equal(as.data.frame(varcat[1:3]),
+                    data.frame(
+                      alias = c("birthyr", "gender", "mymrset"),
+                      name  = c("Birth Year", "Gender", "mymrset"),
+                      type  = c("numeric", "categorical", "multiple_response"),
+                      stringsAsFactors = FALSE
+                    ))
+      expect_error(as.data.frame(varcat[1:3], fields = "Not a field at all"))
+         })
```
nealrichardson: Assert the error message

R/as-data-frame.R
GShotwell: Whoops, sorry for missing that.

R/as-data-frame.R
GShotwell: I guess it just seemed clearer to put all of that error management stuff in the as.data.frame method. It doesn't look like the keys subsetting fails when you try to pass a key which isn't present in the index, so when you ask for something like `index <- lapply(index(x), function(a) a["Banana"]) ` you get a list with no values, not an error. This makes sense because each element of the index will have different variable attributes. For instance if you ask for "subvariables" that is going to be a valid attribute of some variables, but not others, so the `[` method shouldn't fail if it gets asked to return something which isn't an attribute of that variable. We really only want to throw an error when the attribute isn't present in the whole set of variables, which means that we need to look at the dataframe with all of the attributes. That being the case it seemed simpler to just ask for the whole dataset, and then do the error checking in the `as.data.frame` method. 

The other reason is that it seems like `CatalogToDataframe` is supposed to be a general purpose function which works across all catalogs, and I don't know those use cases well enough to say whether asking for non-existent keys is something that should throw an error in those cases.

R/shoji-catalog.R
GShotwell: Just to reference this in case anyone is interested, Tibbles have a few features which make then more appropriate for programatic use than dataframes just because they are a bit stricter in their inputs and fail earlier.  https://cran.r-project.org/web/packages/tibble/vignettes/tibble.html

R/shoji-catalog.R
GShotwell: I couldn't figure out a good way of doing that. The issue is that if I do something like ask whether `length(x) == length(unlist(x)` then I would wrongly thing that things like subvariables with only one subvariable should be stored as vectors. It also seems like we should make clear decisions about how each attribute should be stored, so coding the attribute name makes sense in that way. That said if anyone can think of a way of reliably inferring the proper output type that might also be a good solution.

R/as-data-frame.R
```diff
@@ -73,3 +73,49 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' }
+#' 
+#' @rdname VariableCatalog-to-Data-Frame
+#' @export
+as.data.frame.VariableCatalog <- function(x, 
+    row.names = NULL, 
+    optional = FALSE, 
+    ...,
+    fields = c("alias", "name", "type")) {
+    out <- catalogToDataFrame(x, rownames = NA)
```
nealrichardson: Remind me why "fields" aren't just passed to `catalogToDataFrame`? I think we discussed this before.

R/as-data-frame.R
```diff
@@ -73,3 +73,49 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+    out <- as.data.frame(out, ...)
+    if (all(fields == "all")) {
+        return(out)
+    }
+    if (!all(fields %in% names(out))) {
+        halt("Field name not present in variable catalog, use fields = 'all' to see available fields.")
```
nealrichardson: A better error message would tell you which fields are not valid; this could also be done from inside `catalogToDataFrame`

R/as-data-frame.R
GShotwell: Let me remind myself first.... Oh I think it's that each variable catalog has different attributes, and we only want to throw an error if the field is not present in any of the variables in the dataset. So it seemed better to get all the available fields for each variable, and then check to see if the field names are valid.

R/as-data-frame.R
```diff
@@ -73,3 +73,49 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' 
+#' @rdname VariableCatalog-to-Data-Frame
+#' @export
+as.data.frame.VariableCatalog <- function(x, 
+    row.names = NULL, 
+    optional = FALSE, 
+    ...,
+    fields = c("alias", "name", "type")) {
+    out <- catalogToDataFrame(x, rownames = NA)
+    out <- as.data.frame(out, ...)
```
GShotwell: I think I had this there in order to get around a command check of not having the dots do anything, but I see I can just pass them directly through to `catalogToDataFrame()` which is better.

R/as-data-frame.R
nealrichardson: You could do that in `catalogToDataFrame` too though. Advantages of doing it lower are (1) it would just work for all of our other `*Catalog` objects, and (2) to the extent we care, you'd probably get a performance boost by constructing a much narrower data.frame rather than making the biggest one possible and then paring it back.

GShotwell: I made some changes to the error message can't figure out how to replicate the current behaviour at the lower level. Currently the error message does two things:
1) Checks whether the field names are present in the dataframe; and
2) Suggests fields in the dataframe which aren't present in the requested fields list. 

So if someone asks for "banana" it will say something like "there is no field called 'banana', did you mean 'alias' ..." Now I can replicate the first bit at the `catalogToDataframe()` level by passing the filed name as the `keys` argument, and then checking the resulting data against the requested fields. I don't think I'll be able to provide guidance on available fields without effectively doing what I'm doing now. 

More generally, are we sure we want the same error behaviour for all current and future catalog types? I could envision a scenario where you might want to ask for fields which aren't present in the catalog and not expect an error, so from that perspective it might make more sense to have the `catalogToDataframe()` function just handle pulling the data into a data frame, and then have the `as.data.frame()` methods handle the error checking.

nealrichardson: I want a reasonable behavior that works for all Catalogs so that we don't have to reimplement the method for each Catalog unless we want to specialize it. 

See https://github.com/Crunch-io/rcrunch/blob/master/R/categories.R#L197-L201 for a model of how this kind of validation is handled elsewhere (grep for `serialPaste` if you want to see lots more). You can put something like this in `catalogToDataFrame`, have the error message include the `class()` of the object passed in, and if you want to list all of the names of the valid fields, feel free. So the error message I am picturing would look like "'banana' is not a valid field for class VariableCatalog.", and, should you choose, followed by "Valid fields are [list of fields]", though I don't require the list of fields.

tests/testthat/test-variable-catalog.R
```diff
@@ -99,12 +98,38 @@ with_mock_crunch({
...
+        expect_identical(
+            names(as.data.frame(varcat, keys = "all")),
+            c("name", "discarded", "alias", "type", "id", "description", 
+                "notes", "subvariables", "subvariables_catalog", "resolution", 
+                "rollup_resolution")
+        )
+    })
+    test_that("As.data.frame method errors correctly", {
+        expect_error(as.data.frame(varcat[1:3], keys = "Not a field at all"),
+            "'Not a field at all'is an invalid key for catalogs of class VariableCatalog."
```
nealrichardson: Looks like you're missing a space in your error message. Also, I'd like to see a test where the keys are a mix of valid and invalid entries, like c("name", "banana")

R/as-data-frame.R
```diff
@@ -73,3 +73,45 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#'
+#' This method allows you to import the VariableCatalog into your R sessions in 
+#' order to facilliate interactive use. Modifying the dataframe produced by this
+#' method will not update the Crunch web app, for information about modifying 
+#' variables in the app see: \code{vingette("variables",
+#' package = "crunch")}  
+#' 
+#' @param x
+#' A VariableCatalog produced by \code{variables(ds)}.
+#' @param ... Additional arguments passed to \code{data.frame}
```
jonkeane: This is stylistic, but I usually put the ... at the end (unless it's one of the functions where we use ... for the main arguments like `makeCaseVariable`)

R/as-data-frame.R
```diff
@@ -73,3 +73,45 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' @param row.names part of as.data.frame signature. Ignored.
+#' @param optional part of as.data.frame signature. Ignored.
+#' @param keys 
+#' A character vector of the variable catalog attributes which you would like 
+#' included in the datafame. To include all attributes or see which ones are available
+#' set keys to "all". 
+#' @return A data frame
```
jonkeane: Maybe add a bit more explanation for what's in the dataframe?

R/as-data-frame.R
```diff
@@ -73,3 +73,45 @@ as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE
...
+#' } 
+#' 
+#' @rdname VariableCatalog-to-Data-Frame
+#' @export
+as.data.frame.VariableCatalog <- function(x, 
+    row.names = NULL, 
+    optional = FALSE, 
+    ...,
+    keys = c("alias", "name", "type")) {
+    if (all(keys == "all")){
```
jonkeane: Please add a space before the `{`

R/shoji-catalog.R
```diff
@@ -240,28 +240,94 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
-        row.names=rownames, ...)
-    if (default.rownames) {
-        rownames(out) <- NULL
+catalogToDataFrame <- function(x, keys=TRUE, 
+    rownames = "default", 
+    list_columns = c("subvariables", "subvariables_catalog"),
+    ...) {
+    index <- lapply(index(x), function(a) a[keys])
+    ## Return an empty dataframe if the function is called on an empty catalog 
+    if (length(index) == 0){
```
jonkeane: Please add a space before the `{`

R/shoji-catalog.R
```diff
@@ -240,28 +240,94 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+                list_col <- l[list_col_names]
+                list_df <- data.frame(matrix(nrow = 1, ncol = length(list_col)))
+                names(list_df) <- names(list_col)
+                for (i  in seq_along(list_col)) {
+                    list_df[[1, i]] <- list_col[i] 
+                }
+                entry_df <- cbind(entry_df, list_df)
+            }
+            entry_df
+        }
```
jonkeane: This `entry_to_df` function might be better outside of this function, and then just called later on line 290. I'm not sure if there're any other places we would use it, but even so it would break up how much one has to hold in one's head while reading this.

R/shoji-catalog.R
```diff
@@ -240,28 +240,94 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+            entry_df <- as.data.frame(vect_col, stringsAsFactors = FALSE)
+            
+            if (any(names(l) %in% list_col_names)) {
+                list_col <- l[list_col_names]
+                list_df <- data.frame(matrix(nrow = 1, ncol = length(list_col)))
+                names(list_df) <- names(list_col)
+                for (i  in seq_along(list_col)) {
+                    list_df[[1, i]] <- list_col[i] 
+                }
+                entry_df <- cbind(entry_df, list_df)
```
jonkeane: I don't think that this is a problem, but something to be cognizant of, this will reorder the columns so that the list columns are always to the right of the vector columns. I think this could be rewritten to not do that, although I'm not certain it needs to. Are there any places where we don't want the catalogs to be shuffled?

R/shoji-catalog.R
```diff
@@ -240,28 +240,94 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+            rownames(out) <- NULL
+        }
+        
+        #When a bad key argument is passed to the the shoji index it returns a dataframe
+        # with an NA value. This exclude those columns.
+        exclude_cols <- grepl("^NA", names(out)) & 
+            vapply(out, function(x)all(is.na(x)), FUN.VALUE = logical(1))
+        out <- out[, !exclude_cols]
+        
+        missing_keys <- paste0("'", keys[!(keys %in% names(out))], "'")
```
jonkeane: There's a `dQuote()` function that we try to use when quoting in error messages. It does the right thing vis-a-vis curly versus non curly quotes, which makes mike (and others) happy.

tests/testthat/test-variable-catalog.R
```diff
@@ -40,22 +39,22 @@ with_mock_crunch({
...
+    
     gender.url <- "https://app.crunch.io/api/datasets/1/variables/gender/"
     test_that("Extract methods: character and numeric", {
         expect_is(varcat[[gender.url]], "VariableTuple")
         expect_identical(varcat[[gender.url]]@body,
             index(varcat)[[gender.url]])
         expect_identical(index(varcat[2:3]), index(varcat)[2:3])
     })
-
+
```
jonkeane: I'm not sure why it's not default, but I've tried to keep with linted style using the Strip trailing horizontal whitespace when saving setting (a good argument for adding .Rproj to the repo like you suggested!)

R/shoji-catalog.R
GShotwell: I did not know about `dQuote`!, very useful

R/shoji-catalog.R
```diff
@@ -240,28 +240,109 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
 #' @export
 as.list.ShojiCatalog <- function (x, ...) lapply(names(index(x)), function (i) x[[i]])
 
-#' Utility to get a more human-readable view of a Shoji Catalog
+#' A utility to retun a dataframe from a ShojiCatalog.
```
jonkeane: We should change this to be title case now that we're trying to harmonise that.

R/shoji-catalog.R
```diff
@@ -240,28 +240,109 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+        names(out) <- names
+
+        for (i in seq_along(entry_list)) {
+            for (j in  names(entry_list[[i]])) {
+                out[i, j] <- entry_list[[i]][1, j]
+            }
+        }
+        #################
+
+        out <- as.data.frame(out, rownames = rownames, ...)
```
jonkeane: The pre-NAed dataframe and then embedded `for` loops here strike me as a little bit unRlike. I did some poking, and I've come up with the following that should be able to replace lines 272--286. It's still pretty hairy (which was one of the original things I was trying to clean up, but maybe you can find ways to clean it up.

```
dargs <- unname(entry_list)
all_names <- unique(names(unlist(dargs)))
out <- do.call(rbind, lapply(dargs, function(x) {
                            unlist(lapply(all_names, function(y) x[[y]] %||% NA))}))
colnames(out) <- all_names
out <- as.data.frame(out, stringsAsFactors=FALSE)
```

I got inspiration (~most of the code) from https://stackoverflow.com/questions/17308551/do-callrbind-list-for-uneven-number-of-column which benchmarked this approach against others and this is the most performant. I know it shouldn't matter here since the catalogs should be relatively small, but good to keep similar patterns in the code.

GShotwell: I think pre-allocating a dataframe and then filling it is a common pattern in R (it's in the [R inferno](http://www.burns-stat.com/pages/Tutor/R_inferno.pdf) anyway). And in this case I find a nested for loop easier to read and understand than a nested `lapply` with a `do.call` around it. I admit that a big part of this comes from not really liking `do.call` as a pattern because of some of the stuff that Hadley mentions [here](http://rpubs.com/hadley/do-call2).

Initially this block of code used `plyr::bind_rows()` which is both much simpler and much more performant than any of the other options I'm aware of. I was told that performance is not an issue for this function because we're always going to be waiting for the API call, and that we didn't want to bring in the dependency.

R/shoji-catalog.R
```diff
@@ -240,28 +240,109 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+                list_col <- entry[list_col_names]
+                list_df <- data.frame(matrix(nrow = 1, ncol = length(list_col)))
+                names(list_df) <- names(list_col)
+                for (i  in seq_along(list_col)) {
+                    list_df[[1, i]] <- list_col[i]
+                }
+                entry_df <- cbind(entry_df, list_df)
+            }
+            entry_df
+        }
```
GShotwell: It won't fail exactly, but it will produce odd content. Here are the case: 
1) Non-list variable passed as a list variable
-  The non-list variable is returned as a list column with a single entry in each list
2) Nonsense passed to list variable argument
- No change in function behaviour
3) List variable for example `subvariable` not passed to list variables
- In this case each entry of the list will get its own column in the output database. `catalogToDataframe` will still work but just return a wider than usual dataframe. 

In terms of re-writing to pick up the list automatically. There was some discussion with Neal about this on this PR way back when. Here are my thoughts: 

1) I don't think there's a way to tell the difference between entries which are lists of length 1 and entries which are variables stored in lists of length 1. For instance if you have a variable which has only 1 sub variable, the subvariable attribute will look just like any other attribute which we don't want stored in a list column. It seems like the only way to tell those two cases apart is by hard coding the name. 
2) I think it's better for us to decide which variables _should_ be stored as list-columns in the output dataset, and either set those as the default for `catalogToDataframe()` or write `as.data.frame()` methods for other catalogs which specify the proper list-columns. To me that seems more stable than having it autodetect based on the particular catalog. Especially when you imagine someone using a post-processing script on the resulting dataframe. It's better for the dataframe variable to always have the same class than to have it change based on the catalog.

GShotwell: That said, I can test the other way and throw an error when something looks like a list column but isn't identified as one. Does that seem like a good solution?

GShotwell:
@jonkeane This is ready for review I made all of the changes you requested except for replacing that nested for loop.

R/shoji-catalog.R
```diff
@@ -240,28 +240,120 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+        #When a bad key argument is passed to the the shoji index it returns a dataframe
+        # with an NA value. This exclude those columns.
+        exclude_cols <- grepl("^NA", names(out)) &
+            vapply(out, function(x)all(is.na(x)), FUN.VALUE = logical(1))
+        out <- out[, !exclude_cols, drop = FALSE]
+        #reorder columns to match the order in which keys were supplied
+        if (keys == TRUE) {
+            ordered_names <- names(out)
+        } else {
+            ordered_names <- names(out)[names(out) %in% keys]
```
nealrichardson: This doesn't reorder, it just filters.

R/shoji-catalog.R
```diff
@@ -240,28 +240,120 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+
+        out <- as.data.frame(out, rownames = rownames, ...)
+
+        #When a bad key argument is passed to the the shoji index it returns a dataframe
+        # with an NA value. This exclude those columns.
+        exclude_cols <- grepl("^NA", names(out)) &
+            vapply(out, function(x)all(is.na(x)), FUN.VALUE = logical(1))
+        out <- out[, !exclude_cols, drop = FALSE]
+        #reorder columns to match the order in which keys were supplied
+        if (keys == TRUE) {
```
nealrichardson: This doesn't do what you think it does if keys is a character vector.

R/shoji-catalog.R
```diff
@@ -240,28 +240,120 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+            ordered_names <- names(out)
+        } else {
+            ordered_names <- names(out)[names(out) %in% keys]
+        }
+        out <- out[, ordered_names, drop = FALSE]
```
nealrichardson: Just do `out[, keys, drop=FALSE]`. No need for the `if` above, which doesn't work correctly anyway.

GShotwell:
@jonkeane @nealrichardson this is ready for review

R/as-data-frame.R
jonkeane: Is there a reason we're keeping the ... up at the beginning in the documentation? It's at the end in the function call.

R/shoji-catalog.R
```diff
@@ -240,28 +240,114 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+#' @return A one row data frame
+#'
+entryToDF <- function(entry, list_col_names ){
+            entry[vapply(entry, is.null, logical(1))] <- NA
+            vect_col <- entry[!(names(entry) %in% list_col_names)]
+            mislabled_list_cols <- lapply(vect_col, length) > 1
+
+            error_text <- " contain more than one entry and is not included in list_col_names"
+            if (sum(mislabled_list_cols) == 1) {
+                error_text <- " contains more than one entry and are not included in list_col_names"
```
jonkeane: I think these error texts might be mismatched. line 332 starts of plural (has "contain") but then uses "is" later. The same (though opposite) is true for line 334. I think 332 should probably be:
" contain more than one entry and are not included in list_col_names"
and 334 should be:
" contains more than one entry and is not included in list_col_names"

It also would be a bit easier (on humans) to read if this were in an `if` `else`. When I first was reading through this I thought they were the same message, just accidentally copied and pasted.

tests/testthat/test-variable-catalog.R
```diff
@@ -101,12 +100,77 @@ with_mock_crunch({
...
+          )
+
+      expected_df <- data.frame(alias = "birthyr",
+                                name = "Birth Year",
+                                type = "numeric",
+                                stringsAsFactors = FALSE)
+      expect_identical(entryToDF(entry, c("subvariables", "subvariables_catalog")),
+        expected_df)
+      expect_identical(entryToDF(entry, c("subvariables", "not_a_variable")),
+        expected_df)
```
jonkeane: Aren't these both testing that when you give listcols that don't exist you still get what you expect back?

Additionally, it would be good to test what the format looks like when you expect there to be list columns to be present.

R/shoji-catalog.R
```diff
@@ -240,28 +240,114 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+            }
+            if (any(mislabled_list_cols)) {
+                halt( serialPaste( dQuote(names(vect_col[mislabled_list_cols]))),
```
jonkeane: You can remove these spaces here. We don't usually store things to be used once, but given how long this is, it might be worth storing `names(vect_col[mislabled_list_cols])` or `dQuote(names(vect_col[mislabled_list_cols]))` temporarily for readability (thought it's right on the border, so don't feel obligated).

R/shoji-catalog.R
GShotwell: Thanks for the grammar note, I'm going to leave the pattern as is unless we've made a decision to not use that one. Personally I find assignment then replace a bit better than if/else for this kind of work.

tests/testthat/test-variable-catalog.R
GShotwell: One is asking for two variables which exists, but isn't a list-column, one includes a variable which doesn't exist.

GShotwell:
@jonkeane  This is ready

tests/testthat/test-variable-catalog.R
```diff
@@ -101,6 +100,96 @@ with_mock_crunch({
...
+                name = "mymrset",
+                discarded = FALSE,
+                alias = "mymrset",
+                type = "multiple_response",
+                id = "949d2dc7e7a24e6090cc88bb92e1d2fb",
+                description = "Please select all that apply",
+                notes = "",
+                subvariables = list(
+                    list(
+                        subvariables = list(
```
nealrichardson: This doesn't look right: too much nesting. Make a test on `varcat` above that asserts what the "subvariables" column in the data.frame should look like. I think that will clarify what the output should be.

nealrichardson:
Would really like to be done with this PR but I still have one concern about the shape of the list-column output. Should be easily tested and addressed.

R/shoji-catalog.R
```diff
@@ -240,28 +240,111 @@ setMethod("emails", "ShojiCatalog", function (x) getIndexSlot(x, "email"))
...
+        ################
+        entry_list <- lapply(index, entryToDF, list_col_names = list_columns)
+        names   <- unique(unlist( lapply(entry_list, names)))
+        out <- data.frame(matrix(nrow = length(entry_list), ncol = length(names)))
+        names(out) <- names
+
+        for (i in seq_along(entry_list)) {
+            for (j in  names(entry_list[[i]])) {
+                # Overwriting a entry in a dataframe with a longer list
+                # can trigger a warning, which needs to be suppressed.
```
nealrichardson: What's the warning? I'd read that as a sign that the code isn't doing the right thing.

tests/testthat/test-variable-catalog.R
```diff
@@ -101,6 +100,94 @@ with_mock_crunch({
...
+                stringsAsFactors = FALSE
+            ))
+    })
+    test_that("as.data.frame method returns all fields when keys = 'all'", {
+        expect_identical(
+            names(as.data.frame(varcat, keys = "all")),
+            c("name", "discarded", "alias", "type", "id", "description",
+                "notes", "subvariables", "subvariables_catalog", "resolution",
+                "rollup_resolution")
+        )
```
nealrichardson: Assert what the "subvariables" column in that data.frame should look like.

R/shoji-catalog.R
GShotwell: That is is :)

Issue #90, opened 2017-07-28
login() errors in Rscript


nickp-yg:
when invoked in a script called by `$ Rscript`, `login()` returns the following error:
```
Error in initialize(value, ...) :
  cannot use object of class "shoji" in new():  class "ShojiObject" does not extend that class
Calls: login ... getAPIRoot -> ShojiObject -> new -> initialize -> initialize
```
Brute-force troubleshooting has revealed the root of this error to be the behavior of `$ Rscript` vs  interactive R, as `$ Rscript` does not import the methods package by default. I created a pull request ( #91 ) where I simply explicitly import the `methods` package, please let me know if this is the best course to solve this issue.

PR #91, opened 2017-07-28
Added import of methods package to avoid login() errors on non-interactive use.

nickp-yg:
In reference to issue #90


nealrichardson:
Done with a more targeted import in https://github.com/Crunch-io/rcrunch/commit/89778d207b89b38e01d42b8d6aebd02a0ceab435.

PR #93, opened 2017-08-03
Add cut function for numericVariables

### LINKED ISSUE #69: cut
A cut function like the one that exists in base R would be extremely helpful. It would be mostly be used to derive age group variables.

GShotwell:
This closes #69 by creating a cut method for the crunch `NumericVariable` class. Most of the main function is copied from `base::cut()` with the following exceptions: 
- `max()` and `min()` are used instead of `range()` as I couldn't find a `range()` method for numeric variables
- The `ordered_result()` argument is ignored because it seems like a less common use case
- `makeCaseVariable()` is used to do the main work of creating the variable. As a result a `variable.name()` argument needs to be added to the cut function.


tests/testthat/test-variable-summary.R
```diff
@@ -1,91 +1,97 @@
...
+  
+  test_that("Summary method for numeric", {
+    expect_is(summary(ds$birthyr), "NumericVariableSummary")
+  })
+  test_that("max", {
+    expect_equal(max(ds$birthyr), 1.6662)
+  })
+  test_that("min", {
+    expect_equal(min(ds$birthyr), -1.4967)
+  })
```
jonkeane: Is this just a whitespace removal here, or is there something else that's changed? Github's view is not making it easy for me to tell: it looks like just whitespace to me, but I might be missing something.

GShotwell: I added tests for`max` and `min` but also added and removed some code so may have change the whitespace a bit

R/cut.R
```diff
@@ -0,0 +1,105 @@
...
+#'
+#' @examples
+#' \dontrun{
+#' ds <- loadDataset("mtcars")
+#' ds$cat_var <- cut(ds$mpg, 3, variableName = "new_var")
+#' }
+#'  
+setMethod("cut", "NumericVariable", function(x, 
+                                             breaks, 
+                                             variable.name,
```
nealrichardson: Check out how `combine` works for the variable metadata attributes: https://github.com/Crunch-io/rcrunch/blob/master/R/combine-categories.R#L28

A default "name" is given if one isn't supplied in `...` based on the number of categories in the result--you can do the same here. Other attributes not passed in `...` are brought from the original variable via `copyVariableReferences`.

R/cut.R
```diff
@@ -0,0 +1,105 @@
...
+  env <- environment()
+  if (length(breaks) == 1L) {
+    if (is.na(breaks) || breaks < 2L) {
+      halt("invalid number of intervals")
+    }
+    nb <- as.integer(breaks + 1) # one more than #{intervals}
+    dx <- diff(rx <- c(min(x, na.rm = TRUE), max(x, na.rm = TRUE)))
+    if (dx == 0) {
+      dx <- abs(rx[1L])
+      breaks <- seq.int(rx[1L] - dx/1000, rx[2L] + dx/1000,
```
nealrichardson: For an integer "breaks" value, you could:

1) keep this copy-paste from `base::cut`;
2) reject the request (the main use for the function now is to specify specific cut points); or
3) return a VariableDefinition with a "bin" derivation function. The backend supports this autobinning already.

For expediency, I'd personally go with (2) and wait for someone to for it.

R/cut.R
```diff
@@ -0,0 +1,105 @@
...
+                        length.out = nb)
+    } else {
+      breaks <- seq.int(rx[1L], rx[2L], length.out = nb)
+      breaks[c(1L, nb)] <- c(rx[1L] - dx/1000, rx[2L] + dx/1000)
+    }
+  }else nb <- length(breaks <- sort.int(as.double(breaks)))
```
nealrichardson: Maybe this is copied from the R source, but it's not conforming with modern style best practice (and not with this repository either). No "else" on a single line; no assignment inside another function call.

And 4 space tabs, please.

R/cut.R
```diff
@@ -0,0 +1,105 @@
...
+  if (anyDuplicated(breaks)) halt("'breaks' are not unique")
+  if (is.null(labels)) {#- try to construct nice ones ..
```
nealrichardson: Consider factoring this labeling code out into another function. Will make it easier for us to test, and easier for us to decide to do something different. I'm not sure that set notation will resonate with our target users.

R/cut.R
```diff
@@ -0,0 +1,105 @@
...
+                    varname, 
+                    " & ",  
+                    varname,
+                    comp_2,
+                    breaks[i])
+    )
+  }
+  cases <- lapply(cases, function(x) eval(x, envir = env))
+  case_list <- lapply(seq_along(cases), function(x) list(expression = cases[[x]], name = labels[x]))
+  makeCaseVariable(cases = case_list, name = variable.name)
```
nealrichardson: `...` needs to be passed to `makeCaseVariable`

R/cut.R
GShotwell: Yup, it was from the `base::cut()` code

R/cut.R
GShotwell: Hmmmm, should we bring this behaviour into `makeCaseVariable` then?

nealrichardson: I don't think so. The difference in `makeCaseVariable` is that you could be making a variable with several others as inputs, so there's not a base variable to `copyVariableReferences` from. `combine` is like `cut` in that you give it a variable and get a variable (definition) back derived from the single input variable.

GShotwell: We decided to halt the execution if a variable name is not present rather than provide a default value.

R/cut.R
GShotwell: In this case I think there's a lot of value in just replicating `base::cut()` as much as possible. It's always very surprising when a base method that the user is used to returns a value which isn't what they'd expect, even if the new behaviour is totally better than the old behaviour.

Practically, there's some examples where the two values should match. For instance if someone has a script which does the following:
- Download dataset
- Recode variable using `cut(ds$v1, 3)
- Perform additional analysis

But then they discover that they can recode variables on the app and so change their script to do the recoding before the download, they would expect that the rest of the code should still work. If we prevent the integer breaks, or bin the values different then that won't necessarily be the case.

R/cut.R
nealrichardson: Please name the argument "name" for consistency.

R/cut.R
nealrichardson: This has still not been fixed

R/cut.R
```diff
@@ -0,0 +1,140 @@
...
+            breaks <- seq.int(rx[1L], rx[2L], length.out = nb)
+            breaks[c(1L, nb)] <- c(rx[1L] - dx/1000, rx[2L] + dx/1000)
+        }
+    }else {
+        breaks <- sort.int(as.double(breaks))
+        nb <- length(breaks)
+    }
+    if (anyDuplicated(breaks)) halt("'breaks' are not unique")
+    if (is.null(labels)) { #Autogenerate labels if not supplied
+      labels <- crunch:::generateCutLabels(dig.lab, breaks, nb, right, include.lowest)
```
nealrichardson: don't need to ::: a function in your own package. `check` will whine about this, I think.

R/cut.R
```diff
@@ -0,0 +1,140 @@
...
+        comp_1 <- " <= "
+        comp_2 <- " < "
+    } else {
+        comp_1 <- " < "
+        comp_2 <- " <= "
+    }
+    cases <- vector("character", length = length(breaks) - 1)
+    varname  <- deparse(substitute(x))
+    for (i in 2:length(breaks)) {
+        cases[i - 1] <- parse(
```
nealrichardson: I'd like to see this done without `eval(parse(text=paste0(stuff)))` `x` is a `NumericVariable`; `x < 5` is a `CrunchLogicalExpr`, which is what `makeCaseVariable` wants. No need to deparse, concatenate, parse, eval to get there.

R/cut.R
```diff
@@ -0,0 +1,140 @@
...
+                          comp_1, 
+                          varname, 
+                          " & ",  
+                          varname,
+                          comp_2,
+                          breaks[i])
+        )
+    }
+    cases <- lapply(cases, function(x) eval(x, envir = env))
+    case_list <- lapply(seq_along(cases), function(x) list(expression = cases[[x]], name = labels[x]))
```
nealrichardson: This looks like an mapply.

R/cut.R
```diff
@@ -0,0 +1,140 @@
...
+    makeCaseVariable(cases = case_list, name = variable.name, ...)
+}
+)
+
+
+#' Generate Labels for the cut function
+#' 
+#' A convenience function to generate labels for the cut function. This
+#' function is extracted from base::cut() and is broken out to make it easier to
```
nealrichardson: [base::cut()]

tests/testthat/test-cut.R
```diff
@@ -0,0 +1,177 @@
...
+   )
+   expect_equivalent(
+       dig_lab,
+       cut(ds$birthyr,
+           3,
+           dig.lab = 2,
+           variable.name = "new_var")
+   )
+   expect_identical(
+       crunch:::generateCutLabels(5, c(2.111111, 3, 4, 5), 4, FALSE, FALSE),
```
nealrichardson: don't need to ::: in tests, testthat evals in the package namespace.

tests/testthat/test-variable-summary.R
```diff
@@ -1,91 +1,97 @@
...
-        expect_error(min(ds$textVar),
-            paste(dQuote("min"), "is not defined for TextVariable"))
-        expect_error(max(ds$textVar),
-            paste(dQuote("max"), "is not defined for TextVariable"))
-    })
-
-    test_that("Summary method for numeric", {
-        expect_is(summary(ds$birthyr), "NumericVariableSummary")
-    })
+  ds <- loadDataset("test ds")
```
nealrichardson: Please undo your editor's linting. Tabs should remain at 4 spaces.

tests/testthat/test-cut.R
```diff
@@ -0,0 +1,177 @@
...
+       crunch:::generateCutLabels(2, c(2.111111, 3, 4, 5), 4, FALSE, FALSE),
+       c("[2.1,3)", "[3,4)", "[4,5)")
+   )
+  })
+})
+
+with_test_authentication({
+    ds <- newDataset(df)
+    
+    test_that("cut returns the same thing for crunch variables and identical vectors", {
```
nealrichardson: Please break this into multiple `test_that`s with informative labels that explain what each test case is asserting differently.

tests/testthat/test-cut.R
```diff
@@ -0,0 +1,177 @@
+context("cut function")
+
+with_mock_crunch({
+  ds <- loadDataset("test ds")
+  gen <- ds$gender
+  test_that("cut returns expected output", {
+    
+##################### EXPECTED OUTPUT ######################################
+    basic_output <- list(name = "new_var", derivation = list(`function` = "case",
```
nealrichardson: Could you try to make these a little more readable? The regular `dput` output breaks lines in places that aren't the most natural.

R/cut.R
```diff
@@ -0,0 +1,132 @@
...
+#' 
+generateCutLabels <- function(dig.lab, breaks, nb, right, include.lowest) {
+    for (dig in dig.lab:max(12L, dig.lab)) {
+        ## 0+ avoids printing signed zeros as "-0"
+        ch.br <- formatC(0 + breaks, digits = dig, width = 1L)
+        ok <- all(ch.br[-1L] != ch.br[-nb])
+        if (ok ) break
+    }
+    labels <- if (ok) {
+            paste0(if (right) "(" else "[",
```
jonkeane: For a single line like this, I usually prefer to use `ifelse`

R/cut.R
```diff
@@ -0,0 +1,132 @@
...
+#' @param right  see `cut()`
+#' @param include.lowest see`cut()`
+#'
+#' @return
+#' A character vector of labels
+#' @examples
+#' 
+#' crunch:::generateCutLabels(2, c(2, 3, 4, 5), 4, FALSE, FALSE)
+#' 
+generateCutLabels <- function(dig.lab, breaks, nb, right, include.lowest) {
```
jonkeane: Are there tests for `generateCutLabels` that I'm missing? We should probably add them if we don't have them already.

R/cut.R
```diff
@@ -0,0 +1,132 @@
...
+#' @param breaks see `cut()`
+#' @param nb The number of breaks, equal to the small of 2 or the number of breaks
+#' @param right  see `cut()`
+#' @param include.lowest see`cut()`
+#'
+#' @return
+#' A character vector of labels
+#' @examples
+#' 
+#' crunch:::generateCutLabels(2, c(2, 3, 4, 5), 4, FALSE, FALSE)
```
GShotwell: I'm going to keep the example, but I really think every function should have an example.

GShotwell:

R/cut.R
```diff
@@ -0,0 +1,129 @@
...
+#'
+setMethod("cut", "NumericVariable", function(x,
+                                             breaks,
+                                             name,
+                                             labels = NULL,
+                                             include.lowest = FALSE,
+                                             right = TRUE,
+                                             dig.lab = 3,
+                                             ordered_result = FALSE, ...){
+    env <- environment()
```
jonkeane: Is this `env` used somewhere? I've searched and can't seem to find it used after this

R/cut.R
jonkeane: I might be missing these, but I don't see tests for `generateCutLabels`

R/cut.R
jonkeane: I think `ifelse` would be cleaner here since everything is all on one line. The guidelines I've been using is: 
```
if (foo) {
    bar
} else {
    baz
}
```
Or if it is that simple and I want it on one line `ifelse(foo, bar, baz)`

GShotwell: Me too, this bit was part of the code copied from `base::cut()` and I wasn't sure of the policy around `ifelse`. I'll make the change, but FYI `ifelse`  is still a bit slower than the full `if() { } else` pattern even for one-line functions, though that probably doesn't matter much for this function. 

```
f1 <- function(){
    if( runif(1) > 0.5) {
        "foo"
    } else {
        "bar"
    }
}

f2 <- function(){
    ifelse(runif(1) > 0.5, "foo", "bar")
}

microbenchmark::microbenchmark(f1, f2)
```

```
 expr min lq  mean median   uq  max neval
   f1  25 27 33.05     28 32.5  389   100
   f2  25 27 47.49     29 32.5 1768   100
```

GShotwell:
@jonkeane @nealrichardson  This is ready for review

R/cut.R
jonkeane: Interesting I didn't know about the performance differential there. That's good to know. Also, that distribution of times is very interesting. It looks like there was likely only a handful of really long `ifelse`s given the medians are almost equal.

tests/testthat/test-cut.R
```diff
@@ -0,0 +1,397 @@
...
+    })
+
+    test_that("cut throws error when no variable name supplies", {
+        expect_error( cut(ds$birthyr, 3), "Must provide the name for the new variable")
+    })
+
+    test_that("cut labels are generated correctly", {
+        expect_identical( c("[2,3)", "[3,4)", "[4,5)"),
+            generateCutLabels(2, c(2, 3, 4, 5), 4, FALSE, FALSE)
+        )
```
jonkeane: This is purely stylistic with `expect_identical` because of the symmetrically of identity, but usually the code being tested is the first argument and the check of it is the second.

tests/testthat/test-cut.R
```diff
@@ -0,0 +1,397 @@
...
+
+    test_that("cut labels are generated correcty", {
+        expect_identical(
+            crunch:::generateCutLabels(5, c(2.111111, 3, 4, 5), 4, FALSE, FALSE),
+            c("[2.1111,3)", "[3,4)", "[4,5)")
+        )
+        expect_identical(
+            generateCutLabels(2, c(2.111111, 3, 4, 5), 4, FALSE, FALSE),
+            c("[2.1,3)", "[3,4)", "[4,5)")
+        )
```
jonkeane: You don't need `crunch:::` here, the way that testthat runs its tests it loads even unexported functions

GShotwell:

GShotwell:
@jonkeane I think this is ready

jonkeane:
I just pushed some changes that were quick and easy. I added tests for some of the input validation that was there already / that I added. We've got unit test coverage of all be three lines in cut.R. It looks like the three lines that aren't covered now might be better in an integration test (if I'm reading everything correctly they require the vector to be all the same value, right?). If there are no objections to my changes, this looks good to go.

R/cut.R
```diff
@@ -0,0 +1,129 @@
+#' Cut a numeric Crunch variable
+#'
+#' `crunch::cut()` is equivalent to `base::cut()` except that it operates on
+#' Crunch variables instead of in-memory variables. The function divides the range of
+#' `x` into intervals and codes the values in x according to which interval they fall.
+#' The leftmost interval corresponds to level one, the next leftmost to level two and
+#' so on.
```
nealrichardson: Since we've been working on documentation, let's make this description nicer. I know this is mostly copied from base::cut, but that doesn't mean we can't do better :)

R/cut.R
```diff
@@ -0,0 +1,129 @@
...
+#' It determines the number of digits used in formatting the break numbers.
+#' @param ordered_result	Ignored.
+#' @param ... further arguments passed to [makeCaseVariable]
+#' @return a Crunch [`VariableDefinition`]. Assign it into the dataset to create
+#' it as a derived variable on the server.
+#' @export
+#' @examples
+#' \dontrun{
+#' ds <- loadDataset("mtcars")
+#' ds$cat_var <- cut(ds$mpg, 3, variableName = "new_var")
```
nealrichardson: A more realistic example will include specific cut points and labels for the categories to be created.

Also, `variableName` is not a valid argument.

R/cut.R
```diff
@@ -0,0 +1,129 @@
...
+                                             include.lowest = FALSE,
+                                             right = TRUE,
+                                             dig.lab = 3,
+                                             ordered_result = FALSE, ...){
+    if (missing(name)) {
+        halt("Must provide the name for the new variable")
+    }
+    if (length(breaks) == 1L) {
+        if (is.na(breaks) || breaks < 2L) {
+            halt("invalid number of intervals")
```
nealrichardson: I know this is verbatim what `base::cut.default` says, but the argument is called "breaks" and the message says "intervals"

R/cut.R
```diff
@@ -0,0 +1,129 @@
...
+        } else {
+            breaks <- seq.int(rx[1L], rx[2L], length.out = nb)
+            breaks[c(1L, nb)] <- c(rx[1L] - dx/1000, rx[2L] + dx/1000)
+        }
+    } else {
+        breaks <- sort.int(as.double(breaks))
+        nb <- length(breaks)
+    }
+    if (anyDuplicated(breaks)) {
+        halt(sQuote("breaks"), " are not unique")
```
nealrichardson: likewise, " must be unique"

R/cut.R
```diff
@@ -0,0 +1,129 @@
...
+    }
+    if (is.null(labels)) { # Autogenerate labels if not supplied
+        labels <- generateCutLabels(dig.lab, breaks, nb, right, include.lowest)
+    } else if (length(labels) != nb - 1L) {
+        halt("lengths of ", sQuote("breaks"), " and ", sQuote("labels"), " differ")
```
nealrichardson: This one puzzled me. The message says that the problem is that the lengths of breaks and labels differs, but in order for it to be valid, apparently you have to have one more "break" than "label"! Let's give a better message and clearer instructions in the man page.

Some other considerations: 

* Should we allow "labels" to be full category definitions? There are other attributes that our data type supports.
* Should we add an argument or something that changes the behavior for extreme values, i.e. makes categories for them? So that the correct number of "labels" to provide would be length(breaks) + 1, not -1. Looking at the base implementation, 

```
> cut(1:20, c(4, 8, 13))
 [1] <NA>   <NA>   <NA>   <NA>   (4,8]  (4,8]  (4,8]  (4,8]  (8,13] (8,13]
[11] (8,13] (8,13] (8,13] <NA>   <NA>   <NA>   <NA>   <NA>   <NA>   <NA>  
Levels: (4,8] (8,13]
```

my low and high extremes are just "NA", but we could treat them as different categories. Even without adding an argument about that, we could make them be separate categories (missing: true) that the user could still work with after creating the variable.

tests/testthat/test-cut.R
```diff
@@ -0,0 +1,439 @@
...
+
+        ds$cat_var5 <- cut(ds$v1,
+            3,
+            include.lowest = TRUE,
+            name = "new_var5")
+        expect_identical(
+            cut(df$v1, 3, include.lowest = TRUE),
+            as.vector(ds$cat_var5)
+        )
+    })
```
nealrichardson: I'd say the first use case our users will have is turning "age" or "birthyear" into something like "age4", binned like 18-29, 30-44, 45-64, 65+. A test (and man page example) illustrating that would be very helpful.

R/cut.R
GShotwell: I agree the message is a bit confusing, so I changed it to talk about "resulting categories " which is what should match the number of labels. It's just a bit confusing because `breaks` refers to the start and end points of the category, which is always one more than the number of categories. 

I like the idea of full category definitions but my intuition about this function is that we should try to keep it fairly simple and close to the base functions. Maybe we can hold off including the category definitions and see users request that functionality?

For the extreme value behaviour, I think there's kind of an argument either way. It seems like there's a use case for either distinguishing between high-value NA, and low-value NA, but on the other hand the user could think about numbers which fall outside of the range just as "unspecified" in which case they would want the same value.  I'm not sure an extra argument makes a lot of sense because it's more work than just specifying the breaks to capture those extreme values. For instance: 
```
cut(1:1000, 
       breaks = (5, 10, 100), 
       labels = c("tiny", "small", "medium", "large"),
       code_extreme = TRUE)
``` 
would be equivalent to 
```
cut(1:1000,
      breaks = (-10000, 5, 10, 100, 10000),
      labels = c("tiny", "small", "medium", "large")
```
but the former is more verbose and requires the user to be familiar with another argument.

R/cut.R
```diff
@@ -0,0 +1,137 @@
...
+#' Crunch variables instead of in-memory variables. The function takes a numeric variable
+#' and recodes it into a new categorical variable based on the breaks arugment. You can either
+#' break the variable into evenly spaced categories by specifying the number of breaks, or specify
+#' a numeric vector which identifies the start and end point of each category. For instance specifying
+#' `breaks = 5` will break the vector into five evenly spaced portions while `breaks = c(1, 5, 10)` will
+#' recode the data into two groups based on whether the numeric vector falls between 1 and 5 or 5 and 10.
+#' @param x A crunch variable of the class NumericVariable
+#' @param breaks Either a numeric vector of two or more unique cut points
+#' or a single number giving the number of intervals into which x is to be cut.
+#' @param labels labels for the levels of the resulting category.
```
nealrichardson: Indicate that they should be strings and if breaks are specified, there should be one fewer label than break.

R/cut.R
```diff
@@ -0,0 +1,137 @@
...
+#' By default, labels are constructed using interval notation.
```
nealrichardson: s/By default/If not specified/ (and perhaps give an example of the set notation)

inst/app.crunch.io/api/datasets/1/cube-d06077.json
```diff
@@ -0,0 +1,48 @@
+{
+  "element": "shoji:view",
```
nealrichardson: Fixtures should also be indented 4 spaces instead of 2

R/cut.R
nealrichardson: Fine to defer full category definitions and extreme behavior and see what users think.

PR #94, opened 2017-08-07
Updated write.csv definition to better dispatch appropriately 

jonkeane:
see https://github.com/r-lib/pkgload/issues/45


nealrichardson:
LGTM though it makes me think the generic should actually be defined on write.table and not write.csv.

PR #96, opened 2017-08-10
ordering from Persephone

jonkeane:
I've cleaned up `copyOrder` and added both unit and integration tests. I've left the code for `moveToAfter` there, but unexported for now. When Persephone is back from holiday, I'll work with her to see about cleaning that up, testing it, and then exporting again. I'm happy to remove it from here and just have her copy and paste it back in from her old PR as well if that's preferred.


R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+    new_ord <- lapply(ord, copyOrderGroup,  source = source, target = target)
+
+    # TODO: copy other VariableOrder features?
+
+    return(do.call(VariableOrder, new_ord))
+}
+
+#' Copy the order of a `VariableGroup` (or individual variable url) from `VariableOrder`
+#'
+#' @param gr the group or variable url to be copied
```
nealrichardson: No need to have an obfuscated name like `gr`

R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+#' @param source the dataset you want to copy the order from
+#' @param target the dataset you want to copy the order to
+#' @return returns either a [`VariableGroup`] (if a group is supplied) or a url (if just a variable url is supplied)
```
nealrichardson: is this valid roxygen markdown?

also, URL is capitalized

R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+
+    ord <- entities(ordering(source))
+    new_ord <- lapply(ord, copyOrderGroup,  source = source, target = target)
+
+    # TODO: copy other VariableOrder features?
+
+    return(do.call(VariableOrder, new_ord))
+}
+
+#' Copy the order of a `VariableGroup` (or individual variable url) from `VariableOrder`
```
nealrichardson: Why bother documenting a function like this?

R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+#' @param gr the group or variable url to be copied
+#' @param source the dataset you want to copy the order from
+#' @param target the dataset you want to copy the order to
+#' @return returns either a [`VariableGroup`] (if a group is supplied) or a url (if just a variable url is supplied)
+#' @examples
+#' \dontrun{
+#' ordering(ds) <- copyOrder(ds1, ds)
+#' }
+#' @keywords internal
+copyOrderGroup <- function(gr, source, target) {
```
nealrichardson: This would be simpler (and probably faster) if instead of passing in datasets and calling this `aliasFromURL` repeatedly, pass in two vectors or lists: `source` has URLs as names and aliases as values, and `target` has aliases as names and URLs as values.

R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+    ents <- vector("list", length(gr))
+
+    # deal with non-groups (if there are no non-groups, nothing will happen here)
+    als <- aliasFromURL(entities(gr[!groups]), source)
+    ents[!groups] <- urls(target[als[als %in% aliases(allVariables(target))]])
+
+    # deal with groups (if there are no groups, nothing will happen here)
+    ents[groups] <- lapply(entities(gr[groups]), copyOrderGroup,  source = source, target = target)
+
+    return(VariableGroup(name(gr), ents))
```
nealrichardson: You're probably going to need to call `removeMissingEntities` somewhere here at the end because you can't assume that all aliases in dataset1 are found in dataset2

nealrichardson: And check out how that function handles the recursion; could give you some ideas for simplifying here.

R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+#'
+#' @param x the variable group that `after` is in
+#' @param value the variable or dataset subset you would like to move
+#' @param after the variable you want to precede `value`
+#' @return returns a variable group
+#' @examples
+#' \dontrun{
+#' ordering(ds)[['Demographics']] <- moveToAfter(ordering(ds)[['Grp A']], ds['age'], ds$educ)
+#' }
+moveToAfter <- function(x, value, after){
```
nealrichardson: I'd just remove or comment this entirely if you're not really implementing/testing/exporting this yet.

tests/testthat/test-variable-order.R
```diff
@@ -830,4 +842,76 @@ with_test_authentication({
         expect_identical(urls(ordering(ds)[["Group 2.5"]]),
             urls(variables(ds[c("v4", "v6")])))
     })
+
+    test_that("copyOrder copies across datasets with simple order", {
```
nealrichardson: These tests have some nice features (test simple and complex orders), but they're too simplistic. A fork has the same variable IDs as its parent, so the translation could falsely appear to succeed if we're doing something sloppy. More importantly, I'd expect that users will be copying orders between datasets that they think are the same but really aren't and have some subtle differences in variables/aliases. So that needs to be handled and tested.

R/ordering.R
jonkeane: I know it's not strictly necessary, but it makes it easier when I / someone else comes back to this code in the future to see at least what the intended interface is supposed to be. I'm fine moving this kind of documentation into non-roxygen strings so it doesn't get parsed / we don't have to maintain it as well, but I think having something similar to this is good even for internal functions.

R/ordering.R
jonkeane: Are you asking about the ```[`VariableGroup`]``` part? It is, roxygen does the right thing and flips those around in the Rd into `\code{\link{VariableGroup}}` which (via aliasing) gets us to VariableOrder.Rd.

R/ordering.R
```diff
@@ -82,3 +82,90 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+#' @export
+copyOrder <- function(source, target){
+    if (!is.dataset(source) | !is.dataset(target)) {
+        halt("Both source and target must be Crunch datasets.")
+    }
+
+    ord <- entities(ordering(source))
+    new_ord <- lapply(ord, copyOrderGroup,  source = source, target = target)
+
+    # TODO: copy other VariableOrder features?
```
jonkeane: The features that I see that are missing from the new `VariableOrder` but that are populated when pulled using `ordering()` are `element` `self` and `catalog_url`. I'm not sure if these are useful to add here (their absence doesn't impact assigning the order using `ordering(target) <- order_from_CopyOrderGroup`).

nealrichardson: Both would be inappropriate since they point to the source dataset.

jonkeane: I was imagining that we would put the appropriate target dataset values in here (admittedly, "copy" is not the right word here). One benefit of making sure that `catalog_url` is filled in properly is that the `show` method would work on the output of `copyOrder` since it uses `catalog_url` to fetch aliases. This might be useful if a user wanted to see what order they were copying over before committing  the reordering.

Example output with and without assigning `catalog_url`
```
[crunch] > new_order <- copyOrder(ds, ds_alt)
[crunch] > new_order
(Hidden variable)
(Hidden variable)
[+] Group A
    (Hidden variable)
    (Hidden variable)
[crunch] > new_order@catalog_url <- "http://local.crunch.io:8080/api/datasets/9bee9d4e24ec49229f5ad07e4a76f7ec/variables/?relative=on" # target dataset URL
[crunch] > new_order
v5
v6
[+] Group A
    v4
    v3
[crunch] > 
```

tests/testthat/test-variable-order.R
jonkeane: I've added a test on an alternate target dataset where the target is missing some variables in the target and the target has some extra variables that the source does not have. Additionally, the ids for the common variables are different, so we're checking that the ids aren't what are being used.

R/ordering.R
jonkeane: Thanks, I was messing `removeMissingEntities` I've actually had to include it twice since both `VariableGroup` and `VariableOrder` fail if given `NULL` or `NA` entries. I've simplified  `copyOrderGroup` a bunch since it traverses until it gets to a single character — this isn't exactly the same pattern as `removeMissingEntities` but is close (and much simpler than what was there before!)

R/ordering.R
```diff
@@ -82,3 +82,94 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+#' @export
+copyOrder <- function (source, target) {
+    if (!is.dataset(source) | !is.dataset(target)) {
+        halt("Both source and target must be Crunch datasets.")
+    }
+
+    ord <- entities(ordering(source))
+
+    # make url and alias maps
+    url_to_alias_source <- structure(aliases(variables(source)), .Names=urls(variables(source)))
```
nealrichardson: these should be `allVariables` not `variables` so that hidden variables aren't excluded

R/ordering.R
```diff
@@ -82,3 +82,94 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+#' @param source_map url to alias map for source variables
+#' @param target_map alias to url map for target variables
+#' @return returns either a [`VariableGroup`] (if a group is supplied) or a URL (if just a variable URL is supplied)
+#' @keywords internal
+copyOrderGroup <- function (group, source_map, target_map) {
+    # if there is a single element in group, and it is a character,
+    # just return the URL in the target.
+    if (length(group) == 1 & is.character(group)) {
+        al <- source_map[group]
+        if (al %in% names(target_map)) {
```
nealrichardson: You could use the %||% operator and do this as `target_map[[al] %||% NA`

R/ordering.R
```diff
@@ -82,3 +82,94 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+                           source_map = source_map, target_map = target_map)
+
+    # drop any na entities - those that were not found in target but in source
+    ents <- removeMissingEntities(ents)
+
+    return(VariableGroup(name(group), ents))
+}
+
+# given a URL, return the alias from the dataset.
+aliasFromURL <- function(url, ds) {
```
nealrichardson: This is no longer used right?

R/ordering.R
jonkeane: I've changed them; it looks like ordering always uses and displays all variables. Should I go ahead and change https://github.com/Crunch-io/rcrunch/blob/master/R/ordering.R#L18 to use `allVariables` as well?

nealrichardson: The end result will be the same: https://github.com/Crunch-io/rcrunch/blob/master/R/ordering.R#L29-L33

Only reason to switch would be performance since `variables(ds)` does some work while `allVariables` doesn't. Semantically, I prefer `variables` there because it's a method on the Dataset object, where the standard view excludes hidden variables. Though since 'ordering' isn't excluding hidden variables, maybe it's moot.

R/ordering.R
```diff
@@ -82,3 +82,89 @@ setMethod("ordering<-", "DatasetCatalog", function (x, value) {
...
+    if (length(group) == 1 & is.character(group)) {
+        return(target_map[[source_map[[group]]]] %||% NA)
+    }
+
+    # there are groups, so recurse
+    ents <- lapply(entities(group), copyOrderGroup,
+                           source_map = source_map, target_map = target_map)
+
+    # drop any na entities - those that were not found in target but in source
+    ents <- removeMissingEntities(ents)
```
nealrichardson: I think if you make L158 return `NA_character_`, you can remove this line and L168 will succeed. Then you only have to call this recursive function once, in `copyOrder`, instead of once there and then once again in each group.

PR #100, opened 2017-08-11
Derivation retrieval

### LINKED ISSUE #98: Grab existing derivations
Derived variables can be represented as crunch variables, users should be able to use `derivation` to see what the expression is that generates the derived variable.

jonkeane:
Functions to retrieve and patch derivations. (resolves #98)


R/expressions.R
```diff
@@ -375,3 +375,7 @@ setMethod("duplicated", "CrunchVariable", function (x, incomparables=FALSE, ...)
 setMethod("duplicated", "CrunchExpr", function (x, incomparables=FALSE, ...) {
     zfuncExpr("duplicates", x)
 })
+
+#' @rdname crunch-is
+#' @export
+is.CrunchExpr <- function (x) inherits(x, "CrunchExpr")
```
nealrichardson: Good. Maybe `is.Expr` instead/also? Easy enough to alias this function to both.

R/show.R
```diff
@@ -191,8 +191,8 @@ expressionValue <- function (expr) {
     unlist(expr$column %||% expr$value)
 }
 
-deparseAndFlatten <- function (x, max_length = NULL, ...) {
-    out <- deparse(x, ...)
+deparseAndFlatten <- function (x, max_length = NULL, control=NULL, ...) {
+    out <- deparse(x, control=NULL, ...)
```
nealrichardson: `control=control`

R/variable-derivation.R
```diff
@@ -0,0 +1,82 @@
...
+    return(expr)
+})
+
+absolutifyVariables <- function (expr, var) {
+    if ("function" %in% names(expr)) {
+        expr[["args"]] <- absolutifyExpressionArgs(expr[["args"]], var)
+        return(expr)
+    } else {
+        ## Dunno what this is
+        return("[Complex expression]")
```
nealrichardson: Why this else block?

R/variable-derivation.R
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+absolutifyExpressionArgs <- function (args, var) {
```
nealrichardson: Don't pass the variable, pass its URL

R/variable-derivation.R
```diff
@@ -0,0 +1,82 @@
...
+    args <- lapply(args, function (x, var) {
+        if ("variable" %in% names(x)) {
+            ## GET URL, get alias from that
```
nealrichardson: Invalid copypasta comment

R/variable-derivation.R
```diff
@@ -0,0 +1,82 @@
...
+    }
+    expr <- CrunchExpr(expression=entity(x)@body$derivation)
+    expr@expression <- absolutifyVariables(expr@expression, x)
+
+    return(expr)
+})
+
+absolutifyVariables <- function (expr, var) {
+    if ("function" %in% names(expr)) {
+        expr[["args"]] <- absolutifyExpressionArgs(expr[["args"]], var)
```
nealrichardson: Why do you need two functions to recurse through this list? My intuition is something like

```
absolutifyVariables <- function (expr, base.url) {
    if (is.list(expr)) {
        # Recurse.
        lists <- vapply(expr, is.list, logical(1))
        expr[lists] <- lapply(expr[lists], absolutifyVariables, base.url)
        # For others, look for "variable" references
        if (length(names(expr))) {
            vars <- names(expr) %in% "variable"
            expr[!lists & vars] <- lapply(expr[!lists & vars], absoluteURL, base.url)
            return(expr)
        }
    } else {
        return(expr)
    }
}
```

Does that work? I haven't run it, just typed through the logic.

tests/testthat/test-derive.R
```diff
@@ -16,6 +18,26 @@ with_test_authentication({
...
+
+    test_that("derivation returns NULL if the variable is not derived", {
+        expect_null(derivation(ds$v3))
+    })
+
+    test_that("derivation()<- input validation ", {
+        expect_error(derivation(ds$v3) <- ds$v3 + 1,
+                     paste0("The variable ", dQuote("v3"),
+                            " must already be a derived variable."))
+        try(ds$v3c <- ds$v3 + 5)
```
nealrichardson: There is no "try" (let it fail)

tests/testthat/test-expressions.R
```diff
@@ -17,6 +17,14 @@ test_that(".dispatchFilter uses right numeric function", {
 with_mock_crunch({
     ds <- loadDataset("test ds")
 
+    test_that("is method works for both expressions and logical expressions", {
+        e1 <- try(ds$birthyr + 5)
+        expect_true(is.CrunchExpr(e1))
```
nealrichardson: just `expect_true(is.CrunchExpr(ds$birthyr + 5))`

and you can have these in the unit tests.

tests/testthat/test-derive.R
```diff
@@ -16,6 +18,26 @@ with_test_authentication({
...
+        expect_true(is.derived(ds$v3c))
+        expect_error(derivation(ds$v3c) <- "not an expression",
+                     paste0("The value ", dQuote("not an expression"),
+                            " must be a CrunchExpr, got a character instead."))
+    })
+
```
nealrichardson: I don't see a test for actually changing the derivation function.

tests/testthat/test-derive.R
```diff
@@ -1,5 +1,7 @@
 context("Derive a new variable")
 
+# TODO: mock a derivation for unit testing
```
nealrichardson: You could fabricate one like 

```
age <- ds$age
age@self <- "some other url where you put a variable entity JSON fixture with a derivation"
```

I think.

R/variable-derivation.R
```diff
@@ -0,0 +1,82 @@
...
+#' @rdname derivations
+setMethod("derivation<-", "CrunchVariable", function(x, value) {
+    if (!is.derived(x)) {
+        halt("The variable ", dQuote(name(x)), " must already be a derived variable.")
+    }
+    if (!is.CrunchExpr(value)) {
+        halt("The value ", dQuote(substitute(value)), " must be a CrunchExpr, got a ",
+             class(value), " instead.")
+    }
+
```
nealrichardson: We probably should also support NULL assignment to "integrate" (un-derive) the variable. This is fully implemented in https://www.pivotaltracker.com/story/show/149901735 (accepting that release next). Likewise `is.derived<- FALSE` assignment

R/variable-derivation.R
jonkeane: With a few slight modifications, yeah that works and is much cleaner. Thanks! I was relying a bit too much on the pattern I had seen for other recursive functions like this.

R/variable-derivation.R
jonkeane: It's been removed regardless, but was there to catch in case there was a type that I hadn't tested or caught.

R/variable-derivation.R
jonkeane: Removed

tests/testthat/test-derive.R
jonkeane: This is used throughout this test file, I didn't find a good reason it was there elsewhere either. I'll remove all of them unless there's a reason some of them should be kept.

nealrichardson: Please. It was to prevent tests from exiting prematurely on error but not fail the Jenkins build, but that's been handled elsewhere now.

tests/testthat/test-derive.R
jonkeane: It was a bit more complicated, but I think I have a pattern that works.

R/variable-derivation.R
jonkeane: Oh, I didn't realize this was ready. Great, I've added it. I'm not fully sure that `is.derived(var) <- FALSE` matches other patterns with `is.*` (when creating the docs, no other one has a value argument for example), I've created it for now.

nealrichardson: I believe `is.na(category) <- TRUE` exists but my memory may be wrong. Not a requirement to add it as long as there is at least one way to accomplish it (`derivation(var) <- NULL`)

R/dataset.R
```diff
@@ -23,6 +23,7 @@ getNrow <- function (dataset) {
 #' Is it?
 #' @rdname crunch-is
 #' @param x an object
+#' @param value a value to assign
```
nealrichardson: Perhaps the solution then is to document `is.derived` on the same man page as `derivation`

R/variable-derivation.R
```diff
@@ -0,0 +1,113 @@
...
+        ## Refresh and return
+        dropCache(datasetReference(x))
+        invisible(refresh(x))
+    }
+}
+
+#' @export
+#' @rdname derivations
+setMethod("derivation<-", c("CrunchVariable", "NULL"), function(x, value) integrateDerivedVar(x))
```
nealrichardson: If you made `integrateDerivedVar` take `value` as an arg (and ignore it), you could just pass it in directly here (FWIW).

R/variable-derivation.R
```diff
@@ -0,0 +1,113 @@
...
+
+#' @rdname crunch-is
+#' @aliases is.derived
+#' @export
+setMethod("is.derived", "CrunchVariable", function (x) {
+    stopifnot(is.variable(x))
```
nealrichardson: Method dispatch makes this stopifnot pointless.

R/variable-derivation.R
```diff
@@ -0,0 +1,113 @@
...
+    isTRUE(tuple(x)$derived)
+})
+
+#' @rdname crunch-is
+#' @aliases is.derived<-
+#' @export
+setMethod("is.derived<-", c("CrunchVariable", "logical"), function(x, value) {
+   if (!value) {
```
nealrichardson: To be safe, this perhaps should be `if (!isTRUE(value))` because 

```
> is.logical(NA)
[1] TRUE
> !NA
[1] NA
> if (!NA) print("x")
Error in if (!NA) print("x") : missing value where TRUE/FALSE needed
```

R/variable-derivation.R
```diff
@@ -0,0 +1,113 @@
...
+       integrateDerivedVar(x)
+   }
```
nealrichardson: and you should `return(x)`

R/variable-derivation.R
```diff
@@ -0,0 +1,113 @@
...
+    stopifnot(is.variable(x))
+    isTRUE(tuple(x)$derived)
+})
+
+#' @rdname crunch-is
+#' @aliases is.derived<-
+#' @export
+setMethod("is.derived<-", c("CrunchVariable", "logical"), function(x, value) {
+   if (!value) {
+       integrateDerivedVar(x)
```
nealrichardson: This should assign its return value back to `x`

R/variable-derivation.R
jonkeane: Yeah, that's true. And I like your suggestion below about documenting `is.derived` with derivations. That's exactly how `is.na` for categories is documented, and why I missed that with the other `is.*` functions.

R/variable-derivation.R
```diff
@@ -0,0 +1,124 @@
...
+# silently takes and discards `value` to make method dispatch easier
+integrateDerivedVar <- function (x, value) {
+    if (is.derived(x)) {
+        payload <- toJSON(list(derived=FALSE))
+        crPATCH(self(x), body=payload)
+
+        ## Refresh and return
+        dropCache(datasetReference(x))
+        invisible(refresh(x))
+    }
```
nealrichardson: you're not returning if the variable isn't derived. put your return outside the `if`

nealrichardson: (and test that setting derivation(var) <- NULL doesn't error even if the var is not derived)

R/variable-derivation.R
```diff
@@ -0,0 +1,124 @@
...
+            expr[!lists & vars] <- lapply(expr[!lists & vars], absoluteURL, base.url)
+            return(expr)
+        }
+    }
+    return(expr)
+}
+
+#' @export
+#' @rdname derivations
+setMethod("derivation<-", "CrunchVariable", function(x, value) {
```
nealrichardson: This signature should probably be `c("CrunchVariable", "ANY")` since your other method signature defines for both args.

inst/app.crunch.io/api/datasets/1/variables/birthyrPlus.json
```diff
@@ -0,0 +1,65 @@
+{
```
nealrichardson: 4-space tabs for consistency with other files

tests/testthat/test-derive.R
```diff
@@ -1,8 +1,52 @@
...
+    })
+
+    test_that("derivation()<-NULL and is.derived()<-FALSE issues a PATCH",{
+        expect_PATCH(derivation(birthyrPlus) <- NULL,
+                     'https://app.crunch.io/api/datasets/1/variables/birthyrPlus/',
+                     '{"derived":false}')
+        expect_PATCH(is.derived(birthyrPlus) <- FALSE,
+                     'https://app.crunch.io/api/datasets/1/variables/birthyrPlus/',
+                     '{"derived":false}')
+    })
```
nealrichardson: `expect_no_request` on the ones that shouldn't PATCH (e.g. if setting TRUE on a derived variable, if setting FALSE on a non-derived variable, if setting NULL on a non-derived variable)

and what happens if you try to set `is.derived(var) <- TRUE` on a variable that is not derived? that should error

R/variable-derivation.R
```diff
@@ -0,0 +1,124 @@
+#' Get or set a derived variable's `CrunchExpr`
```
nealrichardson: just say "expression" here and not CrunchExpr. And I think that the first line of docs like this is the title and thus should be Title Cased (yes, I know the current docs aren't consistent about that)

R/variable-derivation.R
```diff
@@ -0,0 +1,124 @@
...
+#' `is.derived` can be used to see if a variable is derived or not. Additionally
+#' setting a derived variable's `is.derived` to `FALSE` will integrate the
+#' derived variable.
+#'
+#' @param x a variable
+#' @param value a `CrunchExpr` to be used as the derivation (for the setter
+#' only) or `NULL` to integrate a derived variable. For `is.derived`, `FALSE`
+#' can be used to integrate a derived variable.
+#'
+#' @return a `CrunchExpr` of the derivation
```
nealrichardson: Update this to say what is.derived and is.derived<- return. Also, derivation<- doesn't return a CrunchExpr

R/variable-derivation.R
```diff
@@ -0,0 +1,124 @@
...
+#' # Crunch expression: v1 + 5
+#'
+#' derivation(ds$derived_v1) <- ds$v1 + 10
+#' derivation(ds$derived_v1)
+#' # Crunch expression: v1 + 10
+#'
+#' # to integrate or instantiate the variable in place (remove the link between
+#' # variable v1 and the derivation) you can:
+#' derivation(ds$derived_v1) <- NULL
+#' }
```
nealrichardson: Work is.derived into these examples

inst/app.crunch.io/api/datasets/1/variables/birthyrPlus.json
jonkeane: Ooops that's what I get for switching to atom for json editing!

jonkeane:
I've merged in my changes, this is ready for another round of review.

R/variable-derivation.R
```diff
@@ -0,0 +1,136 @@
...
+        halt("The value ", dQuote(substitute(value)), " must be a CrunchExpr, got a ",
+             class(value), " instead.")
+    }
+
+    payload <- toJSON(list(derivation=value@expression))
+    crPATCH(self(x), body=payload)
+
+    ## Refresh and return
+    dropCache(datasetReference(x))
+    return(invisible(refresh(x)))
```
nealrichardson: I believe "return" and "invisible" are redundant/opposites. Pick one.

R/variable-derivation.R
```diff
@@ -0,0 +1,136 @@
+#' Get or set a derived variable's expression
+#'
+#' Get a derived variable's derivation formula as a (`CrunchExpr`)[expressions] with
+#' `derivation([variable])`. Set (change) a derived variable's derivation with
+#' `derivation([variable]) <- [expression]`.
+#'
+#' To integrate (aka realize or instantiate) a variable so that it is no longer
+#'  dependent on the variables the derivation relies on, use
```
nealrichardson: "dependent on the variables the derivation relies on" reads awkwardly

PR #101, opened 2017-08-14
Generate weight variable from categorical variables

GShotwell:
`makeWeight` allows the user to specify a categorical variable and the target weighting categories using a formula `ds$var ~ c(30, 30, 40`. The function generates a Variable Definition which uses the server function `rake` to actually create the weighting.


tests/testthat/test-weighting.R
```diff
@@ -124,5 +152,12 @@ with_test_authentication({
                 array(c(110, 100), dim=2L, dimnames=list(v4=c("B", "C"))))
         })
     })
-
+with(test.dataset(df), {
+        test_that("makeWeight returns the expected weights", {
+            ds$weight <- makeWeight(ds$v4 ~ c(30, 70, 0), name = "weight")
+            expected_weights <- c(0.6, 1.4, 0.6, 1.4, 0.6, 1.4, 0.6, 1.4, 0.6, 1.4, 0.6, 1.4,
+                                  0.6, 1.4, 0.6, 1.4, 0.6, 1.4, 0.6, 1.4)
```
jonkeane: This is surely right, but I'm surprised that these are the weights given that everything is divided by 100 at L151 in weights.R

R/weight.R
```diff
@@ -59,3 +59,104 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' @param name
+#' The name of the resulting variable
+#' @rdname makeWeight
+#' @return
+#' A crunch Variable Definiton of the weight variable
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' cars <- mtcars
```
GShotwell: Whoops, this was just there to bring `mtcars` into the global environment, but I see that coercing to a factor does that already.

tests/testthat/test-weighting.R
```diff
@@ -34,11 +34,39 @@ with_mock_crunch({
...
+        with(temp.option(crunch.namekey.dataset ="name"), {
             expect_identical(weightVariables(newds), "Birth Year")
         })
         expect_identical(weightVariables(oldds), c())
     })
+
+    test_that("validateWeightExpression errors correctly", {
+        expect_error(validateWeightExpression("bad_formula"),
+            "bad_formula is an invalid expression, use the form ds$var ~ c(10, 20, 30)", fixed = TRUE)
+        expect_error(validateWeightExpression(oldds$birthyr ~ c(1,2,3)),
```
GShotwell: Definitely, good catch

tests/testthat/test-weighting.R
GShotwell: Oh, the thing which is divided by 100 are the targets, just so the user can type ``c(50, 50)` rather than `c(.5,.5)`. The expected weights here are what comes off the server after the raking has been done.

R/weight.R
```diff
@@ -59,3 +59,106 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    }
+    var     <- eval(expr[[2]], environment(expr))
+    varname <- deparse(expr[[2]])
+    targets <- eval(expr[[3]])
+    n_categories <- length(categories(var))
+    if (n_categories == 0) {
+        halt(paste(varname, "is not a categorical crunch variable"))
+    }
+    if (length(targets) != n_categories) {
+        halt(paste("Number of targets does not match number of categories for", varname))
```
nealrichardson: We should consider padding with 0s if you don't send all of the categories, esp. if the  first `length(targets)` categories contain all of the valid categories, and the remaining `n_categories - length(targets)` categories are all missing categories. It's going to be annoying to always have to type a 0 for the No Data category.

R/weight.R
```diff
@@ -59,3 +59,106 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' Utility function to generate entries for makeWeight
+#'
+#' @param expr
+#' An expression
+#' @keywords internal
+#' @return
+#' A list of the variable id and the target weights.
+#'
+#'
+generateWeightEntry <- function(expr) {
```
nealrichardson: How about merging `validateWeightExpression` and this function? I.e. instead of returning `expr` from validate, add L155-163 from here and return the API expression? We should avoid evaluating the expression twice.

R/weight.R
```diff
@@ -59,3 +59,106 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+        halt(paste(varname, "is not a categorical crunch variable"))
+    }
+    if (length(targets) != n_categories) {
+        halt(paste("Number of targets does not match number of categories for", varname))
+    }
+    if (!all(is.numeric(targets))) {
+        halt(paste("Targets are not numeric for", varname))
+    }
+    if (sum(targets) != 100) {
+        halt(paste("Targets do not add up to 100% for", varname))
```
nealrichardson: The API expects that `targets` will sum to 1. I'd allow that here too. So particularly if you follow my suggestion and merge this function with the `generateWeightEntry`, you'd do something like check if targets sum to 100, and if so, divide by 100, and then assert that they sum to 1.

Also, I believe `stop` (which `halt` calls) effectively does `paste0` inside, so you can skip calling `paste` in the error messages (and add the extra whitespace explicitly)

R/weight.R
```diff
@@ -59,3 +59,106 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' NULL value, function is called for its side effects
+#'
+#' @keywords internal
+#' @examples
+#' \dontrun{
+#' validateWeightExpression(ds$var ~ c(10, 30, 60))
+#' }
+#'
+validateWeightExpression <- function(expr) {
+    if (length(expr) != 3) {
```
nealrichardson: This should be ensuring that the expression is a formula and should use R's formula-parsing utilities, I think. cf. https://github.com/Crunch-io/rcrunch/blob/master/R/formula.R#L11-L17

R/weight.R
```diff
@@ -59,3 +59,106 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+        halt(paste(expr, "is an invalid expression, use the form ds$var ~ c(10, 20, 30)"))
```
nealrichardson: And the error message should probably say "formula" in it somewhere since that is the requirement.

tests/testthat/test-weighting.R
nealrichardson: The other e2e assertion you could make is that `sum(ds$weight) == nrow(ds)`

R/weight.R
jonkeane: `stop` and thus `halt` does indeed call `paste0`

R/weight.R
jonkeane: I totally agree with this here

GShotwell: Oh right that makes a lot of sense.

R/weight.R
GShotwell: Okay I've changed it so that the user can supply `c(30, 30, 40)` or `c(.3, .3, .4)` is that what you were after?

nealrichardson: Yes, thank you

GShotwell:
@nealrichardson @jonkeane This is ready for review

tests/testthat/test-weighting.R
```diff
@@ -66,6 +66,8 @@ with_mock_crunch({
     test_that("makeWeight generates the expected VariableDefinition", {
         expect_equivalent(makeWeight(oldds$gender ~ c(20, 30, 50), name = "weight"),
             expected_weight_definition)
+        expect_equivalent(makeWeight(oldds$gender ~ c(.2, .3, .5), name = "weight"),
+            expected_weight_definition)
```
jonkeane: It would be nice to also check that the 0-padding does the right thing here.

R/weight.R
```diff
@@ -100,57 +99,54 @@ makeWeight <- function(..., name) {
...
 #' validateWeightExpression(ds$var ~ c(10, 30, 60))
 #' }
 #'
-validateWeightExpression <- function(expr) {
-    if (length(expr) != 3) {
-        halt(paste(expr, "is an invalid expression, use the form ds$var ~ c(10, 20, 30)"))
+generateWeightEntry <- function(expr) {
+    formula <- try(as.formula(expr), silent = TRUE)
+    if (is.error(formula)) {
+        halt(expr, " is not a valid formula, use the form ds$var ~ c(10, 20, 30)")
```
jonkeane: I wonder if you should do `substitute(expr)` here (possibly with an `as.character()` around it, just in case (and also possibly adding in a test using a non-character object to the tests below. This way the user sees what they typed.

R/weight.R
```diff
@@ -161,4 +157,4 @@ generateWeightEntry <- function(expr) {
         variable = self(var),
```
jonkeane: This should be explicitly `return`ed

GShotwell:

R/weight.R
```diff
@@ -153,8 +153,9 @@ generateWeightEntry <- function(expr) {
         target_list[[i]] <- c(i, targets[i])
     }
 
-    list(
+    out <- list(
         variable = self(var),
         targets = target_list
     )
+    out
```
jonkeane: This should be `return(out)`

GShotwell: Can you say more about the reasoning? Do we always require an explicit return statement?

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
         return(c())
     }
 })
+
+#' Generate a weight vector
```
nealrichardson: "variable", not "vector". This is creating an expression for a derived variable.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#'
+#' This function allows you to generate a weight vector by supplying a set of
+#' categorical variables and the target distribution for each of the variable's categories.
+#' For instance if you wanted to create a weight variable which equally weighted four categories stored
+#' in `ds$var` you would call `makeWeight(ds$var ~ c(25, 25, 25, 25), name = "weight")`.
```
nealrichardson: Somewhere in the docs we should say something about "rake" or "rim weighting", the methodology used. @malecki can provide the right words.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#'
+#' @param ...
+#' A series of expressions of the form `variable ~ target_weights`. The variable must
+#' be a categorical crunch variable, and the target weights must be a numeric vector whose
```
nealrichardson: "Crunch" should be capitalized.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' length is equal to the number of categories contained in the variable, and whose sum is equal to 100.
```
nealrichardson: or equal to 1.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#'
```
nealrichardson: `...` should also allow other variable metadata to be provided. See https://github.com/Crunch-io/rcrunch/blob/master/R/case-variables.R for a pattern to follow.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' @param name
+#' The name of the resulting variable
+#' @rdname makeWeight
+#' @return
+#' A crunch Variable Definiton of the weight variable
```
nealrichardson: `[VariableDefinition()]`

nealrichardson: (to link to the man page for that)

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' }
+#'
+generateWeightEntry <- function(expr) {
+    formula <- try(as.formula(expr), silent = TRUE)
+    if (is.error(formula)) {
+        halt(dQuote(substitute(expr)), " is not a valid formula, use the form ds$var ~ c(10, 20, 30)")
+    }
+    var     <- eval(expr[[2]], environment(expr))
+    varname <- deparse(expr[[2]])
+    targets <- eval(expr[[3]])
```
nealrichardson: no envir needed for this eval?

R/weight.R
nealrichardson: Should also document that if there are fewer than the number of cats given, the rest will be taken to be 0.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    if (length(targets) < n_categories) {
+        targets <- c(targets, rep(0, n_categories - length(targets)))
+    }
+
+    target_list <- vector("list", length(targets))
+    for (i in seq_along(targets)) {
+        target_list[[i]] <- c(i, targets[i])
+    }
+
+    out <- list(
```
nealrichardson: You can just `return(list(variable=...` here

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    if (sum(targets) != 1) {
+        targets <- targets / 100
+    }
+
+    #Pad with zeros if the user hasn't suppied enough categories
+    if (length(targets) < n_categories) {
+        targets <- c(targets, rep(0, n_categories - length(targets)))
+    }
+
+    target_list <- vector("list", length(targets))
```
nealrichardson: How about just `target_list <- lapply(seq_along(targets), function (i) c(i, targets[i]))` ?

tests/testthat/test-weighting.R
```diff
@@ -34,11 +34,48 @@ with_mock_crunch({
 
     test_that("weightVariables method", {
         expect_identical(weightVariables(newds), "birthyr")
-        with(temp.option(crunch.namekey.dataset="name"), {
+        with(temp.option(crunch.namekey.dataset ="name"), {
```
nealrichardson: Odd space addition.

tests/testthat/test-weighting.R
```diff
@@ -124,5 +161,14 @@ with_test_authentication({
                 array(c(110, 100), dim=2L, dimnames=list(v4=c("B", "C"))))
         })
     })
-
+with(test.dataset(df), {
+        test_that("makeWeight returns the expected weights", {
```
nealrichardson: Can't tell if this is just github, but this block looks overly indented.

nealrichardson: Or rather, the `with` should be indented but is not.

tests/testthat/test-weighting.R
```diff
@@ -34,11 +34,48 @@ with_mock_crunch({
...
+        derivation = list(`function` = "rake",
+            args = list(
+                list(variable = "https://app.crunch.io/api/datasets/1/variables/gender/",
+                    targets = list(c(1, 0.2), c(2, 0.3), c(3, 0.5)
+                    )
+                )
+            )
+        )
+    )
+    test_that("makeWeight generates the expected VariableDefinition", {
```
nealrichardson: Could you break these tests into separate `test_that`s that have more informative descriptions? would make it easier to comprehend what exactly the test is asserting (and thus easier to interpret should it fail).

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    }
+    if (!(sum(targets) == 100 || sum(targets) == 1)) {
+        halt("Targets do not add up to 100% for ", varname)
+    }
+    if (sum(targets) != 1) {
+        targets <- targets / 100
+    }
+
+    #Pad with zeros if the user hasn't suppied enough categories
+    if (length(targets) < n_categories) {
```
nealrichardson: Move this up to where you check that too many targets aren't given. Will make the function read more naturally.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    if (!all(is.numeric(targets))) {
+        halt("Targets are not numeric for ", varname)
+    }
+    if (!(sum(targets) == 100 || sum(targets) == 1)) {
+        halt("Targets do not add up to 100% for ", varname)
+    }
+    if (sum(targets) != 1) {
+        targets <- targets / 100
+    }
+
```
nealrichardson: Are there any other bad-input conditions you might be concerned about? What if the targets are length 0? What if there are `NA`s in `targets`?

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' A crunch Variable Definiton of the weight variable
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' mtcars$cyl <- as.factor(mtcars$cyl)
+#' mtcars$gear <- as.factor(mtcars$gear)
+#' ds <- newDataset(mtcars, "mtcars")
+#' ds$weight <- makeWeight(ds$cyl ~ c(30, 30, 40, 0), ds$gear ~ c(20, 20, 60, 0), name = "weight" )
+#' as.vector(ds$weight)
```
nealrichardson: how about `summary(ds$weight)` instead? we generally should encourage our users to do things that evaluate on the server rather than download column data (i.e. what `as.vector` does).

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' @rdname makeWeight
+#' @return
+#' A crunch Variable Definiton of the weight variable
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' mtcars$cyl <- as.factor(mtcars$cyl)
+#' mtcars$gear <- as.factor(mtcars$gear)
+#' ds <- newDataset(mtcars, "mtcars")
```
nealrichardson: In `newDataset`, if no name is provided, it will grab the name of the data.frame you passed in, so technically you don't need to specify "mtcars" as the name.

R/weight.R
```diff
@@ -59,3 +59,103 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' @keywords internal
+#' @examples
+#' \dontrun{
+#' validateWeightExpression(ds$var ~ c(10, 30, 60))
+#' }
+#'
+generateWeightEntry <- function(expr) {
+    formula <- try(as.formula(expr), silent = TRUE)
+    if (is.error(formula)) {
+        halt(dQuote(substitute(expr)), " is not a valid formula, use the form ds$var ~ c(10, 20, 30)")
```
nealrichardson: after "formula", it would be grammatically better to put a period and start a new sentence with "Use".

R/weight.R
GShotwell: Target of length 0 is going to be non-numeric so I think that one's caught. NA's produce an error, but I should probably just infer that they are zero.

nealrichardson: It's technically possible though less likely for a user to accidentally produce a zero-length numeric vector than include NAs:

```
> x <- numeric(0)
> is.numeric(x)
[1] TRUE
> length(x)
[1] 0
```

And I think you should raise an error if there are NA targets because it's probably not intentional.

GShotwell:
@nealrichardson This is ready for review

R/weight.R
```diff
@@ -24,7 +25,11 @@ weight <- function (x) {
 
 #' @rdname weight
 #' @export
-`weight<-` <- function (x, value) {
+`weight<-` <- function(x, value) {
+    if (class(value) == "VariableDefinition") {
```
nealrichardson: `inherits` is better than `class` for this checking.

R/weight.R
```diff
@@ -24,7 +25,11 @@ weight <- function (x) {
...
+        x <- addVariables(x, value)
+        value <- x[[value$name]]
+    }
     stopifnot(is.dataset(x))
```
nealrichardson: The VariableDefinition if block should happen after the `stopifnot`

R/weight.R
```diff
@@ -59,3 +64,111 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
         return(c())
     }
 })
+
+#' Generate a weight variable
+#'
+#' This function allows you to generate a weight vector by supplying a set of
```
nealrichardson: "variable" not "vector"

R/weight.R
```diff
@@ -59,3 +64,111 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' categorical variables and the target distribution for each of the variable's categories. Weights are
+#' computed by iteratively 'raking' conditional 'cells' to marginal population targets.
+#' For instance if you wanted to create a weight variable which equally weighted four categories stored
+#' in `ds$var` you would call `ds$weight1 <- makeWeight(ds$var ~ c(25, 25, 25, 25), name = "weight1")`. This would
+#' create a new variable which can then be set as the dataset weight with [weight]. You can also create
```
nealrichardson: is `[weight]` supposed to be a markdown link to somewhere?

R/weight.R
```diff
@@ -59,3 +64,111 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    expr_list <- all_dots[!named_entries]
+
+    #args below must be an unnamed list, so we remove the names from expr_list
+    names(expr_list) <- NULL
+    out$derivation <- list(
+            `function` = "rake",
+            args = lapply(expr_list, generateWeightEntry)
+        )
+    class(out) <- "VariableDefinition"
+    out
```
nealrichardson: `return(out)`

R/weight.R
```diff
@@ -59,3 +64,111 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    formula <- try(as.formula(expr), silent = TRUE)
+    if (is.error(formula)) {
+        halt(dQuote(substitute(expr)), " is not a valid formula. Use the form ds$var ~ c(10, 20, 30)")
+    }
+    var     <- eval(expr[[2]], environment(expr))
+    varname <- deparse(expr[[2]])
+    targets <- eval(expr[[3]], environment(expr))
+
+    if (!is.Categorical(var)) {
+        halt(varname, " is not a categorical crunch variable")
```
nealrichardson: Crunch should be capitalized.

R/weight.R
```diff
@@ -59,3 +64,111 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    }
+    #Pad with zeros if the user hasn't suppied enough categories
+    if (length(targets) < n_categories) {
+        targets <- c(targets, rep(0, n_categories - length(targets)))
+    }
+    if (!all(is.numeric(targets))) {
+        halt("Targets are not numeric for ", varname)
+    }
+    if (any(is.na(targets))) {
+        halt(dQuote(paste(expr[2], "~", expr[3])), " contains NA values.")
```
nealrichardson: `paste(capture.output(print(expr)), "contains NA values.")`

R/weight.R
```diff
@@ -59,3 +64,111 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' @examples
+#' \dontrun{
+#' validateWeightExpression(ds$var ~ c(10, 30, 60))
+#' }
+#'
+generateWeightEntry <- function(expr) {
+
+    formula <- try(as.formula(expr), silent = TRUE)
+    if (is.error(formula)) {
+        halt(dQuote(substitute(expr)), " is not a valid formula. Use the form ds$var ~ c(10, 20, 30)")
```
nealrichardson: Minor point, but perhaps in the example `ds$var ~ c(10, 20, 30)`, the percentages should add to 100 as is required.

R/weight.R
nealrichardson: There's also `deparse` instead of `capture.output(print())`

Issue #103, opened 2017-08-17
changeCategoryID causing mismatched category $values


zach-zussman:
The more recently implemented version of changeCategoryID has this line which was not present in the originally used function:
newcat$numeric_value <- to
but it doesn't seem to check if the variable is 'missing' or not.   This makes it so using changeCategoryID on for example a streaming dataset to change categories to match an uploaded dataset, you end up with mismatched category values.  The uploaded dataset by default has NA for values for missing categories and the streamed-and-changed dataset will have the id #'s as values for any missing categories.

jonkeane:
We've just merged some code that updates this behavior. Now, the numeric value of a category will not be changed unless the id and the numeric value are the same before calling `changeCategoryID()`.

PR #104, opened 2017-08-23
Fixed `changeCategoryID()` to not always update the numeric value of the category.

jonkeane:
Added tests of numeric value changes when using `changeCategoryID()` `changeCategoryID()` no longer always makes the numeric value the same as the new ID. This addresses issue #103

### DIFF
--- a/R/categories.R
+++ b/R/categories.R
@@ -254,7 +254,9 @@ setMethod("lapply", "Categories", function (X, FUN, ...) {
 #'
 #' Changes the id of a category from an existing value to a new one.
 #' The variable can be a categorical, categorical array, or multiple response
-#' variable.
+#' variable. The category changed will have the same numeric value and missing
+#' status as before. The one exception to this is if the numeric value is the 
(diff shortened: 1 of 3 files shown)


R/categories.R
```diff
@@ -289,9 +291,11 @@ changeCategoryID <- function (variable, from, to) {
 
     ## Add new category
     newcat <- categories(variable)[[pos.from]]
+    if (newcat$id == newcat$numeric_value %||% FALSE) {
+        newcat$numeric_value <- to
```
GShotwell: I wonder if we should have a warning here. I think this is probably the best solution but it's still somewhat counterintuitive behaviour so it might be good to have a warning. It probably is also a good idea to put a comment explaining why this change is here.

nealrichardson: Agree that a code comment is good, but I don't think a user warning is necessary. If users have set their ids and numeric_values to be the same, they'll expect them to remain the same.

Issue #106, opened 2017-08-30
as.data.frame ids


persephonet:
as.data.frame has been working really well! It would be amazing if it had the categorical='id' option that exportDataset has! Then I would remove exportDataset (and write.csv) from my scrips

persephonet:
Additionally, as.data.frame doesn't currently seem to allow hidden variables.

persephonet:
Another note: as.data.frame does not return the variables in the order you asked for them

jonkeane:
Most of these should be addable pretty easily. For the ordering of variables, were you wanting a syntax like `as.data.frame(ds[c("var5", "var2", "var1")])` to produce a CrunchDataFrame with the columns var5, var2, var1 in that order?

PR #110, opened 2017-09-05
Crunch data frame improvements

jonkeane:
This addresses all but the ordering point of issue #106 plus more `data.frame` like methods (`[`, `[[`, `$`, plus setters for each)


nealrichardson:
On Fri, Sep 8, 2017 at 9:32 AM, Gordon Shotwell <notifications@github.com>
wrote:

> *@GShotwell* commented on this pull request.
>
> These are some kind of big picture things, and I'll hold off on code level
> things (which generally seems good to me) until we figure this stuff out:
>
> I think that the CrunchDataFrame is fundamentally a pretty confusing
> structure in that it's partially a local R dataframe, and partially a
> remote crunchDataset.
>
It's not really hybrid. It uses lazy evaluation to pull data from the
server on demand. But any time you access a variable in a CDF, you're
accessing R vectors.

> This isn't to say that it's a bad datastructure, but it just has some
> potential to be confusing. For example:
>
>    1. It seems odd that calling as.dataframe on a CrunchDataset doesn't
>    produce a dataframe, but instead produces an environment with class
>    CruchDataFrame. My typical expectation as a user is that as.data.frame
>    produces a dataframe. Are there any other packages which create hybrid
>    objects similar to this one? Or others where an object is actually an
>    environment?
>
> If it behaves like a data.frame, why does it matter?

How do you characterize this as different from and more confusing than a
sql backend for dplyr?

>
>    1. Investigating the cloud portion of the dataframe produces confusing
>    results. For instance if I have:
>
> ds <- newDataset(mtcars)
> ds <- as.data.frame(ds)
> class(ds$mpg)
>
> It will look like ds$mpg is a regular ol' variable, but in reality it's a
> CrunchVariable,
>
No, ds$mpg is a regular ol' R vector. You overwrote `ds` with the CDF, and
when you access `mpg`, you get an R vector.

> and I can't overwrite it with ds$mpg <- test.
>
You should be able to. If you can't, we're missing a `$<-` method.

> The same is true for print is.vector etc.
> 3) Filter then subset works ds <- ds[1:5, ]; ds[, "mpg"] but ds[1:5,
> "mpg"] doesn't work. For someone used to dataframes these two expressions
> seem identical.
>
So that method should be added. That's in the scope of this issue/PR.

> 4) Class conversion doesn't work because of the overwriting issue ds$cyl
> <- as.numeric(ds$cyl)
>
If `$<-` works then that will work too.

>
> 5) Some functions which work on dataframes won't work on CrunchDataFrames,
> like t()
>
There is no reasonable application for transposing a data.frame. If you're
transposing a data.frame, the structure you have shouldn't be a data.frame.

> I'm not sure about the best solution to these points of confusion, but my
> thought would be something like this:
>
>    - Add a vignette which tells people why we have two dataframe-y
>    structures, and when they should use one or the other. I think this would
>    also help us figure out if CDFs are doing what they should do
>    - Make CDFs unidirectional, so data goes from Crunch to CDF, but never
>    the other way around (I think this is how they work now, but we should make
>    that explicit)
>    - Whenever a variable in a CDF is accessed whether to read or write,
>    evaluate the promise and replace the variable in the CDF with the actual
>    vector, (so if you called as.numeric(ds$cyl) cyl in the CDF. This
>    would save API calls because you would only be pulling the variable the
>    first time, and also
>
> This is how CDFs work.

>
>    - Allow cloud variables in a CDF to be overwritten, in which case they
>    would changed from CrunchVariables to regular R vectors
>
> As said above, this should already be true, and if not, is in scope for
this issue.

>
>    - Add a message which tells the user when a variable is being
>    converted from CrunchVariables to vectors, I think this would be good
>    because it explains why calling a function on a variable can take some time.
>
> I disagree that messaging is needed--we don't message elsewhere when we
touch the server--but if others feel strongly, we could enable a verbose
mode or something.

jonkeane:
Here are my replies to these issues:


> It seems odd that calling as.dataframe on a CrunchDataset doesn't produce a dataframe, but instead produces an environment with class CruchDataFrame. My typical expectation as a user is that as.data.frame produces a dataframe. Are there any other packages which create hybrid objects similar to this one? Or others where an object is actually an environment?
We can get rid of it looking like an environment pretty easily, it was originally there for convenience, and something I can work on. As for getting a data.frame-like thing that's not truly a data.frame, I'm not super super worried about that, it's somewhat similar to the way that a lot of the tidyverse packages take data.frames and return tibbles without any warning.

> Investigating the cloud portion of the dataframe produces confusing results. For instance if I have:
> ds <- newDataset(mtcars)
> ds <- as.data.frame(ds)
> class(ds$mpg)
> It will look like ds$mpg is a regular ol' variable, but in reality it's a CrunchVariable, and I can't overwrite it with ds$mpg <- test. The same is true for print is.vector etc.
It's definitely possible (and pretty easy) to add this as a possibility. I original prevent overwriting of Crunch variables in case we wanted to go the route where there was more of a connection between CDFs and the server. It sounds from your and neal's comments that we're aiming more towards unidirectional / separation in which case overwriting ability is totally fine.

> 3) Filter then subset works ds <- ds[1:5, ]; ds[, "mpg"] but ds[1:5, "mpg"] doesn't work. For someone used to dataframes these two expressions seem identical.
This should work. It's possible I had a bug in the first set of code that made this not work, but when I tested it locally it worked for me:
```
[crunch] > ds <- newDataset(mtcars)
[crunch] > ds <- as.data.frame(ds)
[crunch] > ds[1:5, "mpg"]
[1] 21.0 21.0 22.8 21.4 18.7
```

> 4) Class conversion doesn't work because of the overwriting issue ds$cyl <- as.numeric(ds$cyl)
This will happen for free if we allow for overwritting of crunchvariables

> 5) Some functions which work on dataframes won't work on CrunchDataFrames, like t()
It would be good to have a list of what functions don't work and we can prioritize implementing them (since it should be fairly straightforward to do, once the CDF structure is set). I agree that supporting `t()` is probably pretty low priority for now.

> 
> I'm not sure about the best solution to these points of confusion, but my thought would be something like this:
> 
> Add a vignette which tells people why we have two dataframe-y structures, and when they should use one or the other. I think this would also help us figure out if CDFs are doing what they should do
> Make CDFs unidirectional, so data goes from Crunch to CDF, but never the other way around (I think this is how they work now, but we should make that explicit)
> Whenever a variable in a CDF is accessed whether to read or write, evaluate the promise and replace the variable in the CDF with the actual vector, (so if you called as.numeric(ds$cyl) cyl in the CDF. This would save API calls because you would only be pulling the variable the first time, and also
I changed the way that CDFs work so that there is a call to the variable whenever it is accessed (relying on caching to help us), but it shouldn't be hard to change this to actually bring the full variable down when it is accessed and maintaining it as a local variable going forward. I didn't do this at first for the same reason I didn't implement overwriting CrunchVariables: if we wanted to keep a closer connection between CDFs and the server we wouldn't necessarily want to have that happen. With a less close connection (especially a uni-directional one) having variables instantiate on access is a good plan.

> Allow cloud variables in a CDF to be overwritten, in which case they would changed from CrunchVariables to regular R vectors
> Add a message which tells the user when a variable is being converted from CrunchVariables to vectors, I think this would be good because it explains why calling a function on a variable can take some time.
I think I would prefer a deep-divey vignette to a warning here. I know many novice R users who would see a warning and think they've done something wrong, even if it says things like "this is normal, we are just letting you know..."

jonkeane:
I've pushed a few changes. The major one is that I've reverted to the behavior where, when a variable is accessed it is instantiated and will no longer call to the server if it's accessed again. This means the tracking of which variables are Crunch and which are local is gone as well. I've made some changes to the way that merge works that makes it a bit cleaner (I actually re-create the CDF when merging in order to separate it from the old version, this lets me also pass over the non-Crunch variables in the correct order for the output).

To recap what was already discussed in slack: RStudio turns out to "help" by touching the variables when doing things like tab completion. For now, I'm leaving that behavior as is. It's something that won't happen in scripts, just in an interactive session with RStudio. I think this addresses most of the comments above, and should be in a good position to have a more detailed code review.

R/as-data-frame.R
```diff
@@ -89,17 +98,20 @@ as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
 #' @rdname dataset-to-R
 #' @export
 as.data.frame.CrunchDataFrame <- function (x, row.names = NULL, optional = FALSE, ...) {
-    x <- x$.crunchDataset
+    ds <- attr(x, "crunchDataset")
     default.stringsAsFactors <- function () FALSE
     limit <- min(c(10000, getOption("crunch.data.frame.limit")))
-    if (nrow(x) * ncol(x) > limit) {
+    if (nrow(ds) * ncol(ds) > limit) {
         ## TODO: switch to downloading CSV and reading that?
```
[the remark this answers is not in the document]
jonkeane: I agree, though in the interest of moving forward on other projects, and since this is a pretty large PR already. I'm going to vote on putting this as a new issue/improvement to be implemented in a different PR. It would be good to have a clean DL csv/read into dataframe within `as.data.frame` as well as as a separate function.

R/as-data-frame.R
```diff
@@ -221,3 +236,217 @@ fix_bys <- function (data, by) {
...
+        col_name <- j[i]
+        x <- set_CDF_var(col_name = col_name, row_inds = row_inds,
+                         cdf = x, value = values[[i]])
+    }
+    return(x)
+}
+
+grab_col_names <- function (x, j) {
+    if (missing(j)) {
+        # if there's no j, grab all rows
```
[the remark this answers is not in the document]
jonkeane: Yeah, that was copy/pasting. It's fixed now.

jonkeane:
Excellent points about file organization. I'll work on that.

I'll work on adding to documentation and adding in some discussion about which functions work and which don't.

Automatic coercion would be really neat, but given the way that CDFs work now, there's not a good way to detect if the `delayedAssign` has been triggered yet or not, without also then triggering it itself. I looked at `pryr::is_promise` and that doesn't seem to actually work with `delayedAssign`

nealrichardson:
We could also consider an alternative implementation that doesn't use delayedAssign. `delayedAssign` makes the default `$` method automatically work, but if you're implementing `$` and `[` methods, maybe we can do it differently. I have some ideas we can discuss.

jonkeane:
I'm definitely interested in alternatives that we might have. I've tried a few different alternatives, and one of the biggest stumbling blocks I had was less with the `[` and `$` methods than it was with getting `eval` and `evalq` to work correctly (which is what it looks like `model.frame` and `lm` call.

What I've tried so far:
* list (which works with `eval`, although storing a promise as an element is not possible without additional dependencies, cf https://cran.r-project.org/web/packages/listenv/vignettes/listenv.html) Additionally we have the same RStudio problem as with environments.
* a mostly empty object with a list or environment attribute which stores the data, `eval` doesn't work hear. I thought about shimming or some other work around, but decided to not attempt that leap just yet.

Want to discuss other options at today's team meeting?

tests/testthat/test-crunch-data-frame.R
```diff
@@ -0,0 +1,184 @@
...
+        expect_equivalent(ds_df[["birthyr"]],true_df[["birthyr"]])
+    })
+    
+    test_that("Extract methods input checking", {
+        ds_df <- as.data.frame(ds)
+        
+        expect_error(ds_df["foo",],
+                     "row subsetting must be done with either numeric or a logical")
+        expect_error(ds_df[,list(list(1))], 
+                     "column subsetting must be done with either numeric or a character")
```
nealrichardson: I guess that's a better message than the data.frame one: 

```
> mtcars[,list(list(1))]
Error in .subset(x, j) : invalid subscript type 'list'
```

tests/testthat/test-crunch-data-frame.R
```diff
@@ -0,0 +1,184 @@
...
+        
+        expect_error(ds_df[c(TRUE),], 
+                     paste0("when using a logical to subset rows, the logical vector ",
+                            "must have the same length as the number of rows"))
+        expect_error(ds_df[,c(TRUE)],
+                     paste0("when using a logical to subset columns, the logical ",
+                            "vector must have the same length as the number of columns"))
+        
+        expect_error(`$`(ds_df,1),
+                     paste0("invalid subscript type 'double'"))
```
nealrichardson: Is this just asserting how R handles bad input on `$`?

tests/testthat/test-crunch-data-frame.R
```diff
@@ -0,0 +1,184 @@
+context("CrunchDataFrame Methods")
```
nealrichardson: Throughout this file (and everywhere), please put a space after a comma

tests/testthat/test-crunch-data-frame-merge.R
```diff
@@ -0,0 +1,168 @@
+context("CrunchDataFrame merging")
```
nealrichardson: Are these tests just moved from the other file or are there changes? Hard to tell from the diff.

tests/testthat/test-crunch-data-frame-dfconformance.R
```diff
@@ -0,0 +1,45 @@
+context("CrunchDataFrame conformance with data.frame methods")
```
nealrichardson: I don't particularly care for this file name. And I'm not sure the value of moving these couple of tests out of test-as-data-frame.R

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
+#' CrunchDataFrame
```
nealrichardson: I agree with malecki that `crdf` is a better shorthand for CrunchDataFrame. Please standardize on that.

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+    values <- split(value, ceiling(seq_along(value)/length(row_inds)))
+    for (i in seq_along(j)) {
+        col_name <- j[i]
+        x <- set_CDF_var(col_name = col_name, row_inds = row_inds,
+                         cdf = x, value = values[[i]])
+    }
+    return(x)
+}
+
+grab_col_names <- function (x, j) {
```
nealrichardson: Add some docstrings/comments for these internal functions. Their names aren't clear enough to explain why they exist (i.e. in this case, why you need a function other than "names()").

Also, the snake casing of these function names is a bit inconsistent with the naming conventions used in the package.

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+    return(get_CDF_var(col_name, x, mode = attr(x, "mode")))
+}
+
+#' @export
+#' @rdname CrunchDataFrame
+`[[<-.CrunchDataFrame` <- function (x, i, value) {
+    col_name <- col_name_from_index(x, i)
+    
+    x <- set_CDF_var(col_name = col_name, cdf = x, value = value)
+    return(x)
```
nealrichardson: Not that we're golfing here, but this could easily be a single line function body, as could the one above.

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+#' @export
+#' @rdname CrunchDataFrame
+`$.CrunchDataFrame` <- function (x, i) {
+    return(get_CDF_var(i, x, mode = attr(x, "mode")))
+}
+
+#' @export
+#' @rdname CrunchDataFrame
+`$<-.CrunchDataFrame` <- function (x, i, value) {
+    x <- set_CDF_var(col_name = i, cdf = x, value = value)
```
nealrichardson: likewise, this can be `return(setCrdfVar(...))` as a single line

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+    return(x)
+}
+
+get_CDF_var <- function (col_name, cdf, ...) {
+    if (!is(cdf, "CrunchDataFrame")) {
```
nealrichardson: `inherits` is preferred to `is`, isn't it?

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+#' @export
+#' @rdname CrunchDataFrame
+`$<-.CrunchDataFrame` <- function (x, i, value) {
+    x <- set_CDF_var(col_name = i, cdf = x, value = value)
+    return(x)
+}
+
+get_CDF_var <- function (col_name, cdf, ...) {
+    if (!is(cdf, "CrunchDataFrame")) { 
+        halt("The cdf argument must be a CrunchDataFrame, got ", class(cdf), " instead.")
```
nealrichardson: Since this function is not exported, do you really need to validate the input object type? Seems like overkill.

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+    }
+    if (!col_name %in% names(cdf)) {
+        halt("The variable ", dQuote(col_name), " is not found in the CrunchDataFrame.")
```
nealrichardson: Is this check required? How cryptic is the error returned by `get` if you let it handle whether the variable exists?

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+    
+    # validate row lengths
+    if (length(value) == 1 & length(row_inds) > 1) {
+        # if there is a single value, but more than one row, then we need to 
+        # repeat the value for as many times as there are rows. This is the 
+        # only exception to the rule that there must be the same number of 
+        # values as rows.
+        value <- rep(value, length(row_inds))
+    } else if (is.data.frame(value)) {
+        # we got a data.frame as values, likelythis is an array or MR varaible
```
nealrichardson: Spelling

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+        # unclass does what data.frame(v1=data.frame(a=1, b=2)) would do
```
nealrichardson: This comment doesn't make sense to me/doesn't seem to match the code. And/or you have left in commented code that should be deleted?

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+        # set multiple variables
+        value <- I(value)
+    } else if (length(value) != length(row_inds)) {
+        halt("replacement has ", length(value),
+             " rows, the CrunchDataFrame has ", nrow(cdf))
+    }
+    
+    # if we are replacing some rows, fill in with NAs, or grab old vector
+    if (length(row_inds) != nrow(cdf)) {
+        if (!col_name %in% names(cdf)) {
```
nealrichardson: I personally like to write this as `if (!(col_name %in% names(crdf)))` so that it's clear what is being negated (the whole thing, not `col_name`), despite the fact that it's technically not necessary due to order of operations.

R/crunch-data-frame.R
```diff
@@ -0,0 +1,292 @@
...
+            empty_vector <- rep(NA, nrow(cdf))
+            empty_vector[row_inds] <- value
```
nealrichardson: Looks like 2 of the 3 lines in each side of this if/else are the same, so you could collapse them.

tests/testthat/test-crunch-data-frame.R
```diff
@@ -0,0 +1,184 @@
...
+        expect_equivalent(ds_df[["catarray"]],true_df[["catarray"]])
+        expect_equivalent(ds_df[[2]],true_df[[2]])
+        expect_equivalent(ds_df[["birthyr"]],true_df[["birthyr"]])
+    })
+    
+    test_that("Extract methods input checking", {
+        ds_df <- as.data.frame(ds)
+        
+        expect_error(ds_df["foo",],
+                     "row subsetting must be done with either numeric or a logical")
```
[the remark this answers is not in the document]
nealrichardson: (Acknowledging, of course, that the first example is pointless, and the second one only works because the data.frame has row names set, which is not a feature I care to implement for crdfs right now.)

R/crunch-data-frame.R
jonkeane: It's not *super* cryptic, but it's not ideal since it references objects rather than columns or variables: `Error in get(col_name, crdf) : object 'fladskj' not found`

R/crunch-data-frame.R
jonkeane: I've harmonized this behavior with our discussion earlier. Array-like variables are now flattened.

R/crunch-data-frame.R
jonkeane: 👍 I used to almost always do ~this, but at some point have stopped. I'm happy to start again. This might be superstition at this point (and likely doesn't matter at all), but I've read that grouping expressions like this with `{` is actually slightly faster than `(` I also think in this circumstance, it's a bit more readable to have different kinds of braces for finding pairs. `(!{col_name %in% names(crdf)})` is just a bit easier to read for me than `(!(col_name %in% names(crdf)))` 

https://radfordneal.wordpress.com/2010/08/15/two-surpising-things-about-r/

tests/testthat/test-crunch-data-frame.R
jonkeane: It looks like there are a bunch of places where we need to fix this throughout the tests. Rather than send a massive change the includes a bunch of white space, I'll add this as an issue to go through and fix separately elsewhere. I've fixed the files I'm working on here.

tests/testthat/test-crunch-data-frame.R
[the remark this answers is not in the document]
jonkeane: We can't currently subset with `CrunchLogicalExpr` we could add that as a functionality, although that depends on the code that makes `as.vector` and `which` work with `CrunchLogicalExpr` which is already done in the Conditional Transforms branch.

As for the `data.frame`s response, that is because all data.frames have rownames (by default it's characters of the row indices (eg `c("1", "2", "3")`), and because `"foo"` isn't there, we get a row of `NA`s). 

We can mimic this, but we'll need to do add a bit more here, I would vote for deferring both of these until after we merge the conditional transforms branch since the code is in there that will allow `CrunchLogicalExpr` to be used already and then we can decide if we want to mimic `data.frame`s behavior with default rownames / ability to add rownames, etc.

tests/testthat/test-crunch-data-frame.R
jonkeane: Yeah, I wanted to make sure that we were covering that, and not getting something weird. Although we don't need to leave it in here since it really is just asserting that R does this for us, it might be nice to remind our future selves that we have checked that possibility.

R/crunch-data-frame.R
```diff
@@ -0,0 +1,291 @@
...
+#' @export
+#' @rdname CrunchDataFrame
+`[<-.CrunchDataFrame` <- function (x, i, j, value) {
+    j <- grabColNames(x, j)
+    row_inds <- grabRowInd(x, i)
+    
+    # TODO: check value length
+    
+    # split value string by the length of row indicators
+    values <- split(value, ceiling(seq_along(value)/length(row_inds)))
```
nealrichardson: I'm not following this. Can you explain?

R/crunch-data-frame.R
```diff
@@ -0,0 +1,291 @@
...
+#' @param allow_logical logical should logical subsetting be allowed? (default: TRUE)
+#' 
+#' @return character names of the columns of the crdf to select
+#' @keywords internal
+grabColNames <- function (x, j, allow_logical = TRUE) {
+    if (missing(j)) {
+        # if there's no j, grab all columns
+        j <- names(x)
+    } else if (is.logical(j)) {
+        # recycle logicals, grab only extant names
```
nealrichardson: why do you have to do this? how does `names(x)[j]`not just do the right thing?

R/crunch-data-frame.R
```diff
@@ -0,0 +1,291 @@
...
+        # data.frame(a=1, b=2)) would do.
+        value <- unlist(value, recursive = FALSE)
+    } else if (length(value) != length(row_inds)) {
+        halt("replacement has ", length(value),
+             " rows, the CrunchDataFrame has ", nrow(crdf))
+    }
+    
+    # if we are replacing some rows, fill in with NAs, or grab old vector
+    if (length(row_inds) != nrow(crdf)) {
+        if (!{col_name %in% names(crdf)}) {
```
nealrichardson: Radford Neal may be right about { being 6% faster than (, and given his surname I'm inclined to trust him, but visually I think that using { like this is an abomination. I'll take the few extra microseconds in exchange for the elegance of the parenthesis's smooth curve.

R/variable-to-R.R
```diff
@@ -148,6 +148,12 @@ paginatedGET <- function (url, query, offset=0, limit=1000, table=FALSE) {
 #' something beyond what is currently supported, you can bring a variable's
 #' data into R with `as.vector(ds$var)` and work with it like any
 #' other R vector.
+#' 
+#' `as.vector` transfers data from Crunch to a local R session. Note: 
+#' `as.vector` returns the vector in the row order of the dataset. If filters 
+#' are set that specify an order that is different from the row order of the 
+#' dataset, the results will ignore that order. If you need the vector ordered 
+#' in that way, use syntax like `as.vector(ds$var)[c(10, 5, 2)]` instead.
```
nealrichardson: It wouldn't be impossible to make `as.vector` do the right thing. `</scope creep>` (Feel free to make an issue for it.)

R/crunch-data-frame.R
jonkeane: That splits the values into a list that corresponds to each column. So it takes the values and splits them up into chunks that are the length of the number of rows that were specified. This is to match this pattern:

```
foo <- data.frame(foo=c(1, 2, 3, 4), bar=c(5, 6, 7, 8), baz = c(1, 2, 3, 4), qux = c(5, 6, 7, 8))
foo[c(1, 2), c(3, 4)] <- c(10, 20, 30, 40)
```

I should add in additional checks that ensure that there are exactly the right number of values  (currently that will break...)

nealrichardson: That's a really unnatural insertion. 

Can you please update the comment, perhaps using this example? "value string" isn't accurate, apparently, and "row indicators" is unclear. In reality, you're "splitting by row" because the base behavior assumes that you're inserting a matrix, which is column major.

PR #111, opened 2017-09-05
Deprecate `autorollback` and `savepoint`when appending to reflect the API change

### LINKED ISSUE #105: Remove the “autorollback” arg from `appendDataset`
The feature has been deprecated in the Crunch API (appends now always automatically roll back in case of failure).

jonkeane:
Resolves #105

### DIFF
--- a/R/append-dataset.R
+++ b/R/append-dataset.R
@@ -6,14 +6,16 @@
 #' dataset before appending. If it is a CrunchDataset, it may be subsetted with
 #' a filter expression on the rows and a selection of variables on the columns.
 #' @param autorollback Deprecated. This option no longer does anything.
-#' logical: If the append fails, revert the dataset back
-#' to its state before attempting to append? Default is \code{TRUE}, and you
-#' probably won't want to change that.
 #' @return `dataset1`, updated with `dataset2`, potentially filtered on rows and
 #' variables, appended to it.
 #' @export
 appendDataset <- function (dataset1, dataset2, autorollback) {
     stopifnot(is.dataset(dataset1))
+
+    if (!missing(autorollback)) {
+        warning("The ", sQuote("autorollback"),
+                " argument is deprecated and has no effect", call.=FALSE)
+    }
 
     if (!is.dataset(dataset2)) {
         temp.ds.name <- paste("Appending to", name(dataset1), now())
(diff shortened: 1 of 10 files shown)


R/append-dataset.R
```diff
@@ -5,14 +5,16 @@
 #' `dataset2` is not a Crunch dataset, it will be uploaded as a new
 #' dataset before appending. If it is a CrunchDataset, it may be subsetted with
 #' a filter expression on the rows and a selection of variables on the columns.
-#' @param autorollback logical: If the append fails, revert the dataset back
+#' @param autorollback Deprecated. This option no longer does anything.
+#' logical: If the append fails, revert the dataset back
```
nealrichardson: Remove this second line since it's no longer true

R/append-dataset.R
```diff
@@ -5,14 +5,16 @@
...
 #' to its state before attempting to append? Default is \code{TRUE}, and you
 #' probably won't want to change that.
 #' @return `dataset1`, updated with `dataset2`, potentially filtered on rows and
 #' variables, appended to it.
 #' @export
-appendDataset <- function (dataset1, dataset2, autorollback=TRUE) {
+appendDataset <- function (dataset1, dataset2, autorollback) {
     stopifnot(is.dataset(dataset1))
+
```
nealrichardson:
```
if (!missing(autorollback)) {
    warning("The 'autorollback' argument is deprecated and has no effect", call.=FALSE)
}
```

and test for that warning.

R/batches.R
```diff
@@ -1,22 +1,19 @@
-addBatch <- function (ds, ..., savepoint=TRUE, autorollback=savepoint, strict=TRUE, body=list(...)) {
+addBatch <- function (ds, ..., savepoint, autorollback, strict=TRUE, body=list(...)) {
```
nealrichardson: This function is not exported, so remove the arguments here, and handle the removal/deprecation warning in the calling function(s).

R/batches.R
```diff
@@ -1,22 +1,19 @@
...
-        ## Just execute and let the "Result URL" print if it fails
-        do_it <- force
+    if (!missing(savepoint)) {
+        warning("savepoint has been deprecated. This option no longer does anything.")
+    }
+    if (!missing(autorollback)) {
+        warning("autorollback has been deprecated. This option no longer does anything.")
     }
-    do_it(crPOST(batches_url, body=toJSON(body)))
+
```
nealrichardson: Retain the comment for why `suppressMessages` is here.

R/append-dataset.R
jonkeane: This is done, but GitHub's magic for hiding comments that are addressed was foiled by it being a line below.

jonkeane:
I've updated with these changes, so this is ready for another round.

PR #117, opened 2017-09-14
Streaming lock v2 mc 150052571

alejandro-rivera:


inst/app.crunch.io/api/datasets/1.json
```diff
@@ -6,7 +6,8 @@
     "body": {
         // In reality, body has attributes found in the tuple too
         "owner": "https://app.crunch.io/api/users/user1/",
-        "id": "dataset1"
+        "id": "dataset1",
+        "streaming": "streaming"
```
nealrichardson: The dataset that we use for all of the mock tests should not be "streaming". That's inconsistent with how we use it and the things we test against it. (Even if it obviously does nothing at this point.)

jonkeane: Would you prefer that we make a new separate set of streaming fixtures, or just remove the streaming property here, knowing that the fixture isn't real reflecting what the response would be?

nealrichardson: There are three datasets in the mock catalog. Pick a different one. Though it seems premature until we actually implement these features in the package.

jonkeane: We are currently using all three of the fixtures for unit testing the streaming abilities of the package in order to get the variety of states that we might see. (https://github.com/Crunch-io/rcrunch/blob/master/tests/testthat/test-dataset-stream.R#L14-L43)

nealrichardson: Ok. (Though we aren't using ds2 for much.) But we haven't implemented anything about this feature, which is schema protection for streaming datasets and the ability to alter that state. If dataset 1 is "streaming", then many of the actions we do to it in other tests (updating values, deleting variables, etc.) would be rejected by a real server.

jonkeane: Yeah, sounds like we should create new fixtures to stream to specifically to avoid confusion and diverging from what a real backend would allow. I'll work on that this evening / monday.

PR #119, opened 2017-09-15
Added arguments for altering color palettes of CrunchBoxes

jonkeane:


R/crunchbox.R
```diff
@@ -5,11 +5,16 @@
...
+#' @param brand_colors an optional list with colors (in hex format) named 'primary', 'secondary', and 'message'
+#' @param static_colors an optional list of static colors (in hex format) to use for categoricals (taken in order)
+#' @param category_color_lookup an optional list of category names to colors (in hex format) to use for that category (for fine grained tuning of category colors.)
 #' @param ... additional metadata for the box, such as "title", "header", etc.
 #' @return The URL to the newly created box.
 #' @seealso \code{\link{preCrunchBoxCheck}} to provide guidance on what you're including in the CrunchBox
 #' @export
-crunchBox <- function (dataset, filters=crunch::filters(dataset), ...) {
+crunchBox <- function (dataset, filters=crunch::filters(dataset),
+                       brand_colors, static_colors,
```
GShotwell: I think it would be helpful to have an example of changing the colour scheme. You can probably just copy the ones from the tests.

R/crunchbox.R
```diff
@@ -41,6 +46,28 @@ crunchBox <- function (dataset, filters=crunch::filters(dataset), ...) {
     ## Add "where" after so that it no-ops if variablesFilter returns NULL (i.e. no filter)
     payload$where <- variablesFilter(dataset)
 
+    ## Add colors if they exist to the payload
+    if (!missing(brand_colors)) {
+        if (!is.list(brand_colors) ||
+            any(!names(brand_colors) %in% c("primary", "secondary", "message"))) {
+            halt(sQuote("brand_colors"), " must be a named list with only ",
+                 serialPaste(dQuote(c("primary", "secondary", "message")), collapse = "or"))
```
GShotwell: Is there any way we can pass the colour scheme without a named list? For instance a character vector of length three which represented the three colours? I'm thinking of `ggplot2::scale_colour_manual()` as an example. Named lists can be a bit tricky for some users. Also a character vector would eliminate problems with  people naming the list incorrectly.

R/crunchbox.R
```diff
@@ -41,6 +46,28 @@ crunchBox <- function (dataset, filters=crunch::filters(dataset), ...) {
...
+        }
+        payload$display_settings$palette$brand_colors <- brand_colors
+    }
+    if (!missing(static_colors)) {
+        if (!is.list(static_colors)) {
+            halt(sQuote("static_colors"), " must be a list of characters")
```
GShotwell: Would a character vector be okay here? If so I think that's probably easier than a list. Also the error specifies that it be a list of character strings, but I don't think that's actually being checked.

R/crunchbox.R
```diff
@@ -41,6 +46,28 @@ crunchBox <- function (dataset, filters=crunch::filters(dataset), ...) {
...
+        }
+        payload$display_settings$palette$static_colors <- static_colors
+    }
+    if (!missing(category_color_lookup)) {
+        if (!is.list(category_color_lookup) || is.null(names(category_color_lookup))) {
```
GShotwell: Do we want to try to check that the categories supplied in the list match categories in the data? Or if there are mismatches on either side?

tests/testthat/test-crunchbox.R
```diff
@@ -111,12 +111,57 @@ with_mock_crunch({
...
         expect_error(crunchBox(), "'dataset' must be a CrunchDataset")
         expect_error(crunchBox(4), "'dataset' must be a CrunchDataset")
         expect_error(crunchBox(ds, 4),
             "'filters' should be a FilterCatalog or NULL")
-
+        expect_error(crunchBox(ds, brand_colors = "a"),
+                     paste0(
+                         sQuote("brand_colors"), " must be a named list with only ",
+                         serialPaste(dQuote(c("primary", "secondary", "message")),
+                                     collapse = "or")))
```
GShotwell: You might want to add some tests for non-characters being passed as arguments like number, NA, NULL

R/crunchbox.R
nealrichardson: 👍

R/crunchbox.R
nealrichardson: I think that might be too expensive (but perhaps something to add to the `preCrunchBoxCheck`?)

tests/testthat/test-crunchbox.R
nealrichardson: And that they're actually valid colors

R/crunchbox.R
nealrichardson: 👍

R/crunchbox.R
jonkeane: I've added this as a todo in precrunchbox check for now.

tests/testthat/test-crunchbox.R
jonkeane: Done and done.

PR #125, opened 2017-09-26
Allow setting of weightVariables

GShotwell:
Adds support for setting weightVariables similar to how the user can get and set the weight of a dataset.


DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Suggests:
 VignetteBuilder: knitr
 RoxygenNote: 6.0.1
 Roxygen: list(markdown = TRUE)
-Collate:
+Collate:
```
nealrichardson: Why?

R/weight.R
```diff
@@ -68,6 +89,49 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    }
+    current <- crGET(shojiURL(varCat, "orders", "weights"))
+    new <- current
+    if (is.null(value)) {
+        new$graph <- NULL
+    } else {
+        if (is.variable(value) || length(value) == 1) {
+            value <- list(value)
+        }
+        var_names <- vapply(value, function(x){
```
nealrichardson: You don't need this unless the input is invalid, so move this work inside the if at L118

R/weight.R
```diff
@@ -68,6 +89,49 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+            err_text <- " is not a numeric Crunch variable."
+            if ( sum(!all_var) > 1) {
+                err_text <- " are not a numeric Crunch variables."
+            }
+            halt( paste0(serialPaste(var_names[!all_var]), err_text))
+        }
+        new$graph <- lapply(value, self)
+    }
+    if (!identical(current, new)) {
+        crPATCH(shojiURL(varCat, "orders", "weights"),
```
nealrichardson: Order resources should take a PUT instead of a PATCH because you're overwriting the entire resource. PATCH is poorly defined.

R/weight.R
```diff
@@ -68,6 +89,49 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
     }
 })
 
+#' @rdname weightVariables
+#' @export
+`weightVariables<-` <- function(x, value) {
+    if (is.dataset(x)) {
```
nealrichardson: This if... else if tells me that this is perhaps better done as an S4 method.

R/weight.R
```diff
@@ -68,6 +89,49 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    } else {
+        halt("Left hand side is not a Crunch Dataset or Variable Catalog.")
+    }
+    current <- crGET(shojiURL(varCat, "orders", "weights"))
+    new <- current
+    if (is.null(value)) {
+        new$graph <- NULL
+    } else {
+        if (is.variable(value) || length(value) == 1) {
+            value <- list(value)
```
nealrichardson: Compare to how `is.na<-` and `hiddenVariables<-` (especially) work. When you assign in a variable (or alias!), it adds the variable to the list. We should be consistent.

DESCRIPTION
jonkeane: I had the same thing happen to me on a separate, I think it's a recent(ish) update to the way desc (called via roxygen2) deals with formatting `DESCRIPTION`s:

> Keep trailing space after the : character, see #14 
from 1.0.1 at https://github.com/r-lib/desc/blob/master/inst/NEWS.md

Roxygen2 using `desc` was supposed to help stop spurios `DESCRIPTION` changes:
> Changes to DESCRIPTION (i.e. Collate: and RoxygenNote) now use the desc package. This will minimise spurious changes (#430).
from 6.0.0 at https://github.com/klutometis/roxygen/blob/master/NEWS.md

If only because we will be fighting ourselves, I say let's adopt that convention / let `desc` / `roxygen` force it on us.

R/weight.R
jonkeane: I agree, it also would be easy to define both using the same code path with this pattern:
```
setMethod("weightVariables<-", "VariableCatalog", function (x, value) {
    #work
})

setMethod("weightVariables<-", "CrunchDataset", function (x, value) {
    varCat <- allVariables(x)
    weightVariables(varCat) <- value
})
```

R/weight.R
```diff
@@ -68,6 +89,49 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+                return(as.character(x))
+            }
+            }, character(1))
+        all_var <- vapply(value, is.Numeric, logical(1))
+        if (!all(all_var)) {
+            err_text <- " is not a numeric Crunch variable."
+            if ( sum(!all_var) > 1) {
+                err_text <- " are not a numeric Crunch variables."
+            }
+            halt( paste0(serialPaste(var_names[!all_var]), err_text))
```
jonkeane: You don't need the outer `paste0` here, `halt` automatically calls it.

GShotwell: Someday I will remember this, but it is not this day!

R/weight.R
GShotwell: So you want something like this?
```
weightVariables(ds) <- var1
weightVariables(ds) <- var2
weightVariables(ds)
> c("var1", "var2")
```
How would you remove variables from weightVariables under that pattern?

nealrichardson: Yeah. That's how `hiddenVariables<-` works, and that's how `is.na<-.Categories` works. I think I was basing that model off of how `is.na<-` works in base. 

I could see a few ways to do remove weight variables with that pattern. One would be `is.weight(var) <- FALSE`. I could invent others that involve subsetting catalogs and orders, but that one is probably enough, and pretty natural to think about if you're a user trying to do this. 

What do you think?

GShotwell: I guess I think we should avoid impurity as much as possible. In this case if`weightVariables(ds) <- ds$var` either set or removed `ds$var` as a possible weight variable the behaviour of the function would change depending on the state of the dataset on the server which can lead to some tricky bugs, for instance: 
- A user sets the `weightVariable()` in  a script and the behaviour of that script changes depending on whether its been run an odd or even number of times
- Two users set `weightVariable(ds) <- ds$var` from separate sessions which would result in no weight variable being set

It also feels a little weird to have the assignment operator remove something. In most cases it seems like assigning either creates or overwrites a reference, and if you want to remove a reference you should assign NULL. 

The package doesn't seem to be totally consistent on this point. The example I was emulating was `weight(ds) <- ds$var` which doesn't  remove `ds$var` as the weighting variable. 

In terms of the `is.weight(ds$var) <- TRUE` pattern, it seems like the main user story for this function is a dataset administrator setting the possible weight variables that a viewer can access. So my intuition is that `weightVariables` is more of a property of datasets, then it is of the variables within the dataset so it makes more sense to have the function operate on the dataset and allow them to set several variables in one step. It also might cause some confusion over whether the variable was allowed to be set as a weight or was currently set as the weight. Do we have users who are asking for this identifier function?

What about changing it so that it accepts a character vector of aliases instead of a list of variables? That way they could append variables with `weightVariables(ds) <- c(weightVariables(ds), "var")` and remove a variable with `weightVariables(ds) <- setdiff(weightVariables(ds), "var")`

nealrichardson: I was not advocating that `weightVariables(ds) <- ds$var` would set and then doing `weightVariables(ds) <- ds$var` again would remove. Nothing in the package works like that. I was suggesting, as you stated it before, 

```
weightVariables(ds) <- var1
weightVariables(ds) <- var2
weightVariables(ds)
> c("var1", "var2")
```

Yes, the assignment function should accept aliases.

GShotwell:
@jonkeane @nealrichardson This is ready for review I believe

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
     } else {
         return(c())
     }
 })
 
+#' @rdname weightVariables
+#' @export
+setMethod("weightVariables<-", "CrunchDataset", function(x, value) {
+    weightVariables(allVariables(x)) <- value
+    x <- refresh(x)
```
nealrichardson: This `refresh` shouldn't be necessary; the setting on the one above should do the necessary cache busting.

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+#' list of numeric Crunch variables or a character vector with the aliases of numeric Crunch variables.
+#' @param type a character string determining how the weightVariables
+#' will be modified:
+#' - `"append"` : add `vars` to the current weight variables
+#' - `"remove"` : remove `vars` from the current list of weight variables
+#' - `"replace"`: replace the current weight variables with `vars`
+#'
+#' @export
+modifyWeightVariables <- function(x, vars, type = "append") {
+        current <- crGET(shojiURL(x, "orders", "weights"))
```
nealrichardson: Fix extra indentation

R/weight.R
```diff
@@ -12,21 +12,23 @@
 #' NULL. For the setter, x, modified accordingly. `weightVariables` returns
 #' the aliases (or names, according to `options(crunch.namekey.dataset)`),
 #' of the variables designated as weights.
-#' @aliases weightVariables
+#' @seealso [weightVariables()] [makeWeight()]
+#' @name weight
+#' @aliases is.weight<-
 #' @export
-weight <- function (x) {
+weight <- function(x) {
```
nealrichardson: Please undo the unnecessary linting that is inconsistent with our package style (space after the word "function")--it makes the diff look way bigger than it is.

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
     return(x)
 }
 
 #' @rdname weight
 #' @export
-setMethod("weightVariables", "CrunchDataset",
-    function (x) weightVariables(allVariables(x)))
+is.weight <- function(x){
+    if (is.variable(x)) {
+        ds <- loadDataset(datasetReference(self(x)))
```
nealrichardson: There is a `datasetReference` method for `CrunchVariable` (https://github.com/Crunch-io/rcrunch/blob/master/R/variable.R#L164) so you don't need the `self()` here (and below where the same pattern is followed)

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+    } else {
+        return(FALSE)
+    }
+}
 
 #' @rdname weight
 #' @export
-setMethod("weightVariables", "VariableCatalog", function (x) {
+setMethod("is.weight<-", "NumericVariable", function(x, value) {
+    if (!(is.logical(value)  && length(value) == 1)) {
```
nealrichardson: Don't put two spaces before `&&` (I saw another case of that somewhere else). Also, this check allows `NA` (which is `logical` length 1) to go through.

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+        halt("Right hand side must be TRUE or FALSE")
+    }
+    ds <- loadDataset(datasetReference(self(x)))
+    varcat <- allVariables(ds)
```
nealrichardson: You don't use `varcat` in this function so skip this line.

R/weight.R
nealrichardson: Also, you could merge this line with the next like `new <- current <- crGET(shojiURL(x, "orders", "weights"))` (and perhaps "old" is clearer than "current" in intent)

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+#' - `"remove"` : remove `vars` from the current list of weight variables
+#' - `"replace"`: replace the current weight variables with `vars`
+#'
+#' @export
+modifyWeightVariables <- function(x, vars, type = "append") {
+        current <- crGET(shojiURL(x, "orders", "weights"))
+    new <- current
+    if (is.null(vars)) {
+        new$graph <- NULL
+        type = "remove"
```
nealrichardson: Always assign with `<-`

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+        if (is.variable(vars) || (length(vars) == 1) & !is.character(vars)) {
+            vars <- list(vars)
+        }
+        if (is.character(vars)) {
+            ds <- loadDataset(datasetReference(self(x)))
+            vars <- lapply(vars, function(v){
+                if (v %in% names(x)) {
+                    ds[[v]]
+                } else {
+                    v
```
nealrichardson: Why this "else" case? Under what circumstance would v not be in names(x) and that's ok?

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+        if (!all(all_var)) {
+            var_names <- vapply(vars, function(v){
+                if (is.variable(v)) {
+                    return(name(v))
+                } else {
+                    return(as.character(v))
+                }
+            }, character(1))
+            err_text <- " is not a numeric Crunch variable."
+            if ( sum(!all_var) > 1) {
```
nealrichardson: Please remove the whitespace after the open ( here and on L210

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+                err_text <- " are not a numeric Crunch variables."
```
nealrichardson: Grammar

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+        } else if (type == "remove") {
+            new$graph <- setdiff( current$graph, lapply(vars, self))
+        } else if (type == "replace") {
+            new$graph <- lapply(vars, self)
+        }
+    }
+    if (!identical(current, new)) {
+        crPUT(shojiURL(x, "orders", "weights"),
+            body = toJSON(new))
+        x <- refresh(x)
```
nealrichardson: This `refresh` is probably not necessary. `crPUT` should bust cache so the next `GET` will read fresh from the server. And nothing in `x` is storing the state of the weight variables order.

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+        return(alias(x) %in% weightVariables(ds))
+    } else {
+        return(FALSE)
+    }
+}
+
+#' @rdname weightVariables
+#' @export
+setMethod("is.weightVariable<-", "NumericVariable", function(x, value) {
+    if (!(is.logical(value)  && length(value) == 1)) {
```
nealrichardson: See comment above. And since the same logic is appearing in multiple places, perhaps this should be factored out to a function for reuse.

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+        halt("Right hand side must be TRUE or FALSE.")
+    }
+    ds <- loadDataset(datasetReference(self(x)))
+    varcat <- allVariables(ds)
+    if (value) {
+        modifyWeightVariables(varcat, x, "append")
```
nealrichardson: FWIW this could be one line with `ifelse(value, "append", "remove")`, or you could pull that ifelse out to its own line like `mode <- ifelse(value, "append", "remove")` and then pass that to `modifyWeightVariables`

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+#' @param vars Variables to add or remove  this can be a numeric Crunch variable,
+#' list of numeric Crunch variables or a character vector with the aliases of numeric Crunch variables.
+#' @param type a character string determining how the weightVariables
+#' will be modified:
+#' - `"append"` : add `vars` to the current weight variables
+#' - `"remove"` : remove `vars` from the current list of weight variables
+#' - `"replace"`: replace the current weight variables with `vars`
+#'
+#' @export
+modifyWeightVariables <- function(x, vars, type = "append") {
```
nealrichardson: This is... unnatural. You have `setMethod("modifyWeightVariables", ...)` but then you also have an exported function called `modifyWeightVariables`. I'd either delete the S4 method dispatch and do `if (is.dataset(x)) modifyWeightVariables(allVariables(x), ...)` pseudo-dispatching inside the function, or rename the function so that it doesn't clash with the methods.

R/weight.R
```diff
@@ -41,33 +43,216 @@ weight <- function (x) {
...
+}
+
+#' @rdname weightVariables
+#' @export
+setMethod("is.weightVariable<-", "NumericVariable", function(x, value) {
+    if (!(is.logical(value)  && length(value) == 1)) {
+        halt("Right hand side must be TRUE or FALSE.")
+    }
+    ds <- loadDataset(datasetReference(self(x)))
+    varcat <- allVariables(ds)
```
nealrichardson: Because you have a `CrunchDataset` method for `modifyWeightVariables`, you don't need to pull out `varcat` here; the dispatch will do that for you.

R/weight.R
```diff
@@ -114,7 +299,7 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
 #' }
 #' @seealso [weight<-()]; [settings()] for the "default weight" for other
 #' dataset viewers.
-makeWeight <- function (..., name) {
+makeWeight <- function(..., name) {
```
nealrichardson: This looks like more unnecessary space removal.

R/weight.R
GShotwell: I think the conjunction is negated right? So it doesn't go through: 
```
 > value <- NA
 > !(is.logical(value) && length(value) == 1)
[1] FALSE
```
Is there a clearer way of doing a NAND gate?

R/weight.R
GShotwell: My thought was if someone assigned something like `is.weight(ds) <- c("weight1", "weight23")` where weight23 wasn't a variable I would want to error by saying "weight 23 is not a valid variable" so I need to preserve the name that was passed to trigger the correct errors later on.

R/weight.R
GShotwell: As per the above I think the logical check does not let NA through

R/weight.R
```diff
@@ -41,18 +43,78 @@ weight <- function (x) {
...
+#' @export
+setMethod("is.weight<-", "NumericVariable", function (x, value) {
+    validateValue(value)
+    ds <- loadDataset(datasetReference(self(x)))
+    if (value) {
+        weight(ds) <- x
+    } else {
+        weight(ds) <- NULL
+    }
+    x <- refresh(x)
```
nealrichardson: This refresh shouldn't be necessary.

R/weight.R
```diff
@@ -41,18 +43,78 @@ weight <- function (x) {
...
+    } else {
+        return(FALSE)
+    }
+}
+
+#' @rdname weight
+#' @export
+setMethod("is.weight<-", "NumericVariable", function (x, value) {
+    validateValue(value)
+    ds <- loadDataset(datasetReference(self(x)))
```
nealrichardson: Another unnecessary `self`

R/weight.R
```diff
@@ -62,12 +124,115 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
     } else {
         return(c())
     }
 })
 
+#' @rdname weightVariables
+#' @export
+setMethod("weightVariables<-", "CrunchDataset", function (x, value) {
+    weightVariables(allVariables(x)) <- value
+    x <- refresh(x)
```
nealrichardson: Refresh should not be necessary.

R/weight.R
```diff
@@ -62,12 +124,115 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    new <- old <- crGET(shojiURL(x, "orders", "weights"))
+    if (is.null(vars)) {
+        new$graph <- NULL
+        type <- "remove"
+    } else {
+        if (is.variable(vars) || (length(vars) == 1) & !is.character(vars)) {
+            vars <- list(vars)
+        }
+        if (is.character(vars)) {
+            ds <- loadDataset(datasetReference(self(x)))
```
nealrichardson: Another `self` you don't need.

R/weight.R
```diff
@@ -62,12 +124,115 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+            vars <- lapply(vars, function (v) {
```
nealrichardson: Perhaps follow the pattern in `hideVariables`, which also takes aliases and gets their URLs: https://github.com/Crunch-io/rcrunch/blob/master/R/hide-variables.R#L40

The `[` method handles your validation of whether you're given valid aliases.

It takes a dataset as an argument, not a dataset or variable catalog, FWIW. Not sure you need to take anything other than dataset here either.

R/weight.R
```diff
@@ -62,12 +124,115 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+        } else if (type == "remove") {
+            new$graph <- setdiff(old$graph, lapply(vars, self))
+        } else if (type == "replace") {
+            new$graph <- lapply(vars, self)
+        }
+    }
+    if (!identical(old, new)) {
+        crPUT(shojiURL(x, "orders", "weights"),
+            body = toJSON(new))
+        x <- refresh(x)
```
nealrichardson: Still don't think you need `refresh` here.

R/weight.R
```diff
@@ -188,3 +352,15 @@ generateWeightEntry <- function (expr) {
...
 }
+
+#' Validate that a value is a logical vector of length 1
+#'
+#' @param value The right hand side of an assignment
+#' @keywords internal
+#' @return nothing, called for its side effects
+#'
+validateValue <-  function (value) {
+        if (!(is.logical(value) && length(value) == 1)) {
```
nealrichardson: Indentation

R/weight.R
```diff
@@ -188,3 +352,15 @@ generateWeightEntry <- function (expr) {
...
+        halt("Right hand side must be TRUE or FALSE")
+    }
+}
```
nealrichardson: How about a unit test then? 

```
> validateValue <-  function (value) {
+     if (!(is.logical(value) && length(value) == 1)) {
+         halt("Right hand side must be TRUE or FALSE")
+     }
+ }
> 
> library(testthat)
> 
> test_that("validateValue validates correctly", {
+     expect_error(validateValue(TRUE), NA)
+     expect_error(validateValue(FALSE), NA)
+     expect_error(validateValue(NA), "Right hand side must be TRUE or FALSE")
+ })
Error: Test failed: 'validateValue validates correctly'
* validateValue(NA) did not throw an error.
```

R/weight.R
```diff
@@ -188,3 +352,15 @@ generateWeightEntry <- function (expr) {
...
     ))
 }
+
+#' Validate that a value is a logical vector of length 1
+#'
+#' @param value The right hand side of an assignment
+#' @keywords internal
+#' @return nothing, called for its side effects
+#'
+validateValue <-  function (value) {
```
nealrichardson: Function name is too generic for this purpose.

R/AllGenerics.R
```diff
@@ -122,6 +122,10 @@ setGeneric("groupClass", function (x) standardGeneric("groupClass"))
 setGeneric("entityClass", function (x) standardGeneric("entityClass"))
 setGeneric("entitiesInitializer", function (x) standardGeneric("entitiesInitializer"))
 setGeneric("weightVariables", function (x) standardGeneric("weightVariables"))
+setGeneric("modifyWeightVariables", function (x, vars, type) standardGeneric("modifyWeightVariables"))
```
nealrichardson: Since you've turned this into a function, you don't need this generic definition.

R/weight.R
```diff
@@ -62,12 +124,115 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+#' @param type a character string determining how the weightVariables
+#' will be modified:
+#' - `"append"` : add `vars` to the current weight variables
+#' - `"remove"` : remove `vars` from the current list of weight variables
+#' - `"replace"`: replace the current weight variables with `vars`
+#'
+#' @export
+modifyWeightVariables <- function (x, vars, type = "append") {
+    if (is.dataset(x)) {
+        x <- allVariables(x)
```
nealrichardson: A confusing thing about doing this is that now the `x` you're returning from the function isn't the dataset. See below; I'd suggest only taking dataset as the argument. I think it simplifies your logic, following the existing models of [hideVariables](https://github.com/Crunch-io/rcrunch/blob/master/R/hide-variables.R#L39) and [deleteVariables](https://github.com/Crunch-io/rcrunch/blob/master/R/variable-delete.R#L16)

R/weight.R
GShotwell: I was just following the existing pattern for the `weightVariables()` getter, do we want to change everything to just take a dataset?

nealrichardson: The `weightVariables` getter doesn't return the dataset/catalog you gave it, it returns the aliases of the weight variables, so it's a little different. 

I'm fine with accepting both dataset and variable catalog if you can do it without undue complexity, but I don't view it as a requirement that the function must accept variable catalog as an argument.

GShotwell: That's a good point. I guess it seemed like if `weightVariables(varcat)` worked then `weightVarialble(svarcat) <- variables` should also work.

nealrichardson: TBH I probably implemented the variable catalog method for `weightVariables` only because I had to get the catalog to get the weight order resource anyway, not because I expected/intended a user to do it.

GShotwell: Great, I'll take out the VariableCatalog, that'll simplify things considerably

R/weight.R
GShotwell: Resolved by removing variable catalog support

R/weight.R
```diff
@@ -41,18 +43,77 @@ weight <- function (x) {
...
+#' be a numeric Crunch variable, list of numeric Crunch variables or
+#' a character vector with the aliases of numeric Crunch variables.
+#'
+#' @return For the getter the variables which can be set as the dataset's weight. For
+#' the setter the dataset duly modified.
+#' @seealso [weight()] [makeWeight()]
+#' @name weightVariables
+#' @examples
+#' \dontrun{
+#' weightVariables(ds) <- ds$weight
```
nealrichardson: Are all of these examples consistent with the final implementation?

R/weight.R
```diff
@@ -41,18 +43,77 @@ weight <- function (x) {
...
+#' weightVariables(ds) <- list(ds$weight, ds$weight2)
+#' weightVariables(ds) <- NULL
+#' weightVariables(ds) <- c("weight", "weight2")
+#' #To append a weight to existing weights
+#' weightVariables(ds) <- c(weightVariables(ds), "weight3")
+#' #To remove a weight
+#' weightVariables(ds) <- setdiff(weightVariables(ds), "weight3")
+#' }
+#' @name weightVaraibles
```
nealrichardson: Spelling

R/weight.R
```diff
@@ -188,3 +341,15 @@ generateWeightEntry <- function (expr) {
...
     ))
 }
+
+#' Validate that a value is TRUE or FALSE
+#'
+#' @param value The right hand side of an assignment
+#' @keywords internal
+#' @return nothing, called for its side effects
+#'
+validateWeightVariableValue <-  function (value) {
```
nealrichardson: How about just "is.TRUEorFALSE" or something? (And then don't say "Right hand side" in the error message). Then we can use this elsewhere where we require a single T/F value.

tests/testthat/test-weighting.R
```diff
@@ -39,6 +56,56 @@ with_mock_crunch({
...
+        )
+        expect_error(
+            weightVariables(oldds) <- list(oldds$gender, oldds$textVar),
+            'Gender and Text variable ftw are not numeric Crunch variables.'
+        )
+    })
+
+    test_that("validateWeightVariableValue errors correctly", {
+        expect_error(validateWeightVariableValue(TRUE), NA)
+        expect_error(validateWeightVariableValue(TRUE), NA)
```
nealrichardson: L99 and L100 are the same.

R/weight.R
```diff
@@ -62,12 +123,105 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    new <- old <- crGET(shojiURL(varcat, "orders", "weights"))
+    if (is.null(vars)) {
+        new$graph <- NULL
+        type <- "remove"
+    } else {
+        if (is.variable(vars) || (length(vars) == 1) & !is.character(vars)) {
+            vars <- list(vars)
+        }
+        if (is.character(vars)) {
+            ds <- loadDataset(datasetReference(x))
```
nealrichardson: Could you please add some inline comments explaining why you're doing some of the things you're doing here? There are lots of `if`s and I can't even remember from one review to the next why you're doing some of this this way, so I'm definitely not going to remember next time I look at the code after we're done with this PR.

R/weight.R
```diff
@@ -62,12 +123,105 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+            body = toJSON(new))
+    }
+    return(x)
+}
+
+#' @rdname weightVariables
+#' @export
+is.weightVariable <- function (x) {
+    if (is.variable(x)) {
+        ds <- loadDataset(datasetReference(self(x)))
```
nealrichardson: Third time I've pointed out an unnecessary `self` inside of `datasetReference` calls. Please find and replace all. There are more than just this one left.

R/weight.R
```diff
@@ -41,18 +43,73 @@ weight <- function (x) {
...
+#'
+#' Editors can change which variables can be set as the weighting variable
+#' for a dataset. For instance if several weights have been calculated they
+#' can let the user choose which of those variables to use a weight, but prevent
+#' the user from choosing other variables as weight. This function allows you
+#' to get and set the `weightVariables` of a dataset, which are the variables which
+#' can be set as the dataset [weight()].
+#'
+#' @param x a CrunchDataset
+#' @param value For the setter variables to set as the `weightVariables` this can
```
nealrichardson: What happens if provide `NULL` as a value?

R/weight.R
```diff
@@ -41,18 +43,73 @@ weight <- function (x) {
...
+#' be a numeric Crunch variable, list of numeric Crunch variables or
+#' a character vector with the aliases of numeric Crunch variables.
+#'
+#' @return For the getter the variables which can be set as the dataset's weight. For
```
nealrichardson: `weightVariables` returns a character vector of the aliases of the variables that are eligible to be used as weights.

R/weight.R
```diff
@@ -41,18 +43,73 @@ weight <- function (x) {
...
+#' the setter the dataset duly modified.
+#' @seealso [weight()] [makeWeight()]
```
nealrichardson: Perhaps see also `modifyWeightVariables`, which gives you greater control. In fact, consider documenting all of these together and indicate that `weightVariables<-` is a convenience method around `modifyWeightVariables(mode="append")`

R/weight.R
```diff
@@ -62,12 +119,115 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+        new$graph <- NULL
+        type <- "replace"
+    } else {
+        if (is.variable(vars) || (length(vars) == 1) & !is.character(vars)) {
+            ## Wrap single variables in a list
+            vars <- list(vars)
+        }
+        if (is.character(vars)) {
+            ## Get variables from aliases
+            ds <- loadDataset(datasetReference(x))
```
nealrichardson: `x` is the dataset, so you don't need to load a copy of it.

R/weight.R
GShotwell: Thanks! I had such a hard time trying to figure out how to say that clearly.

R/weight.R
GShotwell: It will clear the weightVariables. I've added some documentation about that, does that seem okay?

nealrichardson: Yes, that was a rhetorical question because the examples showed using `NULL` but the `@param` didn't say it was allowed, and nothing said what would happen if you did. I'll re-review the docs.

R/weight.R
```diff
@@ -62,12 +99,135 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+    }
+}
+
+#' @rdname modifyWeightVariables
+#' @export
+setMethod("is.weightVariable<-", "NumericVariable", function (x, value) {
+    isTRUEorFALSE(value)
+    ds <- loadDataset(datasetReference(x))
+    action <- ifelse(value, "append", "remove")
+    modifyWeightVariables(ds, x, action)
```
nealrichardson: just `return(modifyWeightVariables(...`

R/weight.R
```diff
@@ -62,12 +99,135 @@ setMethod("weightVariables", "VariableCatalog", function (x) {
...
+            USE.NAMES = FALSE)))
     } else {
         return(c())
     }
 })
 
+#' @rdname modifyWeightVariables
+#' @export
+setMethod("weightVariables<-", "CrunchDataset", function (x, value) {
+    modifyWeightVariables(x, value, type = "append")
```
nealrichardson: same as below, just `return(modifyWeightVariables`

PR #126, opened 2017-09-26
Cleaning up how tests are run (and what happens when using `devtools::load_all()`)

jonkeane:
There are a few things that have changed:
1. When the line that `source`s `inst/crunch-test.R` is not run when the session is interactive. (ie: no more environment cluttering / option changing when running `devtools::load_all()`
1. The code that looks for `inst/crunch-test.R` so that tests can be run in a number of environments has been made more robust, it now searches directories (recursively) one at a time, up to 3 parents. It should stop as soon as it finds one, so this shouldn't add too much time.
1. Moved the helper `envOrOption()` to `misc.R` and exported (this is useful for the crunchdev package, as well as being able to call it in the `pointTo()` function below.
1. Created a convenience function `pointTo()` this lets developers change what backend they are pointing at with ease. It takes places 'local' (sets up to interact with a vagrant-based backend, reads the `test.*` options from the `.Rprofile`) and 'prod' (sets up to interact with the prod (or alpha) backend, reads the `crunch.*` options from the `.Rprofile`)


R/misc.R
```diff
@@ -155,3 +155,105 @@ vectorOrList <- function (obj, type) {
     }
     return(FALSE)
 }
+
+#' Grab either env variable or .Rprofile based option
```
nealrichardson: "options" need not be .Rprofile

R/misc.R
```diff
@@ -155,3 +155,105 @@ vectorOrList <- function (obj, type) {
...
+#' 
+#' .Rprofile options are like "test.api", while env vars are "R_TEST_API", 
+#'  this function will grab the environment variable if it is find, otherwise
```
nealrichardson: Read this sentence(?) aloud

R/misc.R
```diff
@@ -155,3 +155,105 @@ vectorOrList <- function (obj, type) {
...
+    if (nchar(envvar)) {
+        ## Let environment variable override .Rprofile, if defined
+        return(envvar)
+    } else {
+        return(getOption(opt))
+    }
+}
+
+
+#' Grab either env variable or .Rprofile based option
```
nealrichardson: While I'm not yet convinced that this function needs to be in the package at all, I'm definitely sure it doesn't need to be here twice ;)

R/misc.R
```diff
@@ -155,3 +155,105 @@ vectorOrList <- function (obj, type) {
...
+#' 
+#' .Rprofile options are like "test.api", while env vars are "R_TEST_API", 
+#'  this function will grab the environment variable if it is find, otherwise
+#'  it looks for the R-based option value.
+#' 
+#' @param opt the option to get
+#' 
+#' @return the value of the option
+#' 
+#' @export
```
nealrichardson: keywords internal

R/misc.R
```diff
@@ -155,3 +155,105 @@ vectorOrList <- function (obj, type) {
...
+#' @param place a place to point to (either 'local' or 'prod')
+#' @param api_url a url to use by default: "https://app.crunch.io/api/"
+#' @param use_env_api_url should the api url from the environment variables be
+#' used? (default: `FALSE`, this will always be `FALSE` when choosing `prod = 'local'`)
+#' 
+#' @return nothing
+#' 
+#' @keywords internal
+#' @export
+pointTo <- function (place, api_url = "https://app.crunch.io/api/", use_env_api_url = FALSE) {
```
nealrichardson: I don't think this function belongs in the package. I could see something like:

```
setCrunchAPI <- function (subdomain, port=NULL) {
    if (!is.null(port)) {
        api <- paste0("http://", subdomain, ".crunch.io:", port, "/api/")
    } else {
        api <- paste0("https://", subdomain", .crunch.io/api/")
    options(crunch.api=api)
}
```

and you could then have `login` use `envOrOption` for email and pw (see https://github.com/Crunch-io/rcrunch/blob/master/R/auth.R#L50). 

With those tools, could you build in the other ad hoc behaviors to `crunchdev`?

R/misc.R
jonkeane: I was doubling up, just in case we lost one.

R/auth.R
```diff
@@ -29,15 +29,18 @@ deleteSessionInfo <- function () {
...
-#' \code{login()} to authenticate. For running batch jobs, this could be
+#' `crunch.email` and `crunch.pw` for convenience. If you do so, you can simply
+#' `login()` to authenticate. For running batch jobs, this could be
 #' particularly useful. However, be warned that storing your
 #' password in a plain text file such as .Rprofile is a security risk (though
 #' perhaps less so than in every .R script you write), and we
 #' cannot officially recommend that you do so.
 #'
+#' Additionally, your email and password can be stored in and read from the 
+#' environmental variables `R_crunch_email` and `R_crunch_pw` respectively.
```
nealrichardson: The env vars are all caps.

R/misc.R
```diff
@@ -155,3 +155,54 @@ vectorOrList <- function (obj, type) {
     }
     return(FALSE)
 }
+
+#' Grab either env variable or option
+#' 
+#' .Rprofile options are like "test.api", while env vars are "R_TEST_API".
```
nealrichardson: Change these examples not to say "test"

tests/testthat.R
```diff
@@ -1,4 +1,5 @@
 library(httptest)
+
```
nealrichardson: Why?

inst/crunch-test.R
```diff
@@ -63,7 +56,7 @@ silencer <- temp.option(show.error.messages=FALSE)
 
 assign("entities.created", c(), envir=globalenv())
 with_test_authentication <- function (expr) {
-    if (run.integration.tests) {
+    if (getOption("run.integration.tests")) {
```
nealrichardson: Why this change?

tests/testthat/helper.R
```diff
@@ -1,16 +1,37 @@
...
+            pth <- "."
+            
+        } else {
+            pth <- file.path(".", paste0(rep("..", i), collapse = "/"))
+        }
+        fls <- list.files(pth, recursive = TRUE)
+        fl <- fls[grepl(file_name, fls)]
+        if (length(fl) == 1) return(file.path(pth,fl))
+        if (length(fl)  > 1) {
+            warning("More than one file named ", file_name, " using the first one.")
```
nealrichardson: On what file system would it be possible to have two files with the same name?

tests/testthat/helper.R
```diff
@@ -1,16 +1,37 @@
...
+    stop("Found no file named ", file_name, " Stopping tests because they will not function without ", file_name)
+}
+
+## Our "test package" common harness code
+crunch_test_path <- system.file("crunch-test.R", package="crunch")
+if (crunch_test_path == "") {
+    # hack for loadall
+    crunch_test_path <- find_file("crunch-test.R")
+}
+# only source crunch-test.R when the session is not interactive
```
nealrichardson: Why?

tests/testthat/helper.R
```diff
@@ -1,16 +1,37 @@
...
+if (!interactive()) {
+    message("Loading helpers from: ", crunch_test_path)
```
nealrichardson: Why message? I don't want to see this every time I run tests.

tests/testthat/helper.R
```diff
@@ -1,16 +1,37 @@
...
-## `try` added because of some devtools::document weirdness
-try({
-    crunch_test_path <- system.file("crunch-test.R", package="crunch")
-    if (crunch_test_path == "") {
-        # hack for loadall
-        crunch_test_path <- "../../inst/crunch-test.R"
+# find a file starting with the current directory and looking (recursively) in parent folders limited to levels_up
+find_file <- function (file_name, levels_up = 3) {
+    for (i in 0:levels_up) {
+        if (i == 0) {
```
nealrichardson: The contents of the loop could be simpler:

```
    f <- file.path(paste0(rep("..", i), collapse = .Platform$file.sep), file_name)
    if (file.exists(f)) return(f)
```

tests/testthat.R
jonkeane: I find the extra line there easier to read, but I've removed it.

inst/crunch-test.R
jonkeane: It was necessary for another idea I had, but I decided not to implement that. I kept it because it's more consistent with the other options we are using to control what's going on with the tests (`test.user` and `test.pw`). On a much more minor point: It also keeps the integration setting in a harder to read place than a standard object is, so it's a bit harder to accidentally over-write `run.integration.tests` without actively setting an object. Though I admit it is very unlikely that any developer would do this. I don't feel strongly about it either way, I'm happy to change it back to be a global object.

tests/testthat/helper.R
jonkeane: Because the search is recursive, the same file could be found in more than one directory. This is probably overly-defensive since we are the only ones who can't always just reference the installed file.

tests/testthat/helper.R
jonkeane: The main reason for this is that when `crunch-test.R` is sourced, it alters the current global environment. The way that `load_all` works, for better or worse, will always execute all of the helper files (my understanding is the reason for this is so that those helper functions are available to help you write and debug tests). But the main consequence is that when Gordon and I are working on something and we want to be connected to a host other than local, `source`ing `crunch-test.R` (via `devtools::load_all()`) clobbers those settings. (And adds a bunch of stuff that we don't need for debugging to our global environments). 

I know it sounds minor, but I go back and forth and comment this source line many times a day as I'm developing.

There are two other approaches we might have to take:
1. Instead of just testing interactivity, we can test what the ultimate caller is, and not source this file when what is calling it is not `devtools::load_all()`. I've tested this with 
```quote(devtools::load_all) != sys.calls()[[1]][[1]]```
and it works well (even calling `devtools::test()` interactively still loads, running tests via make, via RStudio's test, etc. all work (and necessarily load the helper file). I would want to make this a bit more robust in case for example: some how `sys.calls` doesn't have those indices we still get a `TRUE` like value instead of `logical(0)`.

1. We could make `crunch-test.R` be more respectful of user/developer environments and options and move the code that alters the global environment to a function that is called before the tests are loaded but critically not somewhere that `devtools::load_all()` will run it (ie in `tests/testthat.r` for `R CMD check`, in the `Makefile` strings that run tests for `make test`, in the `crunchdev strings that run tests for anything that uses the `crunchdev` helpers, finally, I don't know of a way to insert into what Rstudio is doing when it runs `devtools::test(".")` which is a pretty big drawback.) 

Conceptually I prefer the second one since it is more respectful of user/developer options and global environments, but given the added hassle it would be to run the setup code/function when it would need to be run plus that it would be a pretty major change in the way that code works, I'm opting for the first (slash 0th — using `interactive()` — option).

File: tests/testthat/helper.R (hunk 12 above)
Comment (jonkeane) [reply tests/testthat/helper.R]: I'm happy to take it out, I was using it mostly as a confirmation that the helpers were loaded when I was making sure that the file was being sourced in all of the situations it needed to be sourced in. I thought it would be a good canary if something changes with the way that any of the different ways that tests are run that is impacting this process.

File: tests/testthat/test-misc.R [hunk 14]
```diff
@@ -117,3 +117,12 @@ test_that("vectorOrList", {
...
     expect_false(vectorOrList(list("a", 1, "c"), "numeric"))
 })
+
+test_that("setCrunchAPI", {
+    old_crunch_api <- getOption("crunch.api")
+    setCrunchAPI('foobar')
+    expect_equal(getOption("crunch.api"), "https://foobar.crunch.io/api/")
+    setCrunchAPI('barfoo', 8888)
+    expect_equal(getOption("crunch.api"), "http://barfoo.crunch.io:8888/api/")
+    options("crunch.api" = old_crunch_api)
```
nealrichardson: It's generally better to do something like this as an `on.exit` up at the top of this block so that you ensure that it is called. Otherwise if the code on L123-126 errors, it will stop execution before resetting your option.

But even better for this case, use the `reset.option` context: https://github.com/Crunch-io/rcrunch/blob/81b02b9636fcce2ec8342ca3aab3529b15ab3044/tests/testthat/helper-contexts.R#L44

tests/testthat/helper.R
jonkeane: I might be missing something obvious, but that wouldn't search recursively, right? I've simplified that for loop in other ways, but in order to get into the `inst` directory when we need to, I've set it up to search recursively. I've also limited the search the folder specified with `paste0(rep("..", i)` or `inst` inside of it. This limits the scope of searching and the code is a bit cleaner.

nealrichardson: That's what goes inside the for loop. So it recurses (starting at cwd) until it finds a file, then breaks and returns it.

R/auth.R
```diff
@@ -29,15 +29,18 @@ deleteSessionInfo <- function () {
...
-#' \code{login()} to authenticate. For running batch jobs, this could be
+#' `crunch.email` and `crunch.pw` for convenience. If you do so, you can simply
+#' `login()` to authenticate. For running batch jobs, this could be
 #' particularly useful. However, be warned that storing your
 #' password in a plain text file such as .Rprofile is a security risk (though
 #' perhaps less so than in every .R script you write), and we
 #' cannot officially recommend that you do so.
 #'
+#' Additionally, your email and password can be stored in and read from the 
+#' environmental variables `R_CRUNCH_EMAIL` and `R_CRUNCH_PW` respectively.
```
nealrichardson: Should update the README to reflect how you can specify credentials in env vars too now.

tests/testthat/helper.R
```diff
@@ -1,16 +1,40 @@
...
+## Our "test package" common harness code
+crunch_test_path <- system.file("crunch-test.R", package="crunch")
+if (crunch_test_path == "") {
+    # hack for devtools::test / testthat::test_package
+    crunch_test_path <- find_file("crunch-test.R")
+}
+
+# only source crunch-test.R when the this is not being called (as the top most
+# call) by devtools::load_all or crunchdev::load_all
+loaded_by_load_all <- sys.calls()[[1]][[1]] == quote(devtools::load_all) |
```
nealrichardson: How about using `rstudioapi::isAvailable()`, like how we do in `login`: https://github.com/Crunch-io/rcrunch/blob/master/R/auth.R#L83-L84

Or even just `if (identical(.Platform$GUI, "RStudio"))`

inst/crunch-test.R
nealrichardson: I'm not worried about a user overwriting `run.integration.tests`. I don't really think it's worth moving this to an option; it's not really more consistent because you aren't setting it with `envOrOption`, and it's naming isn't consistent with the other args (namespace, R_ prefix, etc.). And I don't think that changing all that to make it fully consistent with the other options is worth our time.

tests/testthat/helper.R
jonkeane: Without specifying `inst`, that won't actually work:
```
setwd("recrunch/tests/testthat/")

for (i in 0:4) {
    f <- file.path(paste0(rep("..", i), collapse = .Platform$file.sep), "crunch-test.R")
    if (file.exists(f)) print("Found crunch-test.R")
}
``` 
Never finds `test-crunch.R`. Adding `inst` like I did will work, but I also want to make sure if `test-crunch.R` has been moved out of `inst` already, we still catch it. I might be over thinking this, but I want to make sure we're resilient in case there are any (more) changes in `devtools`/`testthat`'s process for running tests.

nealrichardson: Wait--don't you *not* want to source this file when devtools::load_all is running tests?

So, I think there is some overthinking happening, and not being someone who has any problem with our test setup, I'm somewhat at a disadvantage in reasoning about this, but how about just this:

```
if (file.exists("inst/crunch-test.R")) {
    source("inst/crunch-test.R")
} else if (nchar(system.file("crunch-test.R", package="crunch"))) {
    source(system.file("crunch-test.R", package="crunch"))
}
```

If you've ever installed crunch, the system.file will work.

tests/testthat/helper.R
jonkeane: Both of those work for most situations, although they both prevent running `devtools::test()` interactively in RStudio. That's not something I do frequently (I'm much more likely to run them from RStudio's build pain or a new terminal in RStudio). The only time I can think of when running `devtools::test()` interactively is helpful is dropping a `browser()` in a problematic test.

tests/testthat/helper.R
jonkeane: We actually need the file to be `source`d when `devtools::load_all` is called during the testing process, but not be `source`d when `devtools::load_all` (or `crunchdev::load_all` which is a wrapper around the `devtools` version) is called on its own.

Although `system.file("crunch-test.R", package="crunch")` should work by looking at the installed version of Crunch, it doesn't when the package is being loaded with `devtools::load_all`. I've got a minimal package that shows exactly this behavior, if you care to see it, but the behavior is below: The `mintest` package is installed, but `devtools::load_all` at first doesn't find `inst/test-file`, but confusingly if `devtools::load_all` is run again in the same session it does find the file in `inst`, although it's not in the system library, but in the package directory.
```
Kaf:mintest jkeane$ R -e 'print(system.file("test-file", package="mintest"))'
> print(system.file("test-file", package="mintest"))
[1] "/Users/jkeane/Library/R/3.4/library/mintest/test-file"
> 
Kaf:mintest jkeane$ R -e 'load_all("."); load_all(".")'
> load_all("."); load_all(".")
Loading mintest
[1] ""
Loading mintest
[1] "/Users/jkeane/Documents/mintest/inst/test-file"
> 
```
I've dug through the source of `devtools` and can't find exactly where `load_all` is going wrong. FWIW, with `pkgload::load_all()` this issue seems to be resolved. `pkgload::load_all()` is supposed to replace `devtools::load_all()`, and is already used in the master branch of `devtools`. Though I can't find any sort of roadmap for when that will be released. Once that's transitioned to all of this finding code can go away, hopefully for good.
```
Kaf:mintest jkeane$ R -e 'pkgload::load_all(".")'
> pkgload::load_all(".")
Loading mintest
[1] "/Users/jkeane/Documents/mintest/inst/test-file"
```

jonkeane: Even more confusingly, the reason that `devtools::test()` works when the helper file has `system.file()` alone is that it apparently (accidentally) loads the helpers twice https://github.com/hadley/devtools/issues/1579

jonkeane:
I've updated the way that the detection of if `inst/crunch-test.R` should be `source`ed or not works based on a suggestion from @GShotwell It now uses a combination of `is.interactive()` and checking an environment variable which testthat/devtools sets based on a comment from the devtools maintainers' suggestions for doing exactly what we want.

nealrichardson:
LGTM, please resolve your merge conflict and we can accept.

PR #134, opened 2017-10-07
Updating UserCatalogs, ShojiCatalogs, and a new expropriateUser function 151774405

jonkeane:
ShojiCatalogs now accept a `secondary` argument for index on other properties than name, url, or numeric index.

UserCatalogs default to subsetting by email if supplied with a string. Names or other properties can be accessed with `secondary`

New `expropriateUser` function to transfer all Crunch object ownership from one user to another.


R/user.R
```diff
@@ -72,3 +72,47 @@ resetPassword <- function (email) {
             url_base=absoluteURL("../password/change/${token}/", app)
         ))))
 }
+
+#' @rdname catalog-extract
+#' @export
+setMethod("[", c("UserCatalog", "character"), function (x, i, ...) {
+    dots <- list(x=x, i=i, secondary = emails(x))
```
nealrichardson: Don't think you need `x` and `i` in this part, do you? They must be in `list(...)`.

R/user.R
```diff
@@ -72,3 +72,47 @@ resetPassword <- function (email) {
...
+#' @export
+setMethod("[[", c("UserCatalog", "character"), function (x, i, ...) {
+    dots <- list(x=x, i=i, secondary = emails(x))
+    dots <- modifyList(dots, list(...))
+    do.call("callNextMethod", dots)
+})
+
+#' Expropriate all Crunch objects from a user
+#' 
+#' If you want to transfer all teams, projects, and datasets owned by one user
```
nealrichardson: Can you add a note of why you'd do such a thing? (e.g. when someone leaves the organization, transfer their things before deleting their account).

R/user.R
```diff
@@ -72,3 +72,47 @@ resetPassword <- function (email) {
...
+#' permission to expropriate
+#' 
+#' @param from a character of the email address of the user to expropriate from
+#' @param to a character of the email address of the user who should be the new 
+#' owner
+#' 
+#' @return None
+#' @export
+expropriateUser <- function (from, to) {
+    if (!askForPermission(paste0("Really expropriate user ", dQuote(from), "?"))) {
```
nealrichardson: Can you use different/more words in the prompt here, just to be explicit about what this command will do?

R/dataset-catalog.R
```diff
@@ -68,7 +68,8 @@ setMethod("is.published<-", c("DatasetCatalog", "logical"), function (x, value)
 #' @param name for `$`, the same as `i` for `[[`
 #' @param j Invalid
 #' @param drop Invalid
-#' @param ... additional arguments
+#' @param ... additional arguments (eg `secondary` a vector of elements to match
```
nealrichardson: "e.g." or just say "for example"

inst/WORDLIST
```diff
@@ -102,6 +103,7 @@ didn
 dist
 doesn
 env
+eg
```
nealrichardson: Fix that "eg" in the man page an you can remove this.

R/user.R
jonkeane: Unless I'm misunderstanding, `x` and `i` aren't in `list(...)` it's only all of the other args:

```
> foo <- function(x, i, ...) {
+     print(list(...))
+ }
> foo(x="x", i="i", secondary="sec")
$secondary
[1] "sec"
```

We also need them in the `setMethod` definition since we are dispatching on them, so we can't pull them from there so that they get passed with `...`.

nealrichardson: Right, of course, I must have had stuck in my brain the `match.call` stuff you were talking about last week. My bad.

R/user.R
```diff
@@ -72,3 +72,51 @@ resetPassword <- function (email) {
...
+#' @param from a character of the email address of the user to expropriate from
+#' @param to a character of the email address of the user who should be the new 
+#' owner
+#' 
+#' @return None
+#' @export
+expropriateUser <- function (from, to) {
+    if (!askForPermission(paste0("This will transfer ownership of all teams, ",
+                                 "projects, and datasets from ", dQuote(from), 
+                                 " to ", dQuote(to), " Do you really want to ",
```
nealrichardson: after `dQuote(to),` just do ". Proceed?" (and note the period to end the previous sentence).

PR #139, opened 2017-10-11
Univariate variable and cube transforms

jonkeane:
Transforms with insertions can now be set and viewed on categorical variables. 1d CrunchCubes with a variable that has a transform member will also display the transforms/insertions.

There are many changes here, including making a parent class for insertions and categories called `AbsCat(s)` so that they can share methods (and be coalesced when needed). Although that wasn't strictly necessary for transforms, it will be helpful when trying to do the same with multiple response categories, etc.


nealrichardson:
Some notes on the interface design

```r
ds <- loadDataset("Econ something")
ds$approval
## [1] Tremendously approve
## [2] Somewhat approve
## [3] Neither approve nor disapprove
## [4] Somewhat disapprove
## [5] Totally disapprove

subtotals(ds$approval) <- list(
    Subtotal("Total approval", c(1, 2), after="Somewhat approve"), 
    Subtotal("Total disapproval", grep("disapprove", names(categories(ds$approval)), value=TRUE), after=5)
    Heading("A subtitle", after=0)
)

print(ds$approval)
## [1] Tremendously approve
## [2] Somewhat approve
## Total approval
## [3] Neither approve nor disapprove
## [4] Somewhat disapprove
## [5] Totally disapprove
## Total disapproval

crtabs(~ approval, data=ds)

show.CrunchCube <- function (x, yada, ...) {
    applyTransforms(x)
}

prop.table <- function (..., transforms=TRUE) {

}

or 

prop.table(noNets(cube), 2)

noNets <- function (cube) {
    transforms(cube) <- NULL
    return(cube)
}

# do this?
transformed(categories(ds$approval)) <- transformed(categories(ds$approval))[c(1,2,4,5,6,3,7)]
ds$approval
## [1] Tremendously approve
## [2] Somewhat approve
## [3] Neither approve nor disapprove
## [4] Somewhat disapprove
## [5] Totally disapprove
## Total approval
## Total disapproval

transforms(ds$approval)
nets()
```

R/abstract-categories.R
```diff
@@ -0,0 +1,224 @@
...
+    return(x)
+}
+
+############################################
+## Categories specific methods
+############################################
+
+#' @rdname Categories
+#' @export
+setMethod("ids", "AbsCats", function (x) vapply(x, id, integer(1)))
```
nealrichardson: If these methods are "Categories specific", why are they set on "AbsCats"?

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+#' CrunchCube to return. The options can be any of the following: "cube_cells"
+#' for the untransformed values from the cube itself, "subtotals" for the
+#' subtotal insertions, and "headings" for any additional headings. Any
+#' combination of these can be given, by default all will be given.
+#'
+#' `subtotalArray(cube)` is a convenience function that is equivalent to
+#' `applyTransforms(cube, include = c("subtotals"))`
+#'
+#' @param x a CrunchCube
+#' @param ary an array to use, if not using the default array from the cube
```
nealrichardson: `ary` is weird and no one will remember how to (mis)spell that to use it. Can you pick something better?

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+#' @export
+setMethod("transforms", "CrunchCube", function (x) {
+    trans <- tryCatch(Transforms(data=index(variables(x))[[1]]$view$transform), error = function(e) NULL)
+    return(trans)
+})
+
+#' @rdname Transforms
+#' @export
+setMethod("transforms<-", c("CrunchCube", "NULL"), function (x, value) {
+    # hacked access to delete the transforms. This should have a more systematic
```
nealrichardson: Write this as a "TODO" so we can grep for it more easily when we're looking for TODOs (plus, a good editor/IDE will highlight "TODO")

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+})
+
+#' Remove transformations from a cube
+#'
+#' This is useful if you don't want to see or use any transformations like Subtotals and Headings
+#'
+#' @param cube a CrunchCube
+#'
+#' @return the cube with no transformations
+#'
```
nealrichardson: Add examples.

R/insertions.R
```diff
@@ -0,0 +1,89 @@
...
+    if (any(duplicated(names(object)))) {
+        return("Invalid insertion names: must be unique")
+    }
+    return(TRUE)
+})
+
+is.insertions <- function (x) inherits(x, "Insertions")
+
+validateNewAnchor <- function (anchor) {
+    if (!is.numeric(anchor)) {
```
nealrichardson: Should this be `is.whole`?

R/subtotals-and-headings.R
```diff
@@ -0,0 +1,237 @@
...
+                      serialPaste(dQuote(reqs)), ". Missing: ",
+                      serialPaste(dQuote(reqs[!mems])))
+    } else {
+        val <- TRUE
+    }
+
+    return(val)
+})
+
+init.Subtotal <- function (.Object, ...) {
```
nealrichardson: You don't have to define this as a separate function. Just define it inline with `setMethod("initialize", ...`

R/variable-transforms.R
```diff
@@ -0,0 +1,334 @@
...
+#' Transforms
+#' @param value For `[<-`, the replacement Transforms to insert
+#' @name Transforms
+#' @aliases transforms transforms<-
+NULL
+
+
+getTransforms <- function (x) {
+    var_entity <- entity(x)
+    trans <- var_entity@body$view$transform
```
nealrichardson: If you're doing `@` a lot to get attributes, you're missing a "getter" method,

R/cube-result.R
```diff
@@ -294,6 +297,14 @@ cubeMarginTable <- function (x, margin=NULL, measure=1) {
     ## for which that would do the wrong thing.
     keep.these <- evalUseNA(mt, dims[mt_margins], x@useNA)
     out <- subsetCubeArray(mt, keep.these)
+
+    # only attempt to apply a transform if the margin is 1 rows for now.
+    if (!is.null(margin) && margin == 1) {
+        margin_trans <- dims[[mapped_margins]]$references$view$transform
+        if (!is.null(margin_trans)) {
```
nealrichardson: Seems you already do this check for existence of transforms (more robustly, in fact) in `applyTransforms`: https://github.com/Crunch-io/rcrunch/pull/139/files#diff-fc6c252cf7a4ec51064a3ca7fde28541R91

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+#' #                         NA                         23                         24 
+#' # Neither Agree nor Disagree          Strongly Disagree          Somewhat Disagree 
+#' #                         18                         19                         16 
+#' }
+#'
+#' @aliases subtotalArray
+#' @export
+applyTransforms <- function (x, ary = cubeToArray(x), ...) {
+    # if there are row transforms, calculate them and display them
+    row_trans <- tryCatch(Transforms(data=index(variables(x))[[1]]$view$transform), error = function(e) NULL)
```
nealrichardson: Why not use the (nearly identical) logic in https://github.com/Crunch-io/rcrunch/pull/139/files#diff-fc6c252cf7a4ec51064a3ca7fde28541R151 ? You're going to have to anyway when you implement applying transforms on all dims, not just 1.

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+
+    ary <- subsetCubeArray(ary, keep_these)
+
+    return(ary)
+}
+
+#' @rdname Transforms
+#' @export
+setMethod("transforms", "CrunchCube", function (x) {
+    trans <- tryCatch(Transforms(data=index(variables(x))[[1]]$view$transform), error = function(e) NULL)
```
nealrichardson: Hard-coding `[[1]]` means you're getting a `Transforms` object for the first dim. But let's think about where this is going. `lapply` over the variables (better: `dimensions`, see cube-variables.R for how `variables` isn't necessarily what you think it is), and return a list of Transforms. If there is none for that variable, what should it return? NULL or how about a 0-length Transforms object? If 0-length Transforms would just work (as in, your other code would handle that fine without the special-casing of is.null), maybe that's better.

nealrichardson: Anyway, make this return all the transforms for all dims, and then feel free to move your hard-coded [[1]] to the calling function(s) (applyTransforms)

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+    return(trans)
+})
+
+#' @rdname Transforms
+#' @export
+setMethod("transforms<-", c("CrunchCube", "NULL"), function (x, value) {
+    # hacked access to delete the transforms. This should have a more systematic
+    # accessor
+    x@dims@.Data[[1]]$references$view$transform <- value
```
nealrichardson: `x@dims@.Data` says we have missing methods

`[[1]]` hardcoding isn't necessary here, lapply over

`$references$view$transform` will fail if the "references" has no "view" element, and I don't think there's a guarantee that it will (I see this pattern in a few places)

R/cube-transforms.R
```diff
@@ -0,0 +1,177 @@
...
+        ary <- subsetTransformedCube(ary, x)
+    }
+
+    # TODO: calculate column transforms
+
+
+    return(ary)
+}
+
+subsetTransformedCube <- function (ary, cube) {
```
nealrichardson: Why does this function need to exist? Is there a way that existing similar functions (`cubeToArray`, `subsetCubeArray`, etc.) can be used?

R/variable-transforms.R
```diff
@@ -0,0 +1,334 @@
...
+#' `trans` calculated
+#' @keywords internal
+calcInsertions <- function (vec, elements, var_cats) {
+    # we always calculate insertions at the lowest dimension
+    if (length(dim(vec)) > 1) {
+        halt("Calculating varaible transforms is not implemented for dimensions ",
+             "greater than 1.")
+    }
+
+    vec_out <- vapply(elements, function (element) {
```
nealrichardson: This looks like a function that should be pulled out for easier readability/testing.

R/transform.R
```diff
@@ -0,0 +1,161 @@
...
+#' `is.abscat.subtotal` is `x` a subtotal insertion?
+#' `is.abscat.heading` is `x` a heading insertion?
+#' `is.abscat.category` is `x` a category?
+#'
+#' @param x an AbsCat object
+#'
+#' @return logical if the AbsCat object is
+#'
+#' @name AbsCat-type-tests
+#' @keywords internal
```
nealrichardson: Why do these exist? Shouldn't `is.Subtotal` etc. be what you use? Seems like if you're going to do introspection of the objects to infer their type, you're not getting much from using S4 classes.

R/transform-calculate.R
```diff
@@ -0,0 +1,147 @@
+# calculcate transforms for an array.
```
nealrichardson: Could you please document these functions a bit more verbosely? Would help me to follow along and understand what all this is doing.

tests/testthat/test-cube-transforms.R
```diff
@@ -0,0 +1,234 @@
...
+    test_that("showTransforms works on a variable", {
+        cat_show_trans <- array(c(75, 30, 45, 50, 95, 25, 55, 75, 75, NA),
+                                dim = 10,
+                                dimnames = list(pets = c(
+                                 "First one", "Birds", "Cats", "Dogs",
+                                 "Dogs+Cats", "Lizards", "Birds+Lizards",
+                                 "Toward the end", "Cats+Birds (missing anch.)",
+                                 "Rocks+Birds (incl. missing)")))
+
+        expect_output(trans_pets <- showTransforms(ds$pets))
```
nealrichardson: Why aren't you asserting what the output is?

R/abstract-categories.R
jonkeane: I've changed these to be a bit more strict, now they are only defined for the parent class if they are needed. I was defining them for AbsCats, even if they only currently applied to one subclass because I could see a possible future where they would be needed, but we can always upgrade them in the future.

I've also included a vignette describing abstract categories including tables showing what methods are and aren't implemented.

R/cube-transforms.R
jonkeane: I've added a `dimensions<-` method, so no longer reaching into the object. I've removed the [[1]] hard codings, and added a few extra methods to make the hard coding for rows be all in one place. 

I've also added some checks for `$references$view$transform` view/transform existing. basically, try to select, and use the fact that if any of the named elements don't exist, we get `NULL` (which is used elsewhere with this structure). I've looked at the other uses of that pattern, and they all have similar checks immediately after, so I think we're covered, but if there's a better way to do this, I'm happy to change to that.

R/transform.R
jonkeane: I've refactored how these works, and now everything respect s4 classes / inheritance.

jonkeane:
This is ready for another round of review. I've addressed everything brought up here, which involved a decent amount of refactoring (for the better!). To aide this (and our future selves) I've written [a vignette describing abstract categories ](https://github.com/Crunch-io/rcrunch/blob/univariate-transforms/vignettes/abstract-categories.Rmd) (sorry the output isn't there, it is when the vignette is compiled, though so running `pkgdown::build_site()` locally will render them.

nealrichardson:
Go ahead and merge; this has branched for long enough. We can take up any remaining issues in the next iteration.

Issue #137, opened 2017-10-10
feature request: expose dataset metadata within listDatasets()


deliabailey:
the dataset metadata that is available in the project page in the GUI would be useful to be able to access from listDatasets() rather than having to load the dataset first.

use case is similar to the GUI... I'd like to order the datasets in the project in some way (or search over a characteristic) to choose which one to load.

nealrichardson:
Following https://github.com/Crunch-io/rcrunch/commit/96f0f02b1782e883a751f3739aff5480a599f8bb there is now an `as.data.frame` method for dataset catalogs, so you can get a view of all of the dataset metadata with `as.data.frame(datasets())` or `as.data.frame(datasets(someproject))`, and you can filter that however you want. `as.data.frame` takes a couple of optional arguments that may be relevant for you, so see the help page for it for details.

This was included in the 1.18.4 release, now on CRAN.

PR #140, opened 2017-10-11
Export an as.data.frame method for all catalogs

nealrichardson:


nealrichardson:
This provides a partial solution for https://github.com/Crunch-io/rcrunch/issues/137

PR #142, opened 2017-10-18
Vectorized is selected methods for categories

### LINKED ISSUE #123: is.selected vectorized
is.selected is a property of each Category in a multiple_response variable. It would be great if it was a property of Categories of a multiple_response variable so instead of `names(categories(ds$mr_var))[sapply(categories(ds$mr_var), is.selected)]` I could do `names(is.selected(categories(ds$mr_var)))` like I can with `names(is.na(categories(ds$mr_var)))`

GShotwell:
Closes #123. I also pulled the existing `is.selected` documentation into its own `rd` file, rather than including it in describe categories. This lets us put those methods in the `Array Variables` subsection of the pkgdown site which I think is the more natural place for it.


tests/testthat/test-categories.R
```diff
@@ -484,5 +509,14 @@ with_test_authentication({
...
                 c("Not Asked", "Skipped", "Bird", "Canine", "Cat"))
         })
+        test_that("is.selected method gets and set selection value",  {
+            ds$mr_sub1 <- factor(sample(1:2, nrow(ds), replace = TRUE))
+            ds$mr_sub2 <- factor(sample(1:2, nrow(ds), replace = TRUE))
+            ds$mr <- makeMR(ds[, c("mr_sub1", "mr_sub2")], selections = "1", name = "mr")
+            is.selected(categories(ds$mr)) <- c(TRUE, TRUE, TRUE)
+            expect_true(all(is.selected(categories(ds$mr))))
+            is.selected(categories(ds$mr)[2]) <- FALSE
+            expect_false(is.selected(categories(ds$mr)[2]))
```
nealrichardson: Assert that it is `c(TRUE, FALSE, TRUE)`

R/categories.R
```diff
@@ -187,6 +187,39 @@ NULL
...
+#' @export
+setMethod("is.selected", "Categories", function (x) structure(vapply(x, is.selected, logical(1), USE.NAMES=FALSE), .Names=names(x)))
+
+#' @rdname is-selected-categories
+#' @export
+setMethod("is.selected<-", "Categories", function (x, value) {
+    if (length(value) != length(x)) {
+        halt("You supplied ", length(value), " logical values for ", length(x), " Categories.")
+    }
+    for (i in seq_along(x@.Data)) {
```
nealrichardson: `Map`/`mapply` instead of for loop. cf. https://github.com/Crunch-io/rcrunch/blob/master/R/categories.R#L235, which itself could probably be simplified by passing in the function as `is.na<-` rather than redefining it.

R/categories.R
```diff
@@ -187,6 +187,39 @@ NULL
...
+NULL
+
+#' @rdname is-selected-categories
+#' @export
+setMethod("is.selected", "Categories", function (x) structure(vapply(x, is.selected, logical(1), USE.NAMES=FALSE), .Names=names(x)))
+
+#' @rdname is-selected-categories
+#' @export
+setMethod("is.selected<-", "Categories", function (x, value) {
+    if (length(value) != length(x)) {
```
nealrichardson: This will fail incorrectly if value is length 1.

R/categories.R
```diff
@@ -187,6 +187,39 @@ NULL
 #' @export
 setMethod("is.na", "Categories", function (x) structure(vapply(x, is.na, logical(1), USE.NAMES=FALSE), .Names=names(x)))
 
+#' is.selected for Categories
+#'
+#' Crunch Multiple Response variables identify one or more categories as "selected"
```
nealrichardson: Need a period at the end of this sentence.

R/categories.R
```diff
@@ -187,6 +187,44 @@ NULL
...
+NULL
+
+#' @rdname is-selected-categories
+#' @export
+setMethod("is.selected", "Categories", function (x) structure(vapply(x, is.selected, logical(1), USE.NAMES=FALSE), .Names=names(x)))
+
+#' @rdname is-selected-categories
+#' @export
+setMethod("is.selected<-", "Categories", function (x, value) {
+    if( is.TRUEorFALSE(value)) {
```
nealrichardson: Space goes on the other side of the (

R/dichotomize.R
```diff
@@ -25,10 +25,9 @@ setMethod("is.dichotomized", "Categories",
 
 .dichotomize.categories <- function (x, i) {
     ## Internal method for dichtomizing Categories (or lists)
-    x[i] <- lapply(x[i], function (a) {
-        a$selected <- TRUE
-        return(a)
-    })
+    selections <- is.selected(x)
+    selections[i] <- TRUE
+    is.selected(x) <- selections
```
nealrichardson: This should be just `is.selected(x[i]) <- TRUE`, I believe.

PR #144, opened 2017-10-19
addGeoMetadata updates

jonkeane:
`addGeoMetadata` now will automatically match when a geodatum has duplicate properties.

### DIFF
--- a/R/geo.R
+++ b/R/geo.R
@@ -73,8 +73,8 @@ setMethod("geo<-", c("CrunchVariable", "NULL"),
 #' Add geodata metadata to a crunch variable
 #'
 #' If the variable matches a single geographic shapefile hosted by crunch, 
-#' `addGeoMetadata` will make the appropriate `[CrunchGeography]` to add to a 
-#' variable's `geo()` metadata. It matches based on how well the contents of the
+#' `addGeoMetadata` will make the appropriate [`CrunchGeography`] to add to a 
+#' variable's [`geo()`] metadata. It matches based on how well the contents of the
 #' variable match the feature properties that are in each shapefile.
 #' 
 #' If more than one property of the same geographic shapefile has the same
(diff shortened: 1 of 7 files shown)


R/geo.R
```diff
@@ -105,8 +119,25 @@ addGeoMetadata <- function (variable) {
...
+        # duplications of the same feature.
+        if (length(unique(match_scores$geodatum_name)) == 1 &
+            length(unique(match_scores$geodatum)) == 1) {
+            warning("The geodatum ", dQuote(unique(match_scores$geodatum_name)),
+                    " has multiple properties that match the variable (",
+                    serialPaste(dQuote(match_scores$property)),
+                    "). Using ", dQuote(match_scores$property[1]), ".")
+            match_scores <- match_scores[1,]
+        } else {
+            halt("There is more than one possible match. Please specify the ",
```
nealrichardson: PR seems generally ok to me, though it's too bad that there's no test coverage for this "else" case. Aside from the stats (which could be massaged with different line breaking in the code), my concerns are (1) that the error message construction itself could fail, and (2) that even with the error message I wouldn't know how to "specify the geography manually", and it's not clear to me how the output of that message would help me to know how. 

That said, I recognize that this error message itself isn't new in this PR. And in terms of how much it's worth spending to improve it: how likely is it that someone could get here?

R/geo.R
```diff
@@ -58,7 +58,7 @@ setMethod("geo<-", c("CrunchVariable", "CrunchGeography"),
               geodata <- list(geodata = list(value))
 
               ent <- setEntitySlot(entity(x), "view", geodata)
-              dropCache(cubeURL(x))
+              dropCache(datasetReference(x))
```
nealrichardson: This seems overzealous. Adding geo data to a variable shouldn't require busting cache on everything related to the dataset, should it? What did you find that made dropping only cubes inadequate?

R/geo.R
```diff
@@ -115,6 +143,32 @@ addGeoMetadata <- function (variable) {
...
 }
 
+# format the error message when there are multiple matches
+multiMatchHelper <- function (match_scores, match_field, quoted_var) {
+    msg <- c("There is more than one possible match. You will need to specify ", 
+             "the geography manually using one of the following ",
+             "CrunchGeographies:\n")
+    
+    calls <- apply(match_scores, 1, function(match_row) {
+        # TODO: add coloring to this message when crayon is available
```
nealrichardson: Yes, I was just looking at bringing in crayon for the variable folder stuff I've started. FWIW it's a simple addition, only depends on base packages and doesn't require compilation: https://github.com/r-lib/crayon/blob/master/DESCRIPTION#L36

PR #146, opened 2017-10-20
Variable folder API

nealrichardson:
Work in progress. Some of this will require the new folder API, but I'm seeing how far I can get with our existing backend to push out the R interface design. See docs, examples, and TODOs in folder.R


R/folders.R
```diff
@@ -0,0 +1,135 @@
...
+#' even if all parent directories didn't already exist; and `mv` doesn't
+#' require that the directory to move to already exist. `mkdir` also takes an
+#' optional `variables` argument, which allows you to say "make this directory
+#' **and** put these variables in it" in one command. As a result of these
+#' behaviors, `mv` and `mkdir` here are essentially the same, just with
+#' arguments in reverse order: `mv` is move "variables" to "path", while `mkdir`
+#' is create "path" and put "variables" in it.
+#'
+#' Both functions take "dataset" as the first argument, so they naturally
+#' support pipelining (as with the \code{\%>\%} operator).
```
jonkeane: Did `` `\%>\%` `` fail on this because of parsing?

nealrichardson: There was some failing so I went to magrittr source to see how they did their roxygen and stole it.

PR #147, opened 2017-10-23
Subvariable definition

### LINKED ISSUE #72: addSubvariable(s) allow VariableDefinition
The `addSubvariables` function should allow `VariableDefinition`s as the subvariable argument. That way, rather than having to add a variable to the dataset and then and it to an MR/cat array, you could skip a step.

GShotwell:
Closes #72 

- Adds support for variable definitions within `addSubvariables`
- Allows users to supply mixed lists of variables and subvariables
- There was a workaround in the code to make sure that the subvariable ordering was correct. I found the implementation a bit hard to follow so decided to refactor that to use assignment rather than a second patch.


R/add-subvariable.R
```diff
@@ -12,25 +13,55 @@
...
     ## Do the adding
     crPATCH(shojiURL(variable, "catalogs", "subvariables"),
         body=toJSON(sapply(new.urls, emptyObject, simplify=FALSE)))
-
-    ## Workaround because apparently bind/rebind isn't retaining the order
-    crPATCH(self(variable),
-        body=toJSON(list(subvariables=I(c(subvar.urls, new.urls)))))
-
+    variable <- refresh(variable)
+    subvariables(variable) <- subvariables(variable)[order]
```
nealrichardson: Can you remove this update entirely? There have been changes on the backend since this workaround was put here, so it may not be necessary anymore. 

If it is still necessary, I think we should keep the existing implementation, even if it is less readable, because it is more efficient (saves several potentially large requests). The users that use this function use it heavily, so it's important that we don't make unnecessary requests.

R/add-subvariable.R
```diff
@@ -12,25 +13,55 @@
...
     dropCache(datasetReference(variable))
-    invisible(refresh(variable))
+    return(invisible(refresh(variable)))
 }
 
 #' @rdname addSubvariable
 #' @export
 addSubvariables <- addSubvariable
+
+add_subvar_def <- function(var, subvar){
```
nealrichardson: Please camelCase the function name and insert spaces around the () in the function definition, for consistency.

R/add-subvariable.R
```diff
@@ -12,25 +13,55 @@
...
+        subvar <- list(subvar)
+    }
+
+    vardefs <- vapply(subvar,
+        function(x) inherits(x, "VariableDefinition"),
+        logical(1))
+
+    if (any(vardefs)) {
+        ds <- loadDataset(datasetReference(var))
+        ds <- addVariables(ds, subvar[vardefs])
```
nealrichardson: A more efficient way to do this is to call POSTNewVariable directly, what addVariables does internally. What you really want out of all of this is the URL(s) of the variables being added--that's what you're going to PATCH with in addSubvariables--and that's what POSTNewVariable returns. That way, you don't have to refresh the dataset and try to find the variable entity in the dataset. Then this function should return the vector of urls.

R/add-subvariable.R
```diff
@@ -12,25 +13,55 @@
...
+
+        ## Since this function accepts mixed lists of variables and definitions
+        ## we need to replace the definitions with the variables once they've been added.
+        subvar[vardefs] <- lapply(subvar[vardefs], function(x){
+            if (is.variable(x)) {
+                x
+            } else {
+                ds[[x$name]]
```
nealrichardson: Like I said above, this statement shouldn't be necessary, but for reference, this isn't correct. Variables are indexed by alias by default in datasets, so this will only work if the name and alias are the same, or if the user has changed that default. See "namekey" elsewhere in the code.

R/add-subvariable.R
GShotwell: Ah great, that makes a lot more sense

R/add-subvariable.R
GShotwell: It looks like this is still necessary. When I take out the second patch it returns subvariables in the wrong order. I'll return it to the previous method,

tests/testthat/test-add-subvariable.R
```diff
@@ -13,7 +13,7 @@ with_test_authentication({
             c("Home", "Work", "doggy daycare"))
     })
 
-    test_that("Add more than one subvariable to an array", {
+test_that("Add more than one subvariable to an array", {
```
nealrichardson: Fix indentation.

tests/testthat/test-add-subvariable.R
```diff
@@ -27,4 +27,24 @@ with_test_authentication({
         expect_identical(names(subvariables(ds$allpets)),
             c("Cat", "Dog", "Bird", "Turtle", "Gerbil", "Coyote", "Cockroach"))
     })
+ds <- refresh(ds)
```
nealrichardson: Fix indentation

R/add-subvariable.R
```diff
@@ -22,15 +23,59 @@ addSubvariable <- function (variable, subvariable) {
     crPATCH(shojiURL(variable, "catalogs", "subvariables"),
         body=toJSON(sapply(new.urls, emptyObject, simplify=FALSE)))
 
-    ## Workaround because apparently bind/rebind isn't retaining the order
+        ## Workaround because apparently bind/rebind isn't retaining the order
```
nealrichardson: Fix indentation

R/add-subvariable.R
```diff
@@ -22,15 +23,59 @@ addSubvariable <- function (variable, subvariable) {
...
+        ## Be silent so we can throw the errors together at the end
+
+        ## Check for errors
+        errs <- vapply(new_var_urls, is.error, logical(1))
+        if (any(errs)) {
+            if (length(errs) == 1) {
+                ## Just one variable added. Throw its error.
+                rethrow(new_var_urls[[1]])
+            }
+            halt("The following variable definition(s) errored on upload: ",
```
nealrichardson: If you're here, you know it's plural "definitions"

R/add-subvariable.R
```diff
@@ -22,15 +23,59 @@ addSubvariable <- function (variable, subvariable) {
...
+
+    vardefs <- vapply(subvar,
+        function(x) inherits(x, "VariableDefinition"),
+        logical(1))
+
+    out <- vector("list", length(subvar))
+
+    if (any(vardefs)) {
+        ds <- loadDataset(datasetReference(var))
+        var_cat_url <- shojiURL(ds, "catalogs", "variables")
```
nealrichardson: Perhaps a better way to do this, since I see that this is largely copy-and-pasted from `addVariables`, would be to factor https://github.com/Crunch-io/rcrunch/blob/master/R/add-variable.R#L37-L55 out into a function and reuse it here. That way, you'd also get test coverage for the error handling code below for free, which this branch doesn't have.

PR #151, opened 2017-10-27
collapseCategories in place

### LINKED ISSUE #79: Combine categories in place (update instead of derive)
Occasionally, it is much more useful to combine without deriving because you know that a category or subvariable is permanently going away. For example if there are categories "Don't know" and "Don’t know" [note the fancy quote], a non-derived combine would be much more useful that a derived one. Right now, the code looks something like:

```
ds$var[ds$var %in% "Don't know"] <- "Don’t know"
categories(ds$var) <- categories(ds$var)[!(names(categories(ds$var)) %in% "Don't know")]
```

GShotwell:
Closes #79


R/combine-categories.R
```diff
@@ -199,3 +199,48 @@ combResps <- function (subvars, combs) {
...
+        categories(var) <- c(categories(var),
+            Category(id = max(ids(categories(var))) + 1,
+                numeric_value = max(values(categories(var)), na.rm = TRUE) + 1,
+                name = to)
+            )
+    }
+    if (is.character(from)) {
+        from <- setdiff(from, to) #in case the user tries to collapse a category into itself
+        var[ var %in% from] <- to
+    } else if (is.Expr(from)) {
```
nealrichardson: I'm not sure I see the use case for the expression "from", and I don't see or test or example that would help me see the use case.

tests/testthat/test-combine-categories.R
```diff
@@ -123,6 +123,27 @@ with_mock_crunch({
...
+            "Variable must be a categorical.")
+        expect_error(collapseCategories(ds$gender, "Male", 1),
+            "Destination category must be a character string of length 1.")
+        expect_error(collapseCategories(ds$gender, "Male", c("young", "old")),
+            "Destination category must be a character string of length 1.")
+        expect_error(collapseCategories(ds$gender, 1, "young"),
+            paste0(sQuote('from'), " must be a character vector or logical expression."))
+    })
+    test_that("collapseCategories produces the expected patch", {
+        expect_PATCH(var <- collapseCategories(ds$location, c("London", "Scotland"), "GB"),
```
nealrichardson: Be clear that the "expected PATCH" here is that you're adding a new category.

tests/testthat/test-combine-categories.R
```diff
@@ -123,6 +123,27 @@ with_mock_crunch({
...
+            "https://app.crunch.io/api/datasets/1/variables/location/",
+            '{"categories":[{"id":1,"missing":false,"name":"London","numeric_value":1},{"id":2,"missing":false,"name":"Scotland","numeric_value":2},{"id":-1,"missing":true,"name":"No Data","numeric_value":null},{"id":3,"name":"GB","numeric_value":3}]}'
+        )
+        expect_POST(var <- collapseCategories(ds$location, c("London", "Scotland"), "London"),
```
nealrichardson: Be clear that the "expected PATCH" here is that you aren't including "London" in the update command (though what would the harm be of including it?)

tests/testthat/test-combine-categories.R
```diff
@@ -123,6 +123,27 @@ with_mock_crunch({
...
+
+    test_that("collapseCategories errors correctly", {
+        expect_error(collapseCategories(ds$birthyr, "from", "to"),
+            "Variable must be a categorical.")
+        expect_error(collapseCategories(ds$gender, "Male", 1),
+            "Destination category must be a character string of length 1.")
+        expect_error(collapseCategories(ds$gender, "Male", c("young", "old")),
+            "Destination category must be a character string of length 1.")
+        expect_error(collapseCategories(ds$gender, 1, "young"),
+            paste0(sQuote('from'), " must be a character vector or logical expression."))
```
nealrichardson: I think `dQuote` is the standard in the package (there are 77 instances of `dQuote` and 7 of `sQuote`) so let's use that.

R/combine-categories.R
```diff
@@ -199,3 +199,48 @@ combResps <- function (subvars, combs) {
...
+    }
+    if (!(is.character(from) || is.Expr(from))) {
+        halt(sQuote('from'), " must be a character vector or logical expression.")
+    }
+    cats <- names(categories(var))
+    if (!(to %in% cats)) {
+        cats <- c(cats, to)
+        categories(var) <- c(categories(var),
+            Category(id = max(ids(categories(var))) + 1,
+                numeric_value = max(values(categories(var)), na.rm = TRUE) + 1,
```
nealrichardson: Don't include a numeric_value. Default for it is missing (NULL), which is correct.

R/combine-categories.R
```diff
@@ -199,3 +199,48 @@ combResps <- function (subvars, combs) {
...
+                name = to)
+            )
+    }
+    if (is.character(from)) {
+        from <- setdiff(from, to) #in case the user tries to collapse a category into itself
+        var[ var %in% from] <- to
```
nealrichardson: Errant leading whitespace

R/combine-categories.R
```diff
@@ -199,3 +199,48 @@ combResps <- function (subvars, combs) {
...
+    if (!(to %in% cats)) {
+        cats <- c(cats, to)
+        categories(var) <- c(categories(var),
+            Category(id = max(ids(categories(var))) + 1,
+                numeric_value = max(values(categories(var)), na.rm = TRUE) + 1,
+                name = to)
+            )
+    }
+    if (is.character(from)) {
+        from <- setdiff(from, to) #in case the user tries to collapse a category into itself
```
nealrichardson: Though it wouldn't hurt if you made that request.

R/combine-categories.R
```diff
@@ -199,3 +199,48 @@ combResps <- function (subvars, combs) {
...
+collapseCategories <- function (var, from, to) {
+    if (!is.Categorical(var)) {
+        halt("Variable must be a categorical.")
+    }
+    if (!(length(to) == 1 && is.character(to))) {
+        halt("Destination category must be a character string of length 1.")
+    }
+    if (!(is.character(from) || is.Expr(from))) {
+        halt(sQuote('from'), " must be a character vector or logical expression.")
+    }
```
nealrichardson: Should we also check that from and to are different? If they're the same, just return `var`, no work required.

R/combine-categories.R
```diff
@@ -199,3 +199,48 @@ combResps <- function (subvars, combs) {
...
+    cats <- names(categories(var))
+    if (!(to %in% cats)) {
```
nealrichardson: Note that if `to` is not in `cats` and `from` is length 1, what you're really just doing is renaming the category. No data updating required. So check for that.

tests/testthat/test-combine-categories.R
```diff
@@ -169,4 +190,18 @@ with_test_authentication({
         expect_identical(names(categories(ds$combined_petloc)),
             c("Mammals", "Bird", "Skipped", "Not Asked"))
     })
+
+    test_that("collapseCategories works on categorical variable", {
+        ds$cat <- factor(sample(c("cat", "Cat", "dog", "aphid"), nrow(ds), replace = TRUE))
+        var <- ds$cat
```
nealrichardson: Don't do this--just refer to `ds$cat`. It's good to make sure that `ds$cat <- collapseCategories(ds$cat, ...)` works, that the object you're returning is acceptable for assignment back into the dataset.

Also, assert the category names before you do collapseCategories so that you can see how they are different after.

tests/testthat/test-combine-categories.R
```diff
@@ -169,4 +190,18 @@ with_test_authentication({
...
+        var <- collapseCategories(var, c("cat", "Cat"), "cat")
+        expect_identical(names(categories(var)), c("aphid", "cat", "dog", "No Data"))
+        var <- collapseCategories(var, c("cat", "aphid"), "nope")
+        expect_identical(names(categories(var)), c("dog", "No Data", "nope"))
+        var <- collapseCategories(var, var %in% "nope", "yup")
```
nealrichardson: Oh there's the expression test. Now that I see it, I'm not sure I see the value in supporting this. Seems like it could blow up in unexpected/undesirable ways.

R/combine-categories.R
```diff
@@ -199,3 +199,45 @@ combResps <- function (subvars, combs) {
...
+        halt(dQuote('from'), " must be a character vector.")
+    }
+    if (identical(from, to)){
+        return(var)
+    }
+    cats <- names(categories(var))
+    if (!(to %in% cats)) {
+        if (length(from) == 1) {
+            # This case is equivalent to renaming a category
+            names(categories)[from] <- to
```
nealrichardson: Please add a test for this case.

R/combine-categories.R
```diff
@@ -199,3 +199,45 @@ combResps <- function (subvars, combs) {
...
+        halt("Variable must be a categorical.")
+    }
+    if (!(length(to) == 1 && is.character(to))) {
+        halt("Destination category must be a character string of length 1.")
+    }
+    if (!(is.character(from))) {
+        halt(dQuote('from'), " must be a character vector.")
+    }
+    if (identical(from, to)){
+        return(var)
```
nealrichardson: Please add a comment explaining this early return.

R/combine-categories.R
```diff
@@ -199,3 +199,45 @@ combResps <- function (subvars, combs) {
...
+            names(categories)[from] <- to
+        }
+        cats <- c(cats, to)
+        categories(var) <- c(categories(var),
+            Category(id = max(ids(categories(var))) + 1,
+            name = to)
+        )
+    }
+    from <- setdiff(from, to) #in case the user tries to collapse a category into itself
+    var[ var %in% from] <- to
```
nealrichardson: Extra space

R/combine-categories.R
nealrichardson: Also add a test for this case, using `expect_no_request` to assert that it's not trying to do work.

PR #152, opened 2017-10-27
Implement the new variable folder API

nealrichardson:


inst/app.crunch.io/api/datasets/1/folders/1.json
```diff
@@ -0,0 +1,35 @@
...
+    "size": 5,
+    "body": {
+        "name": "Group 1"
+    },
+    "index": {
+        "../3/": {
+            "type": "folder",
+            "name": "Nested"
+        },
+        "../../variables/birthyr/": {
```
aminusfu: the last path segment here should be the variable id, not the alias

inst/app.crunch.io/api/datasets/1/folders/1.json
```diff
@@ -0,0 +1,35 @@
+{
+    "self": "https://app.crunch.io/api/datasets/1/folders/1/",
+    "element": "shoji:catalog",
+    "size": 5,
+    "body": {
+        "name": "Group 1"
+    },
+    "index": {
+        "../3/": {
```
jjdelc: I think this would be `./3/` ? Since its path would be `/folders/1/3/`? Given that the top level only had 1 and 2.

inst/app.crunch.io/api/datasets/1/folders/2.json
```diff
@@ -0,0 +1,37 @@
...
+            "discarded": false,
+            "alias": "starttime",
+            "type": "datetime",
+            "id": "d7c21314ca9e453c93069168681a285c",
+            "description": "Interview Start Time",
+            "notes": "",
+            "resolution": "D",
+            "rollup_resolution": "s"
+        },
+        "../../catarray/": {
```
aminusfu: should be `../../variables/qwert/`

inst/app.crunch.io/api/datasets/1/folders/1.json
aminusfu: No; all the folder entities are direct children of `folders/`, we're not exposing them as `folders/1/2/3/`.

jjdelc: Ok, I can see how urls could get out of control

inst/app.crunch.io/api/datasets/1/folders/1.json
nealrichardson: yes, though these fixtures have arbitrarily used informative ids, just for test readability. The "id" member of the tuple should be updated to match the URL segment, but I'll take that separately.

inst/app.crunch.io/api/datasets/1/folders/1.json
enriquecastl: @aminusfu What is the advantage of having all folders as direct children of root?

jjdelc: * That the hierarchy is not exposed as part of the folder URL, this allows for different users to have different roots.
* Separation between URL and folder position, when a folder is moved, its URL doesn't change
* URLs don't get as big as the tree nesting

aminusfu: Integrated that answer into https://github.com/Crunch-io/design/blob/folders-endpoint/features/datasets/variables/folders.md#folder-catalogs

nealrichardson:
To do items before merging to master and seeking user feedback:

* [ ] Finish `print` method for folder (use crayon, add tests. Also, for consideration: if the length of the listing is greater than some length, say 20, perhaps just print as a vector and not "/n" separated?)
* [ ] Order entities within a folder using an `after` arg to `mv`: mv(list of refs, ".", after=)
* [ ] Make setName(s) work on folders (rework names<- in general)
* [ ] Search path: URLs, aliases, names
* [ ] Add integration test

vignettes/variable-order.Rmd
```diff
@@ -16,99 +16,175 @@ load("vignettes.RData")
...
+These four functions all take a dataset or a folder as the first argument, and they return the same object passed to it, except for `cd`, which returns the selected folder. As such, they are designed to work with `magrittr`-style piping (`%>%`) for convenience in chaining together steps, though they don't require that you do.
+
+## Viewing the folders
+
+To get started, let's pick up the dataset we used in the [array variables vignette](array-variables.html) and view its starting layout. We can do that by selecting the root folder ("/") and printing it
+
+```r
+library(magrittr)
+ds %>%
+    cd("/") %>%
```
GShotwell: It might be worth including a helper for `cd("/")`, I didn't find it super natural to use `cd` when I wanted to take a look at the variable folders for a dataset I think because my mind only went there when I wanted to change the directory. Another thought is we could modify the print method for datasets to include folder information, and so avoid the method collision between datasets and the top level VariableFolder.

R/folders.R
```diff
@@ -40,33 +34,123 @@
...
+#' \dontrun{
+#' ds <- loadDataset("Example survey")
+#' rmdir(ds, "Demographics")
+#' # Or with %>%
+#' require(magrittr)
+#' ds <- ds %>%
+#'     rmdir("Demographics")
+#' }
+#' @export
+rmdir <- function (x, path) delete(cd(x, path))
```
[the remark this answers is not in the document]
nealrichardson: It will be a soft delete: moves the folder to a "trash" folder, from which you can restore. That backend hasn't shipped yet.

vignettes/variable-order.Rmd
nealrichardson: Yeah, I noticed this too. Changing the print method for CrunchDataset makes some sense, I think.

vignettes/variable-order.Rmd
```diff
@@ -16,99 +16,175 @@ load("vignettes.RData")
...
+    cd("/")
+    mkdir("Tracking questions") %>%
+    mv(., variables(.), "Tracking questions")
 ```
-```{r, echo=FALSE}
-names(step2)
+
+(Curious about the "dot" notation? See the [magrittr docs](http://magrittr.tidyverse.org/articles/magrittr.html).)
+
+The reason we change to the top level folder here is that there is a subtle difference between passing `ds` to `mv()` versus `cd(ds, "/")`. Whatever object, dataset or folder, that is passed into `mv()` determines the scope from which the objects to move are selected. If you pass the dataset in, you can select any variables in the dataset, regardless of what folder they're in. If you pass in a folder, you're selecting just from that folder's contents. It can be convenient to find all variables that match some criteria across the whole dataset to move them, but sometimes we don't want that. In this case, we wanted only the variables sitting in the top level folder, not nested in other folders, so we wanted `variables(cd(ds, "/"))` and not `variables(ds)`.
```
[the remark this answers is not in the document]
nealrichardson: Agree that's awkward. I'm planning to rewrite the vignette with the SO dataset and avoid weird cases like this. I was trying to follow the old vignette's organization.

vignettes/variable-order.Rmd
```diff
@@ -16,99 +16,175 @@ load("vignettes.RData")
...
+ds %>%
+    cd("/") %>%
+    print(depth = 1)
 ```
 
-## Reordering groups and entities
+## Nested folders
 
-`VariableOrder` and `VariableGroup` support standard R forms of indexing for extracting and reordering.
+We can create groups within groups as well. In the "This week" group, we have a set of questions about Edward Snowden: the first four variables in the group. Let's nest them inside their own group inside "This week":
```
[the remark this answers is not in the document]
nealrichardson: Thanks, that's leftover language from the old vignette, will expunge.

nealrichardson:
Re: your other points @GShotwell ,

1) `mv` already moves folders
2) `print` with `depth` is on my to-do list, the checklist I didn't add to this PR, along with a `verbose` argument to print a `data.frame` like version with more attributes, analogous to something like `ls -al` at the command line.

PR #156, opened 2017-11-02
Changed dichotomize argument names

### LINKED ISSUE #73: dichotamize examples/arguments
It would be great if rather than `x` and `i`, `dichotomize` took arguments that were slightly more informative (i.e. `variable` and `selections`). It would also help align it with the makeMR function better.

Additionally, it could use some examples in the documentation.

GShotwell:
Closes #73


R/dichotomize.R
```diff
@@ -37,35 +47,35 @@ setMethod("dichotomize", c("Categories", "numeric"), .dichotomize.categories)
...
-    return(x)
+setMethod("undichotomize", "Categories", function (categories) {
+    is.selected(categories) <- FALSE
+    return(categories)
 })
 
-.dichotomize.var <- function (x, i) {
-    newcats <- dichotomize(categories(x), i)
-    categories(x) <- newcats
+.dichotomize.var <- function (categories, selection) {
```
nealrichardson: This is where I don't think `categories` as an argument makes sense: when the method/function is for a variable (or is defined both for `Categories` and `CategoricalArrayVariable`). `categories(categories)`?

nealrichardson:
I'm not sure that I agree that this is a good idea, for the reason I noted inline. Maybe there's a better way to improve the documentation, adding examples, etc.?

GShotwell:
- Print out example results in comments
- Add see also to `is.selected<-` methods
- Make arguments `x` and `selections`

PR #165, opened 2017-11-28
change as.data.frame to use csv download

GShotwell:
This PR change the as.data.frame method on CrunchDataFrames to use .csv download instead of `as.vector`. Writing and reading the data from csv brings in the data as a character vector, so there is a function which coerces the new data into the classes which match the Crunch Dataset. 

I also added a `rds` object with the expected data frame to the tests. This will let us change the mechanism of downloading the data in the future without needing to change a bunch of unit tests.


R/as-data-frame.R
```diff
@@ -44,38 +44,62 @@ as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
...
+        return(v)
+    }
+    if (is.Array(var)) {
+        out <- lapply(names(subvariables(var)), function(x) coerceFactor(df[[x]]))
+    } else if (is.Numeric(var)) {
+        out <- list(suppressWarnings(as.numeric(df[[name(var)]]))) # To prevent NA coercison message
+    } else if (is.Categorical(var)) {
+        out <- list(coerceFactor(var))
+    } else if (is.Datetime(var)) {
+        out <- list(suppressWarnings(as.Date(df[[name(var)]]))) # To prevent NA coercision warning
```
jonkeane: spelling, should be _coercion_ here and on line 92 above

tests/testthat/test-crunch-data-frame.R
```diff
@@ -216,4 +206,13 @@ with_test_authentication({
...
         expect_equal(ds_df$v4, as.vector(ds$v4))
     })
+
+    test_that("CDF with assigned variables returns dataframe", {
+        ds_df$new_local_var <- 4:1
+        ds_df$new_local_var2 <- 1:4
+        # and we can get a dataframe from this new one.
+        true_df <- as.data.frame(ds_df)
+        expect_equal(as.vector(ds_df$new_local_var), 4:1)
+        expect_equal(as.vector(ds_df$new_local_var2), 1:4)
```
jonkeane: I might be missing something, but it doesn't look like you're asserting anything about `true_df` here?

R/as-data-frame.R
```diff
@@ -44,38 +44,62 @@ as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
...
-        values <- x[[var]]
-        
-        # Flatten all array variables, using only the subvariable name. We need 
-        # to create a list layer for non-dataframe values so that we can unlist 
-        # them later.
-        if (is.data.frame(values)) {
-            values <- as.list(values)
+    tmp <- tempfile()
+    write.csv(ds, tmp)
+    ds_out <- read.csv(tmp, stringsAsFactors = FALSE)
```
jonkeane: I know R cleans this up on a normal exit, but it might be nice to delete this tempfile once we know it's loaded.

R/as-data-frame.R
```diff
@@ -44,38 +44,62 @@ as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
...
     return(structure(out, class="data.frame", row.names=c(NA, -nrow(ds))))
 }
 
+coerceVariable <- function(var, df){
+    coerceFactor <- function(x){
+        na_cats <- is.na(categories(var))
+        v <- as.vector(x)
+        v[v %in% na_cats] <- NA
+        v <- factor(v, levels = names(categories(var))[!na_cats])
+        return(v)
```
jonkeane: It doesn't look like this would work if someone changed the mode that categoricals are being pulled down as. What if someone set `categorical.mode` on the CrunchDataFrame to be numeric or id?

Also, I would put this function definition outside the `coerceVariable` function, so that you can unit test behaviors like that more easily.

R/as-data-frame.R
```diff
@@ -44,38 +44,62 @@ as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
...
-        
-        # Flatten all array variables, using only the subvariable name. We need 
-        # to create a list layer for non-dataframe values so that we can unlist 
-        # them later.
-        if (is.data.frame(values)) {
-            values <- as.list(values)
+    tmp <- tempfile()
+    write.csv(ds, tmp)
+    ds_out <- read.csv(tmp, stringsAsFactors = FALSE)
+    ds_out[] <- unlist(lapply(ds, coerceVariable, ds_out), recursive = FALSE)
```
jonkeane: Is the `unlist` necessary here? It looks like from https://github.com/Crunch-io/rcrunch/pull/165/files#diff-afa1b86266079dbb588fa3c4dda30d20R83 it might not be?

tests/testthat/test-as-data-frame.R
```diff
@@ -77,15 +77,29 @@ with_mock_crunch({
...
+
     test_that("as.data.frame when a variable has an apostrophe in its alias", {
         t2 <- ds
         t2@variables@index[[2]]$alias <- "Quote 'unquote' alias"
         expect_is(as.data.frame(t2), "CrunchDataFrame")
     })
 
     test_that("as.data.frame(as.data.frame())", {
-        expect_true(is.data.frame(as.data.frame(as.data.frame(ds))))
-        expect_true(is.data.frame(as.data.frame(ds, force=TRUE)))
```
jonkeane: We should find a way to re-write these tests / update mocks so that they can still be run without the backend present.

tests/testthat/test-as-data-frame.R
```diff
@@ -96,31 +110,12 @@ with_mock_crunch({
         expect_equal(names(new_ds_df),
                      aliases(variables(new_ds)))
         expect_equal(ncol(new_ds_df), 6)
-        expect_silent(
-            expect_equal(names(as.data.frame(new_ds_df)),
-                         c("birthyr", "location", "subvar2", "subvar1", 
-                           "subvar3", "textVar", "starttime", "subvar2", 
-                           "subvar1", "subvar3")))
-
```
jonkeane: Same as the comment above: it would be good to be able to keep these tests running even without an API.

tests/testthat/test-as-data-frame.R
```diff
@@ -96,31 +110,12 @@ with_mock_crunch({
...
-                           "subvar2", "subvar1", "subvar3")),
-            "Variable gender is hidden")
-    })
-    
-    test_that("as.data.frame size limit", {
-        with(temp.option(crunch.data.frame.limit=50), {
-            expect_error(as.data.frame(ds, force=TRUE),
-                "Dataset too large to coerce")
-            expect_true(is.data.frame(as.data.frame(ds[,1:2], force=TRUE)))
-        })
```
jonkeane: Same as the comment above: it would be good to be able to keep these tests running even without an API.

R/as-data-frame.R
```diff
@@ -44,38 +44,62 @@ as.data.frame.CrunchDataset <- function (x, row.names = NULL, optional = FALSE,
...
+        }
+    })
+    var_names <- unlist(var_names)
+
+    ## Crunch Dataframes contain both server variables and local variables this
+    ## ensures that both are returned in the proper order.
+    out <- lapply(var_names, function(v){
+        if (v %in% names(ds_out)) {
+            if (v %in% names(ds) && tuple(ds[[v]])$discarded) {
+                warning("Variable ", v, " is hidden", call.=FALSE)
```
jonkeane: It would be good to add `dQuote` around `v` here for consistency.

tests/testthat/test-as-data-frame.R
```diff
@@ -77,15 +77,29 @@ with_mock_crunch({
...
+        df <- lapply(ds, as.vector)
+        names(df) <- names(ds)
+        df[] <- lapply(df, as.character)
+        df <- structure(df, class = "data.frame", row.names=c(NA, -nrow(ds)))
+        expect_is(coerceVariable(ds$birthyr, df)[[1]], "numeric")
+        expect_is(coerceVariable(ds$location, df)[[1]], "factor")
+        expect_is(coerceVariable(ds$textVar, df)[[1]], "character")
+        expect_is(coerceVariable(ds$starttime, df)[[1]], "Date")
+        expect_is(coerceVariable(ds$mymrset, df)[[1]], "factor")
+        expect_is(coerceVariable(ds$catarray, df)[[1]], "factor")
```
jonkeane: Should this be a factor? I remember a discussion about how these should end up and if I'm remembering correctly: arrays should have multiple columns, with the subvariable name/alias being the column name.

tests/testthat/test-as-data-frame.R
jonkeane: I just checked, and it looks like this would also bump our line coverage a lot. It looks like `as.data.frame.CrunchDataFrame` is not hit at all during travis's testing: https://codecov.io/gh/Crunch-io/rcrunch/pull/165/src/R/as-data-frame.R

R/as-data-frame.R
GShotwell: I think it's necessary because of how array variables are handled, so the `lapply` produces a list with the same length as `ds` but `ds_out` breaks all the array variables into their own columns.

tests/testthat/test-as-data-frame.R
GShotwell: I think it should, when you pass a MR variable with three subvars to `coerceVariable` it will return a list of length 3 with each entry being a single factor variable with the alias of the subvar as the name of the variable.

GShotwell:

tests/testthat/test-as-data-frame.R
```diff
@@ -76,6 +76,63 @@ with_mock_crunch({
...
+                    "starttime"))
+        )
+        # now we want the hidden vars to be included
+        new_ds_df <- as.data.frame(new_ds, include.hidden = TRUE)
+        expect_warning(
+            expect_equal(names(csvToDataFrame(csv_df, new_ds_df)),
+                c("birthyr", "gender", "location", "subvar2", "subvar1", "subvar3", "textVar",
+                    "starttime")),
+            paste0("Variable birthyr is hidden"))
+    })
```
jonkeane: Nice refactoring for test coverage of the main content here!

You could also add a test to catch the first 4 lines of `as.data.frame.CrunchDataFrame` by testing that the post for the csv download is as expected (cf https://github.com/Crunch-io/rcrunch/blob/master/tests/testthat/test-export-dataset.R#L9-L11). Then the only things that are un-(unit)-tested are: reading the csv and deleting it when you're done.

tests/testthat/test-as-data-frame.R
```diff
@@ -76,6 +76,63 @@ with_mock_crunch({
...
+        names(df) <- names(ds)
+        df[] <- lapply(df, as.character)
+        cdf <- as.data.frame(ds)
+        df <- structure(df, class = "data.frame", row.names=c(NA, -nrow(ds)))
+        expect_is(coerceVariable(ds$mymrset, df, cdf)[[1]], "factor")
+        expect_is(coerceVariable(ds$location, df, cdf)[[1]], "factor")
+        expect_is(coerceVariable(ds$birthyr, df, cdf)[[1]], "numeric")
+        expect_is(coerceVariable(ds$textVar, df, cdf)[[1]], "character")
+        expect_is(coerceVariable(ds$starttime, df, cdf)[[1]], "Date")
+        expect_is(coerceVariable(ds$catarray, df, cdf)[[1]], "factor")
```
jonkeane: It would be good to also assert the shape of these variables. It took me a bit to realize that array like variables have a different shape. It would be good to make that explicit so it's a bit easier to see.

R/as-data-frame.R
```diff
@@ -31,51 +31,93 @@ NULL
...
-        # to create a list layer for non-dataframe values so that we can unlist 
-        # them later.
-        if (is.data.frame(values)) {
-            values <- as.list(values)
+    tmp <- tempfile()
+    csv_mode = ifelse(attr(x, "mode") == "factor", "name", "id")
+    write.csv(ds, tmp, categorical = csv_mode)
+    ds_out <- read.csv(tmp, stringsAsFactors = FALSE, colClasses = "character")
+    unlink(tmp)
+    return(csvToDataFrame (ds_out, x))
```
jonkeane: Super minor: Looks like there's an extra space here.

R/as-data-frame.R
```diff
@@ -31,51 +31,93 @@ NULL
...
-            values <- as.list(values)
+    tmp <- tempfile()
+    csv_mode = ifelse(attr(x, "mode") == "factor", "name", "id")
+    write.csv(ds, tmp, categorical = csv_mode)
+    ds_out <- read.csv(tmp, stringsAsFactors = FALSE, colClasses = "character")
+    unlink(tmp)
+    return(csvToDataFrame(ds_out, x))
+}
+
+csvToDataFrame <- function(csv_df, cdf) {
```
jonkeane: This is super minor, and I know it's internal so it's not a huge deal, but it would be good to use `crdf` instead of `cdf` here for consistency with our other abbreviations.

R/as-data-frame.R
```diff
@@ -31,51 +31,93 @@ NULL
...
+            return(cdf[[v]])
+        }
+    })
+    names(out) <- var_names
     return(structure(out, class="data.frame", row.names=c(NA, -nrow(ds))))
 }
 
+coerceFactor <- function (var_alias, cats, df, cdf) {
+    # csv download doesn't accomodate numeric factor values, so in that case
+    # we need to call pull the variable in the typical way.
```
jonkeane: Again, minor, but "call" here seems redundant to me

tests/testthat/test-crunch-data-frame.R
```diff
@@ -17,37 +17,37 @@ with_mock_crunch({
...
-        
-        ds_df <- as.data.frame(ds_new)
-        true_df <- as.data.frame(ds_new, force = TRUE)
 
+        ds_df <- as.data.frame(ds_new)
+        true_df <- readRDS("dataset-fixtures/test_ds.rds")
+        true_df <- true_df[, c("birthyr", "gender", "location", "textVar", "starttime")]
         # single column/row extraction
-        expect_equivalent(ds_df[1,], true_df[1,])
+        all.equal(ds_df[1,], true_df[1,])
```
jonkeane: What's the reason for the change from `expect_equivalent` to `all.equal`? If it's false, it doesn't look like this would throw an error in test_that

GShotwell: Hmmm, I'm really not sure honestly. I'll revert this one=

PR #170, opened 2017-12-04
Don't bother with dataset catalog if loadDataset(url)

nealrichardson:
I was working on a new feature in httptest and noticed that this could be simpler and do less work. 

I also deleted a couple of unnecessary checks elsewhere. The changes in [test-new-dataset.R](https://github.com/Crunch-io/rcrunch/commit/bf30cdce8cba#diff-b1fe15667db71be5af9a26a5a256aff2) are just deleting two tests and reducing indentation.


nealrichardson:
Thanks. Please merge at your convenience.

PR #173, opened 2017-12-07
More `methods` imports

jonkeane:
Added more imports (`as<-`, `coerce`, and `coerce<-`) from `methods` to allow RScript to work as expected. Alternatively, we could import all of methods [as some recommend ](http://r-pkgs.had.co.nz/namespace.html#import-s4).


GShotwell:
Is there a reason to not import all methods?

jonkeane:
We were trying to be as targeted as possible. I would vote just importing these methods now, but if we run into this again, go ahead and import all of methods.

PR #174, opened 2017-12-08
Clean up for Subtotals and Headings / Transforms

jonkeane:
This PR addresses some rough edges of Subtotals/Heaidngs:
* `Insertions` objects with `Subtotal`s and `Heading`s that have category names (instead of ids) for `after` or `categories` now display nicely, without error, like other abstract categories
* Cubes with categorical arrays or multiple response actively suppress transforms calculations for now (until we have 2d calculation and displaying working)
* cube dimensions that have subreferences are now referred to as `subvariable_items` instead of `multiple_response` since they can be both categorical arrays and multiple response.
* users can now set transforms on CrunchCubes directly.

### DIFF
--- a/R/cube-transforms.R
+++ b/R/cube-transforms.R
@@ -198,12 +198,11 @@ setMethod("transforms<-", c("CrunchCube", "list"), function (x, value) {
              "names of the cube.")
     }
 
-    
     # replace the transforms for each dimension
     dims <- CubeDims(lapply(dimnames, function (dim_name) {
         dim_out <- dims[[dim_name]]
         if (dim_name %in% names(value)) {
-            # grab the proper insertions, make sure they are proper Insertions
+            # grab the matching insertions, make sure they are proper Insertions
             # and then add them to the dimensions to return.
             one_trans <- value[[dim_name]]
             one_trans$insertions <- Insertions(
--- a/R/cube-transforms.R
+++ b/R/cube-transforms.R
@@ -165,8 +165,9 @@ setMethod("transforms", "VariableCatalog", function (x) {
         return(NULL)
     }
     transes_out <- lapply(transes, function (i) {
-        # if insertions are null return NULL
         # TODO: when other transforms are implemented, this should check those too.
+
+        # if insertions are null return NULL
         if (is.null(i$insertions) || length(i$insertions) == 0) {
             return(NULL)
         }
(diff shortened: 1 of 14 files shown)


R/transform.R
```diff
@@ -34,23 +42,32 @@
...
+        inserts <- NULL
+    } else {
+        inserts <- Insertions(data=trans$insertions)
+        # subtype insertions so that Subtotal, Heading, etc. are their rightful 
+        # selves
+        inserts <- subtypeInsertions(inserts)
+    }
+    
+    # if insertions are null return NULL
+    # TODO: when other transforms are implemented, this should check those too.
```
GShotwell: I see this TODO here and above, is it worth breaking this check into a function so that it's more obvious in the future where new transform checks should go? That might also help as a kind of list of currently implemented transforms.

tests/testthat/test-cube-transforms.R
```diff
@@ -85,6 +85,75 @@ test_that("applyTransforms can return what is asked for", {
...
+    
+    expect_equivalent(applyTransforms(cat_array_cube), all)
+    expect_output(cat_array_cube, 
+                  "    CA\nCA    A  B\nmr_1  1  2\nmr_2  2  1\nmr_3  2  1")
+})
+
+test_that("can set transforms on a cube", {
+    transforms(pet_feelings) <- NULL
+    expect_null(transforms(pet_feelings))
+    # transforms(pet_feelings) <- Heading(name = "Fabulous new header", after = 0)
```
GShotwell: Is this commented code still relevant?

R/cube-transforms.R
```diff
@@ -150,22 +164,62 @@ setMethod("transforms", "VariableCatalog", function (x) {
...
 })
 
+#' @rdname Transforms
+#' @export
+setMethod("transforms<-", c("CrunchCube", "list"), function (x, value) {
+    dims <- dimensions(x)
+    dimnames <- names(dims)
+    
+    # check if the names of the dimensions and the names of the transforms line up
+    if (any(!names(value) %in% dimnames)) {
```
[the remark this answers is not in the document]
jonkeane: I totally agree, and try to be good about doing it, but here I forgot.

tests/testthat/test-cube-transforms.R
jonkeane: Nope, removed!

R/transform.R
jonkeane: Aaah, these comments are confusing, the todo should actually be above the checking if the insertions are null. I've moved them around (and spaced them out). The checks here are just checking if there are insertions, and if there aren't any returning `NULL` quickly. The todo is separate from that (and won't replace that when it's implemented...)

PR #175, opened 2017-12-11
Standardized Residuals for CrunchCubes

malecki:
This probably contains more commits than it needs to, but some of the test fixture changes apparently needed to be on here.

The expectations here are identical to those in https://github.com/Crunch-io/whaam/pull/1267. Doing it here revealed a bug in my other implementation, so yay. The simplest is the regular contingency table, which uses the standardized residuals of a χ² test. In R `stdres` is a member of the object returned by `chisq.test`, rather than a method. I chose to implement a method `rstandard` after that in [base R stats]( https://www.rdocumentation.org/packages/stats/versions/3.4.3/topics/influence.measures) for _models_. If people want a different name you could, but the idea of a crunch-cube being effectively a “model” of a contingency table is a suitable abstraction, and can be further extended to do columnwise or other comparisons.

The code handles only three cases: all-categoricalish, one MR on either of two axes; and two MR. More complicated tables will need to be handled by slicing before calling `rstandard`.


tests/testthat/test-cube-residuals-zed-scores.R
```diff
@@ -0,0 +1,156 @@
...
+          -28.64461124699,34.3223793320432),
+        dims = catarray_dims)
+    expect_equal(rstandard(catarray), out)
+})
+
+test_that("residuals for catarray by cat", {
+    expect_error(rstandard(catarray_by_cat), paste0(
+                 "Cannot compute residuals with more than two dimensions. Pick ",
+                 "a slice to evaluate"))
+    # TODO: check a slice
```
nealrichardson: If you did this TODO, you'd see that `[.CrunchCube` isn't implemented. See https://github.com/Crunch-io/rcrunch/compare/cube-slice for a start at that from a year ago.

Since I'm not totally clear on the value of this PR, I'm not certain whether that limitation is important for it or not.

R/cube-residuals.R
```diff
@@ -0,0 +1,141 @@
+# standardized residuals formed by the subvariable-wise margins
+# the internal chisq tests use rowSums and colSums, but we need
+# the more abstract 'margin' (specifically cubeMarginTable) to
+# get the right numbers for multiple response.
+standardizedMRResiduals <- function(cube, types){
+    # grab the dims from as.array because sometimes cat x MR has one too many
```
nealrichardson: Really? Why?

tests/testthat/test-cube-residuals-zed-scores.R
malecki: 2d residuals (a 'table') are what other clients care about. generalizing is easy and fun in R but not really necessary for any known needs here.

tests/testthat/test-cube-residuals-zed-scores.R
```diff
@@ -0,0 +1,156 @@
...
+
+test_that("residuals for MR with (old style) selected arrays errors helpfully", {
+    expect_error(rstandard(mr_by_cat_selected_array), paste0(
+        "rstandard is not implemented with CrunchCubes that use selected ",
+        "arrays. Selected arrays have been deprecated, please recreate ", 
+        "your cube using `as_selected\\(\\)` around multiple response variables."))
+})
+
+
+test_that("residuals for catarray", {
```
malecki: oh, yeah, this doesn't make sense. remove this assertion these are not sane values.

R/cube-residuals.R
malecki: poorly worded. the cube itself has more dims than `as.array(cube)` does. The desired ndims is that of the array

nealrichardson: _You may be interested to know_ that the `dim()` method on `CrunchCube` handles this for you: https://github.com/Crunch-io/rcrunch/blob/master/tests/testthat/test-cube-mr.R#L13-L23

I'll verify and clean this up here.

PR #176, opened 2017-12-12
Implement Head and Tail methods

GShotwell:
Creates head and tail methods for Crunch Datasets, Dataframes, and Variables, this also implements numeric subsetting on dataset rows.


nealrichardson:
This is currently blocked due to overly strict validation on the csv export endpoint on the server.

GShotwell:
The ticket for the validation is: https://www.pivotaltracker.com/story/show/154086828

Remaining todos on the R side are:
- Generate mocks for numeric row subsetting
- Check test coverage

R/as-data-frame.R
```diff
@@ -192,3 +192,48 @@ as.data.frame.FilterCatalog <- function (x,
...
+#' @export
+setMethod("tail", "CrunchDataset", function (x, n=6L, ...) {
+    as.data.frame(x[tail(seq_len(nrow(x)), n),], force=TRUE)
+})
+
+#' @rdname head-tail
+#' @export
+setMethod("tail", "CrunchDataFrame", function (x, n=6L, ...) {
+    return(tail(attr(x, "crunchDataset")))
+})
```
jonkeane: It doesn't look like these would respect the ordering of the CrunchDataFrame which is in the `row.order` attribute.

I don't think it would be too difficult to add that, though we haven't been using that feature too much (it was designed to make merging onto simple feature style geographies possible, but I haven't had the time to push forward on that recently)

R/as-data-frame.R
```diff
@@ -192,3 +192,48 @@ as.data.frame.FilterCatalog <- function (x,
...
+#' Head and tail methods for Crunch objects. See [utils::head()] for more details.
+#'
+#' @param x a CrunchDataset, CrunchDataFrame, or CrunchVariable
+#' @param n a single integer representing the length of the returning object.
+#' @param ... ignored
+#' @name head-tail
+#' @aliases head tail
+NULL
+
+#' @rdname head-tail
```
jonkeane: Since the `head`/`tail` methods that we are imitating are based on S3 classes, we should probably match those and define this (and similar for the following method definitions) as:

`head.CrunchDataset <- function ...`

Then remove the lines in AllGenerics.R that make the generics. This will prevent masking when the package is loaded.

GShotwell:
It looks like this PR is blocked by some odd behaviour around datasets which have an exclusion. When you have a dataset with an exclusion, then try to pass a logical row filter, the API returns incorrect data. We think that this is because the row filter is being applied to the unexcluded data. 

Decision is to wait on this method until we implement https://www.pivotaltracker.com/n/projects/931610/stories/151013797 at which point we can refactor all of the `row_number` filtering to use a more concise version of numeric subsetting.

nealrichardson:
If row filtering on the server doesn't respect the exclusion filter, that sounds like a bug that needs fixing (independent of this PR).

Issue #167, opened 2017-11-29
Get details about a user's interaction with a dataset


jonkeane:
A specific need was: the last time a user accessed a dataset. This can be accomplished with the code below (but should be expanded, validated, etc.:
```
last_accessed <- function (email, dataset_id) {
    users <- getAccountUserCatalog()
    user_url <- urls(users[email])
    ds_url <- shojiURL(UserEntity(crGET(user_url)), "catalogs", "datasets")
    
    httpcache::dropCache(ds_url)
    cat_df <- catalogToDataFrame(ShojiCatalog(crGET(ds_url)))
    one_ds <- cat_df[cat_df$id == dataset_id,]
    
    return(one_ds$access_time)
}
```

nealrichardson:
That only will work if `email` corresponds to a member of the current user's account, which is probably not the use case that was envisioned. 

Instead, dataset editors can access a "users" catalog on the dataset, which has access time on it. Use that.

nealrichardson:
See http://docs.crunch.io/#catalogs

PR #177, opened 2017-12-13
Functions to make it easy to find the last time that users accessed datasets

jonkeane:
I am open to suggestions on better names for these two functions, I'm not absolutely satisfied with the ones here.

Resolves issue #167


R/user-access.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+#' @param project a character of the project name to gather metadata from
+#' @param ... additional arguments passed to [userInfo()] (e.g. user_fields)
+#'
+#' @return a dataframe with information about all users for a particular dataset
+#' @export
+projectUserInfo <- function (project, ...) {
+    sets <- selectDatasetCatalog(project = project)
+
+    list_of_dfs <- base::lapply(sets, userInfo, ...)
```
GShotwell: Is `base::` required because `lapply` behaves differently on `sets`?

R/user-access.R
```diff
@@ -0,0 +1,39 @@
...
+#' have access to a particular dataset.
+#'
+#' @param dataset_entity a dataset object or a dataset entity (e.g. a single
+#' entry from `selectDatasetCatalog()`)
+#' @param user_fields metadata fields about users to include in the dataframe
+#' (default: c("name", "email", "access_time"))
+#'
+#' @return a dataframe with information about all users for a particular dataset
+#' @export
+userInfo <- function (dataset_entity, user_fields = c("name", "email", "access_time")) {
```
GShotwell: I think these function names are fine, but another option is to have a single `userInfo` method which  would return different things depending on whether a dataset or project were passed to it. This might be a good way to go because we might want to add equivalent functions for other objects in the future.

R/user-access.R
```diff
@@ -0,0 +1,39 @@
...
+    users_catalog <- ShojiCatalog(crGET(paste0(self(dataset_entity),"users/")))
+    all_users <- catalogToDataFrame(users_catalog, keys = user_fields)
+    all_users$dataset <- name(dataset_entity)
+    all_users$dataset_id <- id(dataset_entity)
+
+    return(all_users)
+}
+
+#' Get information about when users last any of the datasets in a project
```
GShotwell: This documentation looks duplicated, maybe collapse them into a single documentation entry?

R/user-access.R
nealrichardson: `dataset` not `dataset_entity` for consistency.

R/user-access.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+#' @param dataset_entity a dataset object or a dataset entity (e.g. a single
+#' entry from `selectDatasetCatalog()`)
+#' @param user_fields metadata fields about users to include in the dataframe
+#' (default: c("name", "email", "access_time"))
+#'
+#' @return a dataframe with information about all users for a particular dataset
+#' @export
+userInfo <- function (dataset_entity, user_fields = c("name", "email", "access_time")) {
+    users_catalog <- ShojiCatalog(crGET(paste0(self(dataset_entity),"users/")))
```
nealrichardson: Always avoid `paste0` to construct URLs in `crunch`. That's not how our API works. Use `shojiURL(dataset, "catalogs", "users")`.

nealrichardson: And I believe there already is a `UserCatalog` class, which may help here.

R/user-access.R
nealrichardson: Also, the function that gets the `UserCatalog` from a dataset should be called `users()`, for consistency with e.g. `datasets()`, `variables()`, `projects()`, etc. The conventions help us compose new features, and in principle make it easier for our users to figure out what a function would be called.

If you want to add a special print method for it, or a function (like this) that wraps and prints a special summary, that's fine.

R/user-access.R
nealrichardson: If you're doing `base::` then something is wrong.

R/user-access.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+#' Useful to get dataset-user interaction metadata about all of the users that have
+#' access to all of the datasets in a project.
+#'
+#' @param project a character of the project name to gather metadata from
+#' @param ... additional arguments passed to [userInfo()] (e.g. user_fields)
+#'
+#' @return a dataframe with information about all users for a particular dataset
+#' @export
+projectUserInfo <- function (project, ...) {
```
nealrichardson: Why is this not just `users(project)`? Projects should have the same/similar users catalog API as datasets (and if it doesn't currently, it *should*, so let's do it right--it's on the roadmap).

jonkeane: I've added a `users(project)` method that access that. Brian's use case needs the dataset user catalog to get the last_accessed info from, so I've subsumed that under a new `lastAccessed` helper function.

R/user-access.R
jonkeane: I was trying to avoid the fact that by default `lapply` for ShojiCatalogs iterating over the index rather than the entities in the catalog themselves (cf https://github.com/Crunch-io/rcrunch/blob/master/R/shoji-catalog.R#L181). 

I did some digging, and it looks like that's not actually necessary (anymore?). Unless there's something that depends on it that's not tested, and I'm not seeing, we can remove that `lapply` definition without damaging anything.

R/user-access.R
nealrichardson: So Brian is looking to see the last time a user accessed any dataset within a project? That should be the last_accessed time in the project's user catalog. If it isn't there, we should make it so. Getting every dataset in the project dataset catalog, then getting each dataset's user catalog, isn't the right way to support that.

R/user-access.R
nealrichardson: You don't want to load every dataset in the catalog. As currently implemented, loading a dataset eagerly fetches the variables catalog and variable order, and you don't want or need that. I could suggest a couple of more efficient ways to do that, but as I just commented below, I don't think that's something we should be doing at all.

R/user-access.R
nealrichardson: Agree with this sentiment, and it should be called `users()`. This is in keeping with our current/upcoming dataset/project API changes, which will make them more consistent.

R/user-access.R
jonkeane: From what I understand, he's actually looking for the last time each user accessed each dataset within a project.

nealrichardson: Is that a function we need to write? I'm not convinced that that belongs in the library (though I'm open to arguments). And if it does belong as something that we support, then it should be computed server-side and not require N requests to build. 

Seems like if we have natural, good methods, he (and other users) should be able to compose what they want easily, even if we don't deem it a core piece of functionality. Sounds like he wants `lapply(datasets(project), users)`, or some formulation of that, and he can `as.data.frame` the catalogs and `rbind` or whatever, `purrr`ing and piping or however he wants. Perhaps we should do enough make that work (revisiting the `lapply` method definition you mentioned, for example) and leave it at that.

I had planned to make dataset loading be lighter/lazier as a followup to the variable folders work, and it sounds like a use case like this would be another reason to prioritize it.

GitBrianLaw: The motivating question is from the YG Ops folks in London. My understanding is that an .xls spreadsheet is as technical as they are willing to go and so I'm trying to automate building a spreadsheet for them to monitor Crunch usage - who looked and last time accessed.

nealrichardson: Thanks @GitBrianLaw. Can you elaborate on what they actually want to accomplish or know?

GitBrianLaw: Sure. I made this by hand https://docs.google.com/spreadsheets/d/1ymk4Pv20iNOranspBP9gXzuA-c9bxBd8FiacOjXV_KM/edit#gid=0
by clicking through the datasets b/c doug said "I need this ASAP" a few weeks ago. Now ops "loves it" and wants it going forward but doing it by hand is crazy. That's when I reached out to Jon. Clear-ish?

jonkeane: I totally agree that it's pretty peripheral, but my reasoning for writing and including it was that it would hopefully prevent other implementations that would hit the servers more than necessary (e.g. using `loadDataset()` and then calling `users` on that). I absolutely agree that this is something that we should support server side, which would solve this and a few other requests as well.

From the previous comments, I've got a draft that doesn't over-fetch dataset variables when iterating through a project (though at the cost of having to hack the `users/` endpoint url together). That would solve the immediate need, and we could improve it when we lighten up dataset loading. Additionally, when we get the right user-access history endpoint we could transition that function over to use that without too much disruption. I'm also happy to excise those helpers from the package, and help Brian craft.

nealrichardson: Yes, thank you. How I'd like to tackle this is (1) enable you to write a simple script to build your xls file whenever by exposing these APIs in R (what I said in https://github.com/Crunch-io/rcrunch/pull/177#discussion_r157080057), and (2) take this and reflect on what kinds of features we should support more fully in terms of usage analytics/client management. I don't think it is the business of the `crunch` R package to generate this bespoke report for you, but it should make it easy for you to assemble the pieces you want. 

Does that make sense?

GitBrianLaw: Totally agree and happy to have it implemented that way. Thanks for helping.

nealrichardson: @jonkeane the question for me is the boundary for what goes in the package (and thus is a feature we own and maintain) vs. what is user's code. That you've written it is great, but I suspect that some of what you written belongs on Brian's computer and not ours. I take the concern about efficiency and silently making it more efficient later, but this specific report looks a little too custom for our general purpose package.

jonkeane: I'll take out the helper functions, and send brian the code I have for grabbing what he needs.

PR #179, opened 2017-12-15
Update to use and require testthat 2.0.0

nealrichardson:
This deletes some test code that's now built into the official testthat release and gets off my personal fork of testthat on Jenkins. 

I recommend making a ticket and running this through the build system at least once to get a pass since it involves changes to how Jenkins runs the tests.


nealrichardson:
Appveyor failure is `Package required and available but unsuitable version: 'testthat'`, meaning that the mirrors haven't all propagated with the latest testthat yet. It hit cran two days ago apparently: https://cran.r-project.org/web/packages/testthat/index.html

PR #180, opened 2017-12-18
Store spelling language in package DESCRIPTION

jeroen:
Thank you for using the spelling package. The new version `spelling 1.1` stores the language in the package `DESCRIPTION` field.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,6 +34,7 @@ Suggests:
     rstudioapi,
     spelling
 VignetteBuilder: knitr
+Language: en-US
 RoxygenNote: 6.0.1
 Roxygen: list(markdown = TRUE)
 Collate: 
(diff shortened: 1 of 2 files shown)


nealrichardson:
Thanks, @jeroen, for this (and for `spelling` itself)! Two questions:

1. Do you know if there is there any urgency for us to release this patch to CRAN? Spelling errors shouldn't cause a `check` failure there, but I've had a CRAN package update submission rejected due to "noise" from `spelling` messages even though the build passed. 

2. I see some new spelling errors on Travis (where I do have the build set up to fail on spelling): `Error: Potential spelling errors: API's, dataset's, Dataset's, Snowden's`. Since those are all possessives, I'm wondering if that's an unintended side effect of https://github.com/ropensci/hunspell/blob/master/NEWS#L2 in the latest `hunspell` release. Or is that the intended effect, and now we should add those and also go fix the odd half-contractions from our WORDLIST (e.g. "didn", https://github.com/Crunch-io/rcrunch/blob/master/inst/WORDLIST#L114)?

jeroen:
Hi Neal. No there is no urgency to release a fix to CRAN. It will keep working.

Regarding your other point: yes it is correct the package will now check words with an apostrophe. I am ont a native speaker but I think the proper plural forms would be `APIs` and `datasets` ?

Either way just run `update_dictionary()` to whitelist these words if you want to keep them.

nealrichardson:
Thanks, I wasn't aware of `update_dictionary()`, that's helpful.

PR #181, opened 2018-01-10
Subtotals vignette

GShotwell:
Takes the R parts of Jon's Crunch blog post and turns it into vignette. Also includes changes to images and vignette data.


vignettes/subtotals.Rmd
```diff
@@ -0,0 +1,149 @@
...
+## We'll then reference saved objects in that as if we had just retrieved them
+## from the server
+library(crunch)
+load("vignettes.RData")
+options(width=120)
+```
+
+A common task in the Market research world is to collapse two or more categories together to see how the collapsed categories compare to one another. For example, if you asked people to rate their preference on a scale of 1-10 you might want to see how the people who provide a rating between 1 and 5 compare to those who rated it between 6 and 10. This goes by a number of names like _Top Box_, _Top Two Box_, or _Nets_. But in Crunch it is called _Subtotals_
+
+Subtotals can be applied to any Categorical Crunch Variables, and we can check if a variable has subtotals with the `subtotals()` function. If it returns `NULL`, there are no subtotals.
```
jonkeane: Super super minor, but I would change "Subtotals can be applied to any Categorical Crunch Variables" to "Subtotals can be applied to any Categorical Crunch Variable"

jonkeane: You might also want to say something like "and we can check if a variable has subtotals by using the `subtotals()` function on the variable. If it returns `NULL`, there are no subtotals. If there are subtotals, it returns an object with the properties of the subtotals (which are discussed in more detail below)."

vignettes/subtotals.Rmd
```diff
@@ -0,0 +1,149 @@
...
+Now, if we check `subtotals()`, we can see that we have saved them. In this output we see a few different aspects of subtotals: the `anchor` is the id of the category to put the subtotal after (matching the `after` argument in `Subtotal()`), name, aggregation functions and `args` which in the this case are the category ids to include in the subtotal.
+
+```{r new subtotals, eval = FALSE}
+subtotals(ds$manningknowledge)
+```
+```{r, echo = FALSE}
+sub_subtotals1
+```
+
+This shows up in the Categorical variable card like so:
```
jonkeane: This is pretty explicit already, especially with the photo, but it might be nice to toss webapp in here just to make clear what variable card we are referring to.

vignettes/subtotals.Rmd
```diff
@@ -0,0 +1,149 @@
...
+            categories = c("Somewhat disapprove", "Strongly disapprove"),
+            after = "Strongly disapprove"))
+```
+
+```{r check some categories, eval = FALSE}
+subtotals(ds$snowdenleakapp) <- approve_subtotals
+subtotals(ds$congapp) <- approve_subtotals
+```
+
+Notice here, because each of the categories for these variables has slightly different ids, the `args` in the output differs slightly. But, because we used category names when we were constructing our list of subtotals, when we store them on the variable itself Crunch does the right things and converts them over to the correct ids.
```
jonkeane: "does the right things" -> "does the right thing"

vignettes/subtotals.Rmd
```diff
@@ -0,0 +1,149 @@
...
+subtotals(ds$obamaapp)
+```
+```{r no subtotals print, echo = FALSE}
+sub_initial_subtotals
+```
+
+To add subtotals, we can save a list of `Subtotal` objects. Each `Subtotal` object has three things:
+
+* `name` the label to identify the subtotal
+* `categoires` the categories to add subtotal (here you can use either category names or category ids)
```
jonkeane: categoires -> categories

vignettes/subtotals.Rmd
```diff
@@ -0,0 +1,149 @@
...
+* `after` the category that the subtotal should follow (again, either category names or category ids)
+
```
jonkeane: Add info about the `position` argument here

jonkeane:
Ok, this looks good to me! I pulled the branch down to see how it looked in pkgdown, and noticed a few minor inconsistencies which I've corrected.

PR #182, opened 2018-01-13
Subtotals and headings now have a position argument

jonkeane:
The position argument takes top/bottom/relative as values. If relative (the default) is given, then an after argument must be present.

### DIFF
--- a/R/insertions.R
+++ b/R/insertions.R
@@ -170,6 +170,10 @@ setMethod("anchor", "Insertion", function (x) {
 
 # method for getting an anchor from a user-friendly abstracted Subtotal or Heading
 .convertAnchor <- function (x, var_categories) {
+    if (x$position != "relative") {
+        return(x$position)
+    }
+
     # map chars/nums to ids
     if (is.character(x$after)) {
         # TODO: better error if var is null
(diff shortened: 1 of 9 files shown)


R/insertions.R
```diff
@@ -205,7 +214,7 @@ setMethod("func", "Heading", function (x) return(NA))
 #' @rdname Insertions
 #' @export
 setMethod("anchors", "Insertions", function (x) {
-    f <- vapply(x, anchor, integer(1))
+    f <- vapply(x, anchor, 1)
```
nealrichardson: This line looks odd.

R/subtotals-and-headings.R
```diff
@@ -91,74 +95,89 @@
...
+    return(new("Subtotal", list(name = name,
+                                categories = categories,
+                                position = position,
+                                after = after)))
+}
+
+validatePosition <- function (position, after) {
+    if (position == "relative") {
+        # if position is realtive, make sure there is an after
+        if(is.null(after)) {
```
nealrichardson: Space between `if` and `(`

R/subtotals-and-headings.R
```diff
@@ -91,74 +95,89 @@
...
-                      ". Did you mean to make a Subtotal?")
-    } else {
-        val <- TRUE
-    }
+#' @rdname SubtotalsHeadings
+#' @export
+Heading <- function (name,
+                     position = c("relative", "top", "bottom"),
+                     after = NULL) {
+    # match.args position
```
nealrichardson: I'm not sure this comment adds any value

R/insertions.R
jonkeane: Oph, yeah. Since `anchor()` could now return either an integer or a character I've switched this to `unlist(lapply())`

PR #192, opened 2018-01-25
Added formulas argument to conditionalTransform

jonkeane:
Additionally, added more args checking and polished docs, error messages, etc.


R/conditional-transform.R
```diff
@@ -49,14 +51,25 @@
...
+                 " argument and ", dQuote("...")
+                 ,"\n","Please supply conditions in only one.")
+        }
+        formulas <- dot_formulas
+    }
     var_def <- Filter(Negate(is_formula), dots)
 
     if (length(formulas) == 0) {
-        halt("no conditions have been supplied; please supply formulas as conditions.")
+        halt("conditions must been supplied\n",
```
nealrichardson: Grammar seems off here, and \n seems inconsistent.

jonkeane: The new line I took from the style guide proposed here: http://style.tidyverse.org/error-messages.html

I'm happy to take it out, though I generally like the style that is proposed separating the error info from the hinting info. On second reading of that style those hints should be questions instead of "Please..." if we want to aim for the tidyverse style I'll change these to that format before I merge.

nealrichardson: /shrug

IMO new style guides are good but internal consistency is more important: things seem less broken when everything seems to have a single voice. If we want to change to hadleystyle, let's decide to do that and make that our voice everywhere. But _that_ really sounds like scope creep here :)

jonkeane: Yeah, absolutely. So maybe as a compromise between these two: we try and follow the tidyverse style phrasing without the formatting change for now, and then make a push in the future to make all of them consistent spacing (and listing out as many problems as can be seen at once instead of one at a time) separately on its own? Does that maintain enough voice consistency while making errors more helpful and make it easier for the (possible) error refresh?

nealrichardson: Sounds good; that, and capitalizing the first letter of the error messages (like sentences), and turning L70 "conditions must been supplied" into English.

PR #193, opened 2018-01-25
Change the default selections method for crtabs to `as_selected`

jonkeane:
`selected_array` is still available (for now) via the option `crunch.mr.selection`


tests/testthat/test-cube-with-array-selected-array.R
```diff
@@ -0,0 +1,158 @@
+context("Cubes with categorical array and multiple response (deprecated selected_array)")
```
nealrichardson: Why not move all of the old-way tests below into this file?

R/show.R
```diff
@@ -285,7 +285,13 @@ showMultitable <- function (x) {
     # eg remove selected_array()
     out <- c(out, "Column variables:",
              vapply(x@body$template, function (expr) {
-                 paste0("  ", formatExpression(expr$query[[1]]))
+                 if (names(expr$query[[1]]) == "each") {
+                     # if the first element of the query is each, then this is
+                     # an MR so take the second argument instead.
```
nealrichardson: Actually, it's any array variable, not just MR. Though maybe you can't put a categorical array in a multitable?

And anyway, looking at the test output again, I'm not sure there's any value in showing `as_selected(allpets)` instead of just `allpets`.  So if we're going to special-case in here, perhaps we should get that right. (Recognizing that this is scope creep here)

tests/testthat/test-cube-with-array-selected-array.R
jonkeane: Thanks, I missed removing a few old-way tests in `test-cube-with-array.R` when I made this file. Those have been removed (and checked to confirm they were in the old-way file) so `test-cube-with-array.R` is all `as_selected` and `test-cube-with-array-selected-array.R` is all `selected_array`

PR #195, opened 2018-01-25
Add vignette on combine cut interaction functions

GitBrianLaw:
Brian first draft of vignette on combine cut and interAction functions


vignettes/combine-cut-interact.Rmd
```diff
@@ -0,0 +1,140 @@
...
+```{r, eval = FALSE}
+categories(ds$age4)
+```
+```{r, echo = FALSE}
+cat(summary.age4.var, sep="\n")
+```
+
+And now we have a new Categorical type variable with the alias `age4`, the name "Age (4 categories)", and it's made of 4 categories.
+
+# Re-Combining Answer Choices
```
jonkeane: When I read re-combining at first, I found it a bit awkward. I wonder if maybe just "Combining answer choices" would be better here?

vignettes/combine-cut-interact.Rmd
```diff
@@ -0,0 +1,140 @@
...
+```{r, eval = FALSE}
+ds$boap_combined
+```
+```{r, echo = FALSE}
+show_boap_combined
+```
+
+We have created a new Multiple Response type variable with the alias `boap_combined`, the name "Approval of Obama on issues (Combined Subvariables)", which has 4 subvariables instead of the original 13.
+
+# Re-Combining Variables
```
jonkeane: Again, I think this might be ok just called "Combining variables"

vignettes/combine-cut-interact.Rmd
```diff
@@ -0,0 +1,140 @@
...
+Besides re-combining answer choices, we can also re-combine variables. For example, if our dataset has a categorical type variable for gender and another categorical type variable for Age, then we can cross these two together into a new variable using the `interactVariables()` function (named after the use of 'interaction terms' in regression). 
+
```
jonkeane: It might be nice to add a more elaborated use case in here.

vignettes/combine-cut-interact.Rmd
GitBrianLaw: Yeah. Re-combining is not great. Let's switch

vignettes/combine-cut-interact.Rmd
GitBrianLaw: yup

PR #196, opened 2018-01-26
Allow modifying anchors with anchor(subtotal)<- syntax

jonkeane:
Refactored some `PATCH`es to use more package-standard approaches.


tests/testthat/test-subtotals-and-headings.R
```diff
@@ -184,4 +234,47 @@ with_test_authentication({
...
+        expect_equal(names(subtotals(ds$v4)), c("This is a subtitle", "B alone",
+                                                "C alone", "B+C"))
+        expect_equal(anchors(subtotals(ds$v4)), c("top", 1, 2, "bottom"))
+        expect_equal(arguments(subtotals(ds$v4)[[1]]), NA)
+        expect_equal(arguments(subtotals(ds$v4)[[2]]), 1)
+        expect_equal(arguments(subtotals(ds$v4)[[3]]), 2)
+        expect_equal(arguments(subtotals(ds$v4)[[4]]), c(1,2))
+
+        # changing names
+        expect_silent(name(subtotals(ds$v4)[[1]]) <- "The new subtitle")
```
nealrichardson: why all the `expect_silent`? seems to mask (from my readability) the operations you're doing that are changing state, for little value.

R/subtotals-and-headings.R
```diff
@@ -218,18 +218,10 @@ setMethod("subtotals<-", c("CrunchVariable", "ANY"), function (x, value) {
...
-    if (!is.null(old_inserts)) {
-        inserts <- modifyCats(old_inserts, inserts)
-    }
-
-    bd <- wrapEntity("view" = list("transform" = list("insertions" = inserts)))
-    crPATCH(self(x), body=toJSON(bd))
     dropCache(cubeURL(x))
+    bd <- list("transform" = list("insertions" = inserts))
+    # setEntitySlot manages old inserts so they are not duplicated
+    ent <- setEntitySlot(entity(x), "view", bd)
```
nealrichardson: I'd put the `dropCache` statements after doing the PATCH (`setEntitySlot`)

tests/testthat/test-subtotals-and-headings.R
jonkeane: Old habits — I've taken them out.

R/subtotals-and-headings.R
jonkeane: That makes sense (and my first intuition as well), I was modeling this after https://github.com/Crunch-io/rcrunch/blob/master/R/variable.R#L115-L116, which I've changed to drop the cache after as well.

nealrichardson: Good call, thanks.

PR #197, opened 2018-01-26
Doc cleanup

jonkeane:
Added missing man pages to pkgdown; edited the category classes vignette to reflect the changes (and correct a previous mistake about which methods are available for the container classes)


R/as-data-frame.R
```diff
@@ -154,7 +154,7 @@ csvToDataFrame <- function (csv_df, crdf) {
 #' as.data.frame(vars, row.names = urls(vars))
 #' }
 #'
-#' @rdname catalog-to-data-frame
+#' @name as.data.frame.catalog-method
```
nealrichardson: This makes for an ugly URL/filename. Think of something better?

PR #199, opened 2018-01-30
Population counts

### LINKED ISSUE #194: let me get/set dataset `population`
https://github.com/Crunch-io/apidocs-sphinx/pull/19/files

GShotwell:
Creates R access to get and set target population. Closes #194


R/dataset.R
```diff
@@ -88,6 +88,59 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+    if (missing(size)) {
+        if (is.null(pop$size)) {
+            stop("Dataset does not have a population, please supply one")
+        }
+        size <- pop$size
+    }
+    if (!(magnitude %in% c(3, 6, 9))){
+        stop("Magnitude must be either 3, 6, or 9")
+    }
+    if (!identical(size, pop$size) |
```
nealrichardson: You shouldn't need this `if`; `settings<-` will do this check for you.

tests/testthat/test-dataset-entity.R
```diff
@@ -27,6 +27,26 @@ with_mock_crunch({
             '{"notes":"Ancillary information"}')
     })
 
+    test_that("Population setting", {
+        expect_equal(popSize(ds), 90000000)
+        expect_equal(popMagnitude(ds), 3)
+        expect_silent(setPopulation(ds, popSize(ds), popMagnitude(ds)))
```
nealrichardson: Use `expect_no_request` instead of `expect_silent` here: it's clearer what you're asserting.

R/dataset.R
```diff
@@ -88,6 +88,56 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+#' @export
+setMethod("popSize", "CrunchDataset", function (x) {
+    return(settings(x)$population$size)
+})
+
+#' @rdname population
+#' @export
+setMethod("popMagnitude", "CrunchDataset", function (x) {
+    return(settings(x)$population$magnitude)
+})
```
jonkeane: Would it make sense to expose `popSize<-` and `popMagnitude<-` methods to let users update size/magnitude? I'm not 100% set that we should add it, just a pattern I know we support elsewhere (and in classic R fashion: why have one way to do something when you can have multiple?). There's a complication for the initial setting that you need both at the same time, but after they're set, you should be able to `PATCH` either independently. 

Something like:
```
setMethod("popSize<-", "CrunchDataset", function (x, value) {
    if (is.null(settings(x)$population$size)) {
        halt("Please set with `setPopulation(ds, ...)`)
    }
    settings(x)$population$size <- value
    return(invisible(x))
}
```

R/dataset.R
```diff
@@ -88,6 +88,56 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+            stop("Dataset does not have a population, please supply one")
+        }
+        size <- pop$size
+    }
+    if (!(magnitude %in% c(3, 6, 9))){
+        stop("Magnitude must be either 3, 6, or 9")
+    }
+    settings(x)$population <- list(magnitude = magnitude, size = size)
+    invisible(return(x))
+})
```
jonkeane: Is there a way for me to remove a population? Would `setPopulation(ds, NULL, NULL)` do it? When you remove population in whaam, the whole entity `population` is set to NULL (rather than `population: {magnitude: NULL, size: NULL}`, which I suspect would happen with the command here), which we should copy if possible.

R/dataset.R
GShotwell: My intuition is that if we're going to do an assignment method it should work on all the datasets and having the user switch back and forth between two different ways of setting something based on the dataset state is a bit confusing.  It seems like this is mostly going to come up with people setting populations on a dataset for the first time so in that case we could just default to a magnitude of thousands. What about something like this?:
```
setMethod("popSize<-", "CrunchDataset", function (x, value) {
    if (is.null(settings(x)$population$magnitude)) {
        warning("magnitude not set, defaulting to thousands")
        setPopulation(x, value, 3)
    }
    settings(x)$population$size <- value
    return(invisible(x))
}
```
This doesn't work for setting magnitude because we don't want to default to a population, but that's probably not going to come up as much.

jonkeane: That works for me. I was mostly wanting to give R users an easy way to adjust both the size and magnitude without adjusting the other. So for `popMagnitude<-` we would just outright error if it's not set, but users could still change magnitudes easily after the fact.

R/dataset.R
```diff
@@ -88,6 +88,77 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+#' population estimates from survey percentages. This function lets you work with
+#' the population size and magnitude.
+#'
+#' @param x a Crunch Dataset
+#' @param value For the setters, the `size` or `magnitude` to be set
+#' @param size the target population size, to remove a population set to `NULL`
+#' @param magnitude the order of magnitude with which to display the population
+#' size. Must be either `3`, `6`, or `9` for thousands, millions, and billions respectively.
+#' @return `popSize` and `popMagnitude` return the population size or
+#' magnitude. `setPopultion` returns the modified dataset.
```
nealrichardson: setPopultion

R/dataset.R
```diff
@@ -88,6 +88,77 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+        if (is.null(pop$magnitude)) {
+            warning("Dataset magnitude not set, defaulting to thousands")
+            magnitude <- 3
+        } else {
+            magnitude <- pop$magnitude
+        }
+    }
+    if (missing(size)) {
+        if (is.null(pop$size)) {
+            stop("Dataset does not have a population, please set one before attempting to change magnitude")
```
nealrichardson: use `halt`

nealrichardson: Why do we care that someone is setting magnitude before setting size? Does something break?

GShotwell: It's really that you need to supply both size and magnitude together and there's not a sensible way to supply a default population. So if there's no magnitude we set it to thousands and warn the user, but it there's no population we just error.

R/dataset.R
GShotwell: I updated this to accept and return NULL values.

R/dataset.R
```diff
@@ -88,6 +88,80 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+    }
+    if (missing(size)) {
+        if (is.null(pop$size)) {
+            halt("Dataset does not have a population, please set one before attempting to change magnitude")
+        }
+        size <- pop$size
+    }
+    if (is.null(size)) {
+        size <- 0
+    }
```
jonkeane: If the size is being set to `NULL` there should be no `size` and no `magnitude` attributes but rather `population` should be `NULL`. That's what whaam does when population is removed and is more in line with other settings. This might be more easily accomplished by making a method with `setMethod("setPopulation<-", c("CrunchDataset", "NULL"), function ...`

An example of something that is similar is geography metadata: a variable that has a attached geography information will have a `view.geodata` attribute that includes things like `geodatum` `feature_key` etc. but when you set that to `NULL` in R, it empties out `view.geodata` to be empty: https://github.com/Crunch-io/rcrunch/blob/master/R/geo.R#L63-L70 rather than just deleting the individual `geodatum` and `feature_key` attributes.

R/dataset.R
```diff
@@ -88,6 +88,80 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+    if (is.null(magnitude) || !(magnitude %in% c(3, 6, 9))){
+        halt("Magnitude must be either 3, 6, or 9")
+    }
+    settings(x)$population <- list(magnitude = magnitude, size = size)
+    invisible(return(x))
+})
```
jonkeane: This looks pretty complicated (although it looks like accomplishes what we want).

One alternative to clean this up, that would also follow a couple of other patterns would be to have popSize and popMagnitude set their respective portions separately. This would make the `PATCH`es a bit cleaner because they would be only changing the one thing that changed rather than resending the same value.

R/dataset.R
```diff
@@ -88,6 +88,80 @@ setMethod("notes<-", "CrunchDataset", function (x, value) {
...
+NULL
+
+#' @rdname population
+#' @export
+setMethod("popSize", "CrunchDataset", function (x) {
+    size <- settings(x)$population$size
+    if (size == 0) {
+        return(NULL)
+    }
+    return(size) #the API value 0 stands for "No Population set"
```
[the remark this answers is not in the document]
GShotwell: Ah, I think this is one of the things which has changed  prior to https://www.pivotaltracker.com/story/show/155015518 I think you the only way you could clear values for population size was to set it to zero.

R/dataset.R
GShotwell: That's a good idea. The reason I went this route is that you have to send both size and magnitude together if there's no population set on the dataset. So that means that we need to send them both regardless, and I thought it was simpler to do it this way.

tests/testthat/test-dataset-entity.R
```diff
@@ -27,6 +27,71 @@ with_mock_crunch({
...
+        expect_PATCH(popSize(ds) <- 6000,
+            "https://app.crunch.io/api/datasets/1/settings/",
+            '{"population":{"magnitude":3,"size":6000}}'
+            )
+        expect_PATCH(popMagnitude(ds) <- 6,
+            "https://app.crunch.io/api/datasets/1/settings/",
+            '{"population":{"magnitude":6,"size":90000000}}'
+        )
+    })
+    test_that( "setPopulation errors correctly", {
```
nealrichardson: Delete weird leading space

tests/testthat/test-dataset-entity.R
```diff
@@ -335,6 +400,26 @@ with_test_authentication({
...
+            expect_equal(popSize(ds), 54321)
+            expect_equal(popMagnitude(ds), 3)
+            ds <- setPopulation(ds, magnitude = 6)
+            expect_equal(popSize(ds), 54321)
+            expect_equal(popMagnitude(ds), 6)
+            ds <- setPopulation(ds, size = NULL, magnitude = 6)
+            expect_null(popSize(ds))
+            expect_null(popMagnitude(ds))
+
+            expect_error(ds <- setPopulation(ds, size = 12345, magnitude = NULL),
```
nealrichardson: Move to mock test? Doesn't require real server does it?

jonkeane:
🐴 is indeed well beaten! I'll add NEWS post-merge

PR #201, opened 2018-02-06
A set of `as.*` methods to create derived variables …

jonkeane:
…that have been converted to new types (eg text to numeric, numeric to datetime, etc.)


R/variable-as-methods.R
```diff
@@ -0,0 +1,143 @@
...
+#' Each type of Crunch variable (text, numeric, categorical, etc.) has an `as.*`
+#' function (`as.Text`, `as.Numeric`, and `as.Categorical` respectively) that 
+#' takes the input given as `x`, and makes a new derived variable that is now of 
+#' the type specified. See below for detailed examples.
+#' 
+#' For `as.Text` and `as.Numeric`, aliases to the R-native functions 
+#' `as.character` and `as.numeric` are provided for convenience.
+#'
+#' @param x a Crunch variable to derive and convert to a new type
+#' @param format for `as.Datetime`, when the variable in `x` is a text or
```
GShotwell: I found this a bit confusing, possibly because of the multiple uses of the word "format". Maybe something like "When x is text or categorical, the new date-time variable's format"?

R/variable-as-methods.R
```diff
@@ -0,0 +1,143 @@
...
+#'
+#' @param x a Crunch variable to derive and convert to a new type
+#' @param format for `as.Datetime`, when the variable in `x` is a text or 
+#' categorical variable, the format that the date-time is formatted in (default: 
+#' \code{"\%Y-\%m-\%d \%H:\%M:\%S"}); for `as.Text` and `as.Categorical`, the 
+#' format that the datetime should be formatted as (default: 
+#' \code{"\%Y-\%m-\%d \%H:\%M:\%S"}).
+#' @param resolution for `as.Datetime`, when the variable in `x` is a numeric 
+#' variable, the resolution of the number (e.g. `"ms"` for milliseconds, `"s"` 
+#' for seconds, etc. link)
```
GShotwell: Is this an extra word?

R/variable-as-methods.R
```diff
@@ -0,0 +1,143 @@
...
+})
+
+#' @rdname variable-as-methods
+#' @export
+setMethod("as.Datetime", "CrunchVariable", function (x, format, resolution) {
+    haltIfArray(x, callingFunc = "as.Datetime()")
+    
+    # datetimes are special, so each source type must be determined
+    if (is.Numeric(x)) {
+        # TODO: validate resolution
```
GShotwell: Might be worth adding a little to this comment explaining the validation you wanted to do.

R/variable-as-methods.R
```diff
@@ -0,0 +1,143 @@
...
+#' @export
+as.double.CrunchVariable <- function (x, ...) as.Numeric(x)
+
+#' @rdname variable-as-methods
+#' @export
+as.character.CrunchVariable <- function (x, ...) as.Text(x)
+
+# halt if variable is an array variable. If callingFunc is provided, provide the function that
+# the user used to get to this point. 
+haltIfArray <- function (variable, callingFunc) {
```
GShotwell: This is a great little function. I wonder if we want to start collecting these kinds of functions in one place, like a `utils.R` file. I feel like there are tons of them scattered around but since they are defined next to the code which they were first written to support they can be hard to find/remember.

jonkeane: There is a misc.R which I moved this to.

R/variable-as-methods.R
jonkeane: Was a thing to remind me to add a link (which I missed). Done now.

jonkeane:
Ok, these should be cleaned up now, I added a check when updating variables that `halts` if there are `cast` (and the datetime-specific) ZCL functions in the update. There are likely other ZCL functions that we would want to `halt` for there, possibly even halt for all derivations, but for now this prevents someone from accidentally doing `ds$var <- as.Text(ds$var)`

GShotwell:
This looks good to me. Thanks for adding the `ds$var <- as.Text(ds$var)` check

tests/testthat/test-variable-as-methods.R
```diff
@@ -0,0 +1,228 @@
...
+        
+        ds$num_times_years <- rep(1, 20)
+        ds$times_from_years <- as.Datetime(ds$num_times_years, resolution = "Y")
+        expect_true(is.derived(ds$times_from_years))
+        expect_true(is.Datetime(ds$times_from_years))
+        # need rollup() here because as.vector doesn't currently support other rollups
+        expect_equal(as.vector(rollup(ds$times_from_years, "D")), 
+                     as.Date(rep("1971-01-01", 20)))
+    }) 
+})
```
jonkeane: test comment for slobodan.

tests/testthat/test-variable-as-methods.R
```diff
@@ -0,0 +1,228 @@
...
+    
+    test_that("datetime to text", {
+        ds$v5_text <- as.Text(ds$v5)
+        expect_true(is.derived(ds$v5_text))
+        expect_true(is.Text(ds$v5_text))
+        expect_equal(as.vector(ds$v5_text), 
+                     format(as.vector(ds$v5), format = "%Y-%m-%d %H:%M:%S"))
+    }) 
+    
+    test_that("datetime to text", {
```
nealrichardson: *categorical

tests/testthat/test-variable-as-methods.R
```diff
@@ -0,0 +1,228 @@
...
+        expect_true(is.Categorical(ds$v5_cat))
+        expect_equal(as.vector(ds$v5_cat), 
+                     as.factor(
+                         format(as.vector(ds$v5), format = "%Y-%m-%d %H:%M:%S")
+                     ))
+    }) 
+    
+    
+    test_that("text to datetime", {
+        ds$text_times <- format(as.vector(ds$v5), format = "%Y-%m-%d")
```
nealrichardson: instead of `as.vector(ds$v5)`, you can just `df$v5` (the R version you uploaded)

R/variable-as-methods.R
```diff
@@ -0,0 +1,133 @@
...
+
+#' @rdname variable-as-methods
+#' @export
+setMethod("as.Text", "CrunchVariable",
+          function (x, format = "%Y-%m-%d %H:%M:%S") {
+    haltIfArray(x, callingFunc = "as.Text()")
+    
+    # datetimes are special
+    if (is.Datetime(x)) {
+        return(zfuncExpr("format_datetime", x, list(value = format)))
```
nealrichardson: I think the default `format` should depend on the `resolution` of the datetime variable.

R/variable-as-methods.R
```diff
@@ -0,0 +1,133 @@
...
+#' @rdname variable-as-methods
+#' @export
+setMethod("as.Categorical", "CrunchVariable",
+          function (x, format = "%Y-%m-%d %H:%M:%S") {
+    haltIfArray(x, callingFunc = "as.Categorical()")
+    
+    # datetimes are special
+    if (is.Datetime(x)) {
+        # no direct datetime to cat ZCL function, but can just cast the string
+        return(zfuncExpr("cast",
```
nealrichardson: instead of `return`ing here, 

    x <- zfunc("format_datetime", x, list(value = format))

tests/testthat/test-variable-as-methods.R
```diff
@@ -0,0 +1,228 @@
...
+        expect_true(is.derived(ds$times_from_secs))
+        expect_true(is.Datetime(ds$times_from_secs))
+        expect_equal(as.vector(ds$times_from_secs), 
+                     as.POSIXlt(rep("1976-01-01", 20), tz = "UTC"))  
+        
+        ds$num_times_years <- rep(1, 20)
+        ds$times_from_years <- as.Datetime(ds$num_times_years, resolution = "Y")
+        expect_true(is.derived(ds$times_from_years))
+        expect_true(is.Datetime(ds$times_from_years))
+        # need rollup() here because as.vector doesn't currently support other rollups
```
nealrichardson: What's that? Are we missing something in as.vector?

jonkeane: Yeah, currently rollups other than day, seconds, and milliseconds return NA in R. I'll make a separate issue that we can tackle separately. I need to dig into POSIX[cl]t to see if there are good ways to represent rolled up dates like that.

PR #204, opened 2018-02-20
rollupResoluton() setter

### LINKED ISSUE #188: Add ability for rollupResolution to set the default rollup on the server
Circa: https://github.com/Crunch-io/rcrunch/blob/master/R/expressions.R#L284-L303

GShotwell:
Closes #188


R/expressions.R
```diff
@@ -290,6 +292,8 @@ rollup <- function (x, resolution=rollupResolution(x)) {
     return(zfuncExpr("rollup", x, list(value=resolution)))
```
jonkeane: super minor, but spaces around the `=` would be good here.

tests/testthat/test-expressions.R
```diff
@@ -144,6 +144,13 @@ with_mock_crunch({
         print(ds$birthyr * 3 + 5)
         print(3 * (ds$birthyr + 5))
     })
+
+    test_that("rollup functions generate expected requests", {
+        expect_identical(rollupResolution(ds$starttime), "s")
+        expect_PATCH(rollupResolution(ds$starttime) <- "M",
+            'https://app.crunch.io/api/datasets/1/variables/starttime/',
+            '{"view":{"rollup_resolution":"M"}}')
+    })
```
jonkeane: Would be nice to have tests of the validation failing.

R/expressions.R
```diff
@@ -298,6 +302,16 @@ rollupResolution <- function (x) {
     }
 }
 
+#' @rdname expressions
+#' @export
+setMethod("rollupResolution<-",  "DatetimeVariable",
+    function(x, value = c('Y', 'M', 'D', 'h', 'm', 's', 'ms')) {
```
jonkeane: This was literally just merged while you are out (so no wonder it's not here!), but we have a function that will automatically validate rollups. https://github.com/Crunch-io/rcrunch/blob/master/R/misc.R#L229-L237

I know that in general a list like this and then `match.arg()` is the right pattern, in this case we aren't getting much more than using the validator that exists. Specifically, the man pages for `*<-` methods don't include those values as far as I can see.

tests/testthat/test-expressions.R
```diff
@@ -144,6 +144,13 @@ with_mock_crunch({
...
 })
 
 with_test_authentication({
```
jonkeane: I'm not sure if it's worth the added functional test here, but it might be good to add one — just to make sure that the API doesn't change without us realizing. 

If it's well tested in backend-http, and it would be slow here, then maybe we should skip here.

tests/testthat/test-expressions.R
GShotwell: Is this still necessary after switching to `validateResolution`? I'm assuming that function is tested

tests/testthat/test-expressions.R
```diff
@@ -327,4 +327,10 @@ with_test_authentication({
         skip("'which' isn't implemented correctly")
         expect_identical(which(duplicated(ds$v4)), 3:20)
     })
+
+    test_that("rollupResolution can be set", {
+        expect_null(rollupResolution(ds$v5))
+        expect_silent(rollupResolution(ds$v5) <- "M")
```
jonkeane: Thanks for adding these tests. For consistency, I would remove this `expect_silent()`, since any warnings or errors will propagate and fail the build regardless. This isn't my first tendency either, I usually do `expect_silent` explicitly, but we've eschewed it many (all?) other places in the package.

tests/testthat/test-expressions.R
jonkeane: It's tested as it's called from other functions (I should have added separate tests of `validateResolution` when I introduced it, but didn't, my bad). We should probably separate out tests for `validateResolution` and test that separately. I also am of the opinion that we should also test that `validateResolution` is being called in the more user-facing function as well — just in case something happens to the call to validate, we can catch that and correct it. The tests of `rollupResolution` don't need to cover all possible bad inputs, but at least one to ensure that we do hit `validateResolution` is good.

tests/testthat/test-expressions.R
```diff
@@ -320,4 +330,15 @@ with_test_authentication({
...
         expect_identical(which(duplicated(ds$v4)), 3:20)
     })
+
+    test_that("rollupResolution can be set", {
+        #ensure that only the rollup resolution changes
+        body <- ds$v5@tuple@body
+        expect_null(rollupResolution(ds$v5))
+        rollupResolution(ds$v5) <- "M"
+        expect_identical(rollupResolution(ds$v5), "M")
+        comp <- mapply(identical, body, ds$v5@tuple@body, SIMPLIFY = TRUE)
```
jonkeane: What if we ran this as `mapply(expect_equal...` or, even ran it in a for loop?

jonkeane: Also does this v5 already have other view settings? Maybe assert what those are before the resolution setting and then also that they are the same after. I think that would be a bit clearer to our future selves.

nealrichardson: Right, it's not that we're worried that another `body` field is edited. That is covered elsewhere. We were concerned that if body were like

```
list(
    name="something",
    ...,
    view=list(
        some_setting=TRUE
    )
)
```

that `view$some_setting` was still unchanged.

PR #206, opened 2018-02-22
List datasets RStudio Addin

GShotwell:
Adds a shiny gadget to `listDatasets` which allows the user to select which dataset they want to load. The gadget outputs the text of the correct `loadDataset()` call.


R/get-datasets.R
```diff
@@ -44,18 +44,27 @@ datasets <- function (x=getAPIRoot()) {
...
 #' @return Character vector of dataset names, each of which would be a valid
 #' input for [loadDataset()]
 #' @export
 listDatasets <- function (kind=c("active", "all", "archived"), project=NULL,
-                          refresh=FALSE) {
-    dscat <- selectDatasetCatalog(kind, project, refresh)
-    return(names(dscat))
+    refresh=FALSE, shiny = FALSE) {
+    if (shiny) {
+        if (rstudioapi::isAvailable("0.99.878")) {
```
nealrichardson: Since `rstudioapi` is under Suggests, you can't assume that it is installed here. https://github.com/Crunch-io/rcrunch/blob/master/R/auth.R#L86 is one way to check (without loading it); I think you can use `require` or something along those lines too, and here's how HH Hadley does it in httr: https://github.com/r-lib/httr/blob/master/R/utils.r#L34-L42

R/gadgets.R
```diff
@@ -0,0 +1,58 @@
+#' Open dataset selector
+#'
+#' This function launches a shiny gadget which allows you to navigate your Crunch
+#' projects and datasets. This is useful if you can't remember a dataset's project
+#' and also saves typing long dataset names.
+#'
+#' @inheritParams listDatasets
+#' @return A `listDatasets()` call is pasted into your RStudio session`
```
nealrichardson: It returns a `loadDatasets()` call

R/gadgets.R
```diff
@@ -0,0 +1,58 @@
...
+#' @export
+listDatasetGadget <- function(kind=c("active", "all", "archived"),
+    refresh = FALSE){
+    projects <- c("Personal Project", names(projects()))
+    personal_datasets <- listDatasets(kind = kind, refresh = refresh)
```
nealrichardson: You perhaps should also confirm here that the three Suggests packages this function requires are installed.

R/gadgets.R
```diff
@@ -0,0 +1,58 @@
+#' Open dataset selector
+#'
+#' This function launches a shiny gadget which allows you to navigate your Crunch
+#' projects and datasets. This is useful if you can't remember a dataset's project
+#' and also saves typing long dataset names.
+#'
+#' @inheritParams listDatasets
+#' @return A `listDatasets()` call is pasted into your RStudio session`
+#' @export
```
nealrichardson: Is it necessary to export this function? Seems that the preferred way to get to it is through `listDatasets()` and not calling it directly.

R/gadgets.R
```diff
@@ -0,0 +1,58 @@
...
+            assignment <- ifelse(nchar(input$ds_name) > 0,
+                paste0(input$ds_name, " <- "),
+                "")
+            if (input$project == "Personal Project") {
+                code <- paste0(assignment, "loadDataset('", input$dataset,"')")
+            } else {
+                code <- paste0(assignment, "loadDataset('",
+                    input$dataset,
+                    "', project = '",
+                    input$project, "')" )
```
nealrichardson: This will result in invalid syntax if either dataset or project name has an apostrophe in it. 

Consider factoring out this syntax construction to a function that you can unit test.

GShotwell: Replying here to the general review comment about testing shiny apps. 

It seems like the best method for testing shiny apps in general is the shinytest package (https://github.com/rstudio/shinytest) which allows you to record actions taken in a shiny session. Unfortunately this currently doesn't support shiny apps which are launched from a project function which makes this not that useful for this case. 

What I've done is pulled as much logic as I can out of the shiny function and tested it using our existing approach. I considered doing something like mocking the input list and comparing the output of the server function to a baseline, but the shiny environment which is returned by the render functions is a bit complicated to make that useful. Moreover it might cause spurious test failures if the shiny library changes. 

So there are two basic ways to test the shiny logic: 
1) Extend `shinytest` to accommodate apps launched from functions
2) Pull the server and ui elements into a folder based app and test that separately. 

In this case it might not be worth setting that up as the shiny logic is pretty straightforward, but I'm happy to look into it further if we think it's a priority.

R/get-datasets.R
```diff
@@ -44,18 +48,28 @@ datasets <- function (x=getAPIRoot()) {
...
-                          refresh=FALSE) {
-    dscat <- selectDatasetCatalog(kind, project, refresh)
-    return(names(dscat))
+    refresh=FALSE, shiny = FALSE) {
+    if (shiny) {
+        if (all(c("rstudioapi", "shiny", "miniUI") %in% rownames(installed.packages())) &&
+                rstudioapi::isAvailable("0.99.878")) {
+            listDatasetGadget(kind, refresh)
+        } else {
+            halt("Dataset viewer can only be accessed from RStudio versions 0.99.878 or later.")
```
jonkeane: We should have some reference to the packages if that's what's causing the error (I would split this into two blocks, one that checks if the packages are there, and the other that checks RStudio version. That way the error message can be tailored to the package or the version problem.

tests/testthat/test-misc.R
```diff
@@ -154,6 +154,12 @@ test_that("default date formater", {
     expect_equal(datetimeFormater(NULL), "%Y-%m-%d %H:%M:%S")
 })
 
+test_that("checkInstalledPackages", {
+    expect_error(checkInstalledPackages(c("not", "installed")),
+        paste0("Missing required packages: ", dQuote("not"), " and ", dQuote("installed")))
+    expect_silent(checkInstalledPackages("stats"))
```
jonkeane: v minor: remove the `expect_silent` here, please

jonkeane:
This is looking good. I know there weren't super great methods for testing shiny packages right now, but I wonder if we could get some coverage by splitting out and then testing the shiny server and UI objects like you had mentioned doing. We don't have to test all aspects of each object, but it would be good to have confidence that they can execute without error.

nealrichardson:
As an example, https://github.com/Crunch-io/crunchy/tree/master/tests/testthat has some bare tests that do little more than assert that the code has the right number of open and close braces. But it's at least something.

R/misc.R
```diff
@@ -266,3 +266,20 @@ has.function <- function (query, funcs) {
 
     return(FALSE)
 }
+
+escapeQuotes <- function(str) {
```
nealrichardson: Why does this escape both single and double quotes? Seems like you only need to escape whatever you're using as the string quote in `buildLoadDatasetCall` (which appears to be `'`.

R/gadgets.R
```diff
@@ -0,0 +1,61 @@
...
+                ds_name = input$ds_name)
+            shiny::stopApp(returnValue = rstudioapi::insertText(text = code))
+        })
+    }
+    shiny::runGadget(ui, server)
+}
+
+buildLoadDatasetCall <- function(project, dataset, ds_name = "") {
+    dataset <- escapeQuotes(dataset)
+    assignment <- ifelse(nchar(ds_name) > 0,
```
nealrichardson: Maybe it's just how GitHub is rendering this, but the indentation looks misaligned in this function.

R/gadgets.R
```diff
@@ -0,0 +1,61 @@
+#' Open dataset selector
+#'
+#' @inheritParams listDatasets
+#' @return A `loadDataset()` call is pasted into your RStudio session`
+listDatasetGadget <- function(kind=c("active", "all", "archived"),
```
nealrichardson: Space between `function` and `(` please (ditto for all new functions in this PR).

R/get-datasets.R
```diff
@@ -44,18 +48,28 @@ datasets <- function (x=getAPIRoot()) {
...
 #' input for [loadDataset()]
 #' @export
 listDatasets <- function (kind=c("active", "all", "archived"), project=NULL,
-                          refresh=FALSE) {
-    dscat <- selectDatasetCatalog(kind, project, refresh)
-    return(names(dscat))
+    refresh=FALSE, shiny = FALSE) {
+    if (shiny) {
+        checkInstalledPackages(c("rstudioapi", "shiny", "miniUI"))
+        if (rstudioapi::isAvailable("0.99.878")) {
```
nealrichardson: If you use `rstudioapi::verifyAvailable`, you don't need this `if`: https://github.com/rstudio/rstudioapi/blob/55d3200e6c8b1627b204215341501ca6e4e03212/R/code.R#L23

GShotwell:
For the testing issue, I think in this case we should leave the gadget untested rather than try to pull the code out to test somewhere else. The reasons for this are: 
- It's not a core part of the package in the sense that if there's a bug with the gadget it won't prevent the user from accomplishing something (they can just call `listDatasets()` without the shiny argument)
- It's not something that can be called programatically, so a bug won't break any automated processes
- The gadget is fairly well isolated so it's unlikely that changes to other crunch functions or objects will break it (assuming that `projects` and `listDatasets` tests are passed)
- The `shiny` features called by the gadget are limited and fairly stable. 

It also seems like bugs which come up with the gadget are going to be fairly easy to replicate and fix because there's not that much logic in the app. So my feeling is that the expected value of avoiding gadget bugs is lower than the cost of setting up and maintaining automated testing. This is probably worth revisiting if we end up adding more complex or important shiny products in the future.

PR #208, opened 2018-02-23
Better progress handling

nealrichardson:
* Option to silence progress reporting (not just in tests)
* Export `pollProgress` and use that in the progress timeout error message


R/new-dataset.R
```diff
@@ -89,7 +89,9 @@ createDataset <- function (name, body, ...) {
 #' @export
 prepareDataForCrunch <- function (data, ...) {
     ## Get all the things
-    message("Processing the data")
+    if (isTRUE(getOption("crunch.show.progress", TRUE))) {
+        message("Processing the data")
+    }
```
jonkeane: It might be good to have a separate `progressMessage` function that has the check for show.progress in it.

R/progress.R
```diff
@@ -1,3 +1,16 @@
...
+#' @param progress_url A Crunch progress URL
+#' @param wait Number of seconds to wait between polling. This time is increased
+#' 20 percent on each poll.
+#' @return The percent completed of the progress. Assuming the
+#' `options(crunch.timeout)` (default: 15 minutes) hasn't been reached, this
+#' will be 100. If the timeout is reached, it will be the last reported progress
+#' value.
+#' @export
 #' @importFrom httpcache uncached
 pollProgress <- function (progress_url, wait=.5) {
```
jonkeane: I wonder if it would make sense to have a `timeout` option here that could override the `crunch.timeout` option? Might be useful if there are some tasks that we want to allow to go for longer (or shorter...) without having to futz with temp options. Specifically, I'm thinking for things like csv exports where we might want to stay polling for a while (though 15minutes ought to be enough).

R/new-dataset.R
nealrichardson: Maybe so, though if we're really thinking about it, I'm skeptical that these messages are really helpful anyway. /shrug

Factoring out a `progressMessage` does help with the coverage game, so I'll invest the 90 seconds on it :)

R/progress.R
nealrichardson: I'm skeptical and don't want to pull that in to this PR.

PR #209, opened 2018-02-27
webApp methods for variables

GShotwell:
This changes our existing `webApp` function to use the cross platform `browseURL`, and adds `webApp` methods for variables.


R/variable.R
```diff
@@ -205,6 +205,17 @@ unbind <- function (x) {
     invisible(out)
 }
 
+#' @rdname webApp
+#' @export
+#' @importFrom utils browseURL
+setMethod("webApp", "CrunchVariable", function (x) browseURL(APIToWebURL(x)))
+
+setMethod("APIToWebURL", "CrunchVariable", function (x) {
+    url <- gsub("(app.crunch.io\\/api\\/datasets)", "app.crunch.io/dataset", self(x))
```
nealrichardson: 1) I'd just match on `/api/datasets/` instead of including `app.crunch.io` in it since the domain won't always be `app` (e.g. profiles, but think also alpha, test01, etc.).
2) Seems like this part should be the same as what the `CrunchDataset` method is. I mean, logically it must be, given the frontend URL scheme. So it would make sense if it used the same code.

R/variable.R
```diff
@@ -205,6 +205,17 @@ unbind <- function (x) {
...
+    url <- gsub("\\/variables\\/", "\\/browse?variableId=", url)
```
nealrichardson: If you change this to `gsub("^(.*)/variables/(.*?)/.*$", "\\1/browse?variableId=\\2", url)`, you don't need the next line, and you get some insurance in case the variable entity grows query parameters or something. 

Alternatively, you could also construct the URL rather than use regular expressions. The variable link here is the dataset link + `browse?variableId=` + `id(x)`. 

Side note: you don't need to `\\` escape `/` in your regexp, and you don't need to escape anything in the replacement text since it isn't a regexp.

tests/testthat/test-variables.R
```diff
@@ -105,6 +105,11 @@ with_mock_crunch({
         expect_error(digits(ds$birthyr) <- "0.7", "digit specifications should be an integer")
         expect_error(digits(ds$gender) <- 0, "digit specifications can only be set for numeric variables")
     })
+
+    test_that("APIToURL generates correct URL", {
```
nealrichardson: Passing observation: maybe we have enough of these web-app functions now that they should move to their own test file.

R/variable.R
```diff
@@ -23,6 +23,11 @@ setMethod("name<-", "CrunchVariable",
     function (x, value) setTupleSlot(x, "name", validateNewName(value)))
 #' @rdname describe
 #' @export
+setMethod("id", "CrunchVariable", function (x) {
+    return(x@tuple@body$id)
+})
```
jonkeane: I'm glad we have this now! I have totally called it expecting it to work and then been reminded "oh right I should go add that...."

tests/testthat/test-variables.R
```diff
@@ -130,7 +131,6 @@ with_test_authentication({
     test_that("before modifying", {
         expect_identical(name(ds$v1), "v1")
         expect_identical(description(ds$v2), "")
-        expect_identical(alias(ds$v1), "v1")
```
jonkeane: Was removing this an accident? I'm having trouble figuring out if it's needed or not?

tests/testthat/test-webApp.R
```diff
@@ -0,0 +1,25 @@
...
+
+    test_that("Dataset URLs", {
+        with(temp.options(crunch.api="https://fake.crunch.io/api/v2/"), {
+            expect_identical(APIToWebURL(ds),
+                "https://fake.crunch.io/dataset/511a7c49778030653aab5963")
+        })
+        expect_identical(webToAPIURL("https://app.crunch.io/dataset/b6c2325a8de9438ebab5d9a42d376b90/browse/eyJhcHBTdGF0ZVN0b3JlIjp0cnVlLCJhbmFseXplIjp7fSwidmFyaWFibGVzTmF2aWdhdG9yIjp7Iml0ZW0iOiIvZWU2NTI0YWFjMzFiNDkyZjk4M2ZiYzM0MGJjODYzYzkvIn19"),
+            "https://app.crunch.io/api/datasets/b6c2325a8de9438ebab5d9a42d376b90/")
+        expect_identical(webToAPIURL("https://app.crunch.io/dataset/3f57d1924a914176b24969bc6cc9059d?variableId=000194"),
+            "https://app.crunch.io/api/datasets/3f57d1924a914176b24969bc6cc9059d/")
```
jonkeane: more style than anything else: do we want to sanitize these ids? They are pretty long, and we could probably get away with shorter ones (the dataset ID identifier doesn't rely on the length of the ID, it doesn't appear). I'm also fine leaving them in since they do represent real world cases, and those ids don't give away anything...

tests/testthat/test-variables.R
GShotwell: Was indeed an accident, sorry!

tests/testthat/test-webApp.R
GShotwell: These tests were just copied over from their initial location to have all the webApp functions tested in one file. I can't really comment on whether we should modify them, but my intuition is that if they're already in the package we shouldn't change them unless there's a compelling reason.

tests/testthat/test-variables.R
GShotwell: Weirdly I don't see this line as being removed in my local repo.... `git confused`

R/variable.R
nealrichardson: The way to define this method that is consistent with the rest above and below it is `tuple(x)$id`.

R/dataset.R
```diff
@@ -428,27 +425,24 @@ webToAPIURL <- function (url) {
...
+#' @name webApp
+#' @importFrom utils browseURL
 #' @export
-webApp <- function (dataset) {
-    if (.Platform$OS.type == "unix") {
-        cmd <- ifelse(grepl("apple", R.version$platform), "open", "xdg-open")
-        url <- APIToWebURL(dataset)
-        system_call(cmd, url)
+webApp <- function(x) {
+    url <- try(APIToWebURL(x), silent = TRUE)
```
nealrichardson: Rather than try/catching here, you could make `APIToWebURL` fail with a nice message. Define a `APIToWebURL` method for `"ANY"` that does `halt("Web URL not available for objects of class ", class(x))`. cf. https://github.com/Crunch-io/rcrunch/blob/master/R/shoji.R#L69 or https://github.com/Crunch-io/rcrunch/blob/master/R/AllGenerics.R#L241-L244.

PR #210, opened 2018-03-01
Changed default behaviour for after

GShotwell:
Small change to set default `after` value for subtotal if the position is relative and after is not supplied. If `after` is not supplied it defaults to the category specified in the `categories` argument to `Subtotal`. Last in this case refers to the order of the variable's categories, not the order that those categories are supplied to `Subtotal`.

### DIFF
--- a/R/subtotals-and-headings.R
+++ b/R/subtotals-and-headings.R
@@ -215,7 +215,8 @@ setMethod("subtotals<-", c("CrunchVariable", "ANY"), function (x, value) {
             } else {
                 var_cats <- names(categories(x))
             }
-            ordered_cats <- a$categories[order(match( a$categories, var_cats))]
+            sub_cats <- a$categories[a$categories %in% var_cats]
+            ordered_cats <- sub_cats[order(match(sub_cats, var_cats))]
             a$after <- rev(ordered_cats)[1]
             return(a)
         } else {
--- a/R/subtotals-and-headings.R
+++ b/R/subtotals-and-headings.R
@@ -42,7 +42,8 @@
 #' the position of the subtotal or heading, either at the top, bottom, or
 #' relative to another category in the cube (default).
 #' @param after character or numeric if `position` is "relative", then the
-#' category name or id to position the subtotal or heading after
+#' category name or id to position the subtotal or heading after. If not supplied
+#' this defaults to the last of the `categories` supplied to `Subtotal`.
 #'
 #' @examples
 #' \dontrun{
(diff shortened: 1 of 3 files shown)


R/subtotals-and-headings.R
```diff
@@ -125,18 +126,9 @@ Subtotal <- function (name,
...
-    } else {
-        # if position is realtive, make sure there is not an after
-        if (!is.null(after)) {
-            halt("If position is not relative, you cannot supply a category id",
-                 " or name to the ", dQuote("after"), " argument")
-        }
+    if (position != "relative" && !is.null(after)) {
+        halt("If position is not relative, you cannot supply a category id",
+            " or name to the ", dQuote("after"), " argument")
     }
```
jonkeane: It feels so good to get rid of this more complex if/then statement.

R/subtotals-and-headings.R
```diff
@@ -214,6 +206,22 @@ setMethod("subtotals<-", c("CrunchVariable", "ANY"), function (x, value) {
...
+    # If position is relative we default to setting after to be the last category
+    # in the Subtotal category.
+    value <- lapply(value, function(a){
+        if (is.null(a$after) && a$position == "relative") {
+            if (is.numeric(a$categories)) {
+                var_cats <- ids(categories(x))
+            } else {
+                var_cats <- names(categories(x))
+            }
+            ordered_cats <- a$categories[order(match( a$categories, var_cats))]
```
jonkeane: Super minor: there's an extra space after `match(`

jonkeane: What happens if the category specified in the subtotal is not in the categories in `var_cats`? That shouldn't be a common condition, but I could imagine it happening if someone is applying the same `Subtotal` sets to different variables that have mostly the same, but some different categories.

Also, what happens if the last category happens to be a missing category? I don't see a test for that case, but it would be a good one to cover.

jonkeane: Actually, it looks like having a missing category in the subtotal _is_ covered by the test (line 129), but having multiple non-missing isn't. Maybe add a test where c(1, 2) are given as categories as well (and a comment on the first test that we're testing what happens if there is a missing category would be great!)

GShotwell: Great, I added those tests and comments. It looks like I also added a quick check that the subtotal categories match the variable categories.

jonkeane:
I looked through this again, and I ran it locally. It looks good to me. I tried a couple of things to 🦎 it (e.g. supplying names of categories) and everything worked fine.

PR #212, opened 2018-03-02
New import vignette

nealrichardson:
* Rewrites intro vignette to show how to import and do basic data cleaning
* Uses `httptest` mock loading in the vignette so it's not cooking-show faked
* Adds SO dataset (replaces https://github.com/Crunch-io/rcrunch/pull/185) 
* Adds some structure to the "articles" ToC in pkgdown

(Currently still WIP)


vignettes/crunch.Rmd
```diff
@@ -0,0 +1,229 @@
...
+counts <- counts[rev(order(counts))]
+subvariables(ds$WantWorkedLanguageMR) <- subvariables(ds$WantWorkedLanguageMR)[names(counts)]
+```
+```{r echo = FALSE}
+knitr::include_graphics("images/WantWorkMRReodered.png")
+```
+
+## Archiving and deleting datasets
+
+Datasets can also be deleted permanently. This action cannot be undone, so it should not be done lightly. `crunch` provides two ways to delete a dataset: a `delete()` method on a dataset object, like
```
jonkeane: This transition from working with variables, etc. to "Now you can delete datasets!" felt pretty abrupt. Maybe insert a section that talks about how we don't expect you to delete things, especially inside of a dataset, it's better to hide a variable etc. then say "But, if you really absolutely want to, you _can_ delete variables..." and then this would feel a little less sudden.

vignettes/crunch.Rmd
```diff
@@ -0,0 +1,229 @@
...
+```
+
+The second way to delete is `deleteDataset()`: you supply a dataset name. This way is faster if you have not already loaded the dataset object into your R session: no need to fetch something from the server just to then tell the server to delete it.
+
+```{r, include=FALSE}
+logout()
+end_vignette()
+```
+
+<!-- * [Datasets](datasets.html): creating, loading, and manipulating datasets in Crunch
```
jonkeane: I think it would be good to have some sort of closing and keep some of this table of contents so that people know where to go next.

vignettes/datasets.Rmd
```diff
@@ -1,59 +1,40 @@
...
+## listDatasets (including projects)
 
-```{r, eval=FALSE}
-load("../vignettes/economist.RData")
-dim(economist)
-```
-```{r, echo=FALSE}
-dim.ds
-```
+## searchDatasets
```
jonkeane: Is this section not yet filled in?

vignettes/crunch.Rmd
```diff
@@ -0,0 +1,229 @@
...
+ds[, "CompanySize"]
+```
+
+For the most part, the **crunch** package will bring the minimum amount of data down from the server. Most of the methods work by calculating values on the server, and then transmitting the results to you R session instead of transmitting all the data. In this case the server calculates and transmits the variable summary instead of sending the data. If you ever want to actually access the data in the variable you can call `as.vector(var)`.
+
+Crunch also allows you to store additional metadata in the variables. For instance we can add the survey questions from the schema to the variable description like this:
+
+```{r state change3, echo=FALSE}
+change_state()
+```
```
jonkeane: This is not a change that should happen in this PR, but it would be super clever if you could make change_state happen automatically when a specific option is passed to a markdown chunk (or, even better, if it could detect that a change had happened and trigger itself right before...)

vignettes/crunch.Rmd
nealrichardson: I believe this is left over from an old vignette, should be moved to another or just deleted.

PR #218, opened 2018-03-15
Source crunch-test to disposable environment

GShotwell:
Small change to make `devtools::test()` work from the console without having to source `inst/crunch_test.R`

### DIFF
--- a/inst/crunch-test.R
+++ b/inst/crunch-test.R
@@ -1,4 +1,5 @@
 library(httptest)
+test_env <- environment()
 
 # use test.api or R_TEST_API if it's available, if not use local
 if (!is.null(crunch::envOrOption("test.api"))) {
@@ -81,7 +82,7 @@ with_DELETE <- function (resp, expr) {
 
 silencer <- temp.option(show.error.messages=FALSE)
 
-assign("entities.created", c(), envir=globalenv())
+assign("entities.created", c(), envir=test_env)
 with_test_authentication <- function (expr) {
     if (run.integration.tests) {
         env <- parent.frame()
@@ -95,10 +96,10 @@ with_test_authentication <- function (expr) {
         ## Any time an object is created (201 Location responts), store that URL
         tracer <- quote({
             if (!is.null(loc)) {
-                seen <- get("entities.created", envir=globalenv())
+                seen <- get("entities.created", envir=test_env)
                 assign("entities.created",
                     c(seen, loc),
-                    envir=globalenv())
+                    envir=test_env)
             }
         })
         with_trace("locationHeader", exit=tracer, where=crGET, expr={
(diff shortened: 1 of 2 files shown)


nealrichardson:
It seems like the question is: am I calling `devtools::load_all()` myself, or is it because I'm running tests? If running tests, we always want to source crunch-test.R. If just loading the package, we don't want to clutter the workspace with test objects. So try sniffing `sys.calls()` to see if the top-level calling function is `load_all`.

jonkeane:
I've updated this to reflect the new `setup` and `teardown` files that were recently added to testthat. Both of those are only run when tests actually run. This will improve testing and development flow of crunch and other packages because:

1. It will allow `devtools::load_all()` to ignore the sourcing of `crunch-test.R`
2. Even if tests are run interactively, the tests are more respectful of user-environments and no longer change options or create new objects (with the exception of `entities-created`)

jonkeane:
I'll add the documentation to more formally describe how dependencies ought to use these files / what to put where. I'm thinking in a new section of `README` but could make a separate vignette / add it to the internals vignette if we think the `README` is already too cluttered.

As for functions that would be nice to have in `helper.R` two that I find myself manually sourcing a lot are `loadCube` and `cubify` which currently reside in `crunch-test.R`, but could just as easily go in `helper.R` and that would mean they would be available when writing tests (which I find invaluable). They could also go in the package as internal functions if we want to make it easy for other packages to use them in testing.

nealrichardson:
Right, those functions are in crunch-test.R. (And if you're interactively writing tests, you should probably source crunch-test.R.) My question was whether there was anything currently in helper.R that had any business not being moved to setup.R. Since historically, helper.R _was_ the test setup, everything I put in there was setup for tests.

jonkeane:
Yeah, you're right. I've moved the options and seed setting from helper.r to setup.R since those really should only be called when the tests are run programmatically / as part of testthat running the tests. Though some of the functions like `newDatasetFromFixture` should remain in helper.R

I've looked through a bunch of the discussion on devtools, pkgload, and testthat's GitHub Issues/PRs about helpers, setup, and their differences. And what I've been able to put together about the philosophy behind them is the following (though even in the testthat documentation, this isn't particularly clearly spelled out):
* helpers.R should contain helpers which should be accessible when writing tests because they make writing tests interactively easier. Which is why it's loaded into the package namespace when `load_all` is called. The philosophy that I've seen come up again and again in the discussions about this is that it makes it easy to iterate your code and iterate test writing. I know this isn't the way that everyone develops, but it is a way that I have found useful and appears to be the philosophy of the developers of testthat as well. If we can support both ways with the same setup, I would appreciate that.
* The new setup.R / teardown.R are for things that need to alter the testing environment when run programmatically but would be impolite or problematic to change in the current session if the developer/user runs `load_all` to check and see if the changes just made work interactively.

I would propose the split be that test utilities and helper functions be in helper.R (or be in crunch-test.R and that be loaded into the correct namespace e.g. testthat's philosophy and behavior is that this be the package namespace) and that functions / objects that track what has been created as well as changes to options be put in setup.R (with the appropriate tearing down in teardown.R). I didn't originally here because I was trying to have as light a touch as possible on the current setup — I'm happy to try and align the setup we have here even more closely with the testthat philosophy (without breaking current usage by other packages) if we want to go that direction (I certainly would appreciate it!).

PR #220, opened 2018-03-19
Cube extract methods

GShotwell:
Creates `[` method for Crunch Cubes.


R/cube-result.R
```diff
@@ -13,6 +13,130 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+
+#' Replace missing elements with TRUE
+#'
+#' When subsetting missingness stands in for selecting all of the elements of that
+#' dimension. This gets tricky when you are selecting n dimensions and so need to
+#' capture the dimensions with `...`.  This function checks if an element of a list
+#' is missing and replaces that element with `TRUE` in order to handle this case.
+#'
+#' @param l a list
+#' @return a list
```
jonkeane: Should we add `@keywords internal` here?

R/cube-result.R
```diff
@@ -13,6 +13,130 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+    dim$any.or.none <- dim$any.or.none[idx]
+    dim$missing <- dim$missing[idx]
+    dim$name <- dim$name[idx]
+    return(dim)
+}
+
+#' Translate user facing cube subset to programmatic cube subset
+#'
+#' Cubes that include multiple response variables create a special kind of complexity.
+#' Multiple response variables are actually 2d arrays with the selections along one dimension (`cat`, `dog`, `fish``)
```
jonkeane: The terminology for the dimension with `cat`, `dog`, `fish` is not super standard, but Mike and I have been trying to standardise on "items", but we should definitely not call them selections (since that's generally what the _other_ dimension is referred to as)

Additionally, there's an extra \` after `fish`

R/cube-result.R
```diff
@@ -13,6 +13,130 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+#' and the selection status along the second dimension (`selected`, `not_selected`).
+#' When an MR variable is crossed with a categorical variable it creates a 3d array
+#' with the categorical variable's categories along one dimension and the MR dimensions
+#' on the other two.
+#'
+#' The complexity is that while the real MR cube includes two dimensions per MR,
+#' we only show the user one dimension per MR which represents only the `selected` cases.
```
jonkeane: Here `cases` should be changed to `items`

R/cube-result.R
nealrichardson: "responses" is the other word we use, which calls back to variable type name.

R/cube-result.R
```diff
@@ -13,6 +13,130 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+#' @rdname cube-methods
+#' @export
+
+setMethod("[", "CrunchCube", function (x, i, j, ..., drop = TRUE) {
+    subset <- eval(substitute(alist(i, j, ...)))
+    subset <- replaceMissingWithTRUE(subset)
+    translated_subset <- translateCubeIndex(x, subset, drop)
+
+    out <- x
+    out@arrays$count <- subsetByList(out@arrays$count, translated_subset, drop)
```
nealrichardson: `lapply` over `out@arrays`, there may be many measures (e.g. via `crplyr::summarize()`)

R/cube-result.R
```diff
@@ -13,6 +13,130 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+    out[match(user_names, prog_names)] <- subset
+    vars <- as.data.frame(variables(x))
+    mr_vars <- vars$alias[vars$type == "subvariable_items"]
+    if (drop) {
+        for (i in seq_along(prog_names)) {
+            if (i == 1) {
+                next
+            }
+            if (prog_names[i] %in% mr_vars &&             # Check if MR variable
+                    prog_names[i] == prog_names[i - 1] && # Check if MR selection variable
```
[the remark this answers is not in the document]
GShotwell: I discovered `is.selectedDimension` which I can use to handle these cases.

R/cube-result.R
```diff
@@ -13,9 +13,249 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+#' dimension. This gets tricky when you are selecting n dimensions and so need to
+#' capture the dimensions with `...`.  This function checks if an element of a list
+#' is missing and replaces that element with `TRUE` in order to handle this case.
+#'
+#' @param l a list
+#' @return a list
+#' @keywords internal
+replaceMissingWithTRUE <- function(l){
+    out <- lapply(l, function(x){
+        if (is.symbol(x)) {
```
GShotwell: This works, and I think is safe, but I'd still prefer something cleaner for this test. There is a great function in `rlang` which does just the trick, if we want to include that dependency. https://www.rdocumentation.org/packages/rlang/versions/0.1.6/topics/missing_arg

R/cube-result.R
```diff
@@ -13,9 +13,249 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+        halt("Invalid subset:\n",
+            paste0(errs[1:min(sum(invalid_subsets), 5)], collapse = "\n"))
+    }
+
+    translated_subset <- translateCubeIndex(x, subset, drop)
+    if (x@useNA == "no") {
+        translated_subset <- skipMissingCategories(x, translated_subset)
+    }
+
+    out <- x
```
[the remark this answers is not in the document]
GShotwell: I think I'll keep this as `out`, or else I'm going to start getting confused :)

tests/testthat/test-cube-subset.R
```diff
@@ -0,0 +1,232 @@
...
+                    missing = FALSE,
+                    id = 1L, name = "cats"),
+                list(numeric_value = NULL,
+                    missing = TRUE,
+                    id = -1L,
+                    name = "No Data"),
+                list(numeric_value = 2L,
+                    missing = FALSE,
+                    id = 2L,
+                    name = "dogs")
```
nealrichardson: The dogs category should probably be subset off too.

tests/testthat/test-cube-transforms.R
```diff
@@ -325,7 +325,7 @@ with_test_authentication({
               sep = "\n"),
               fixed = TRUE)
         expect_is(trans_cube, "array")
-        expect_equal(dim(pets_cube), 6)
+        expect_equal(dim((showMissing(pets_cube))), 6)
```
nealrichardson: extra parens

R/cube-result.R
```diff
@@ -13,9 +13,248 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+#' @rdname cube-missingness
+#' @export
+setMethod("hideMissing", "CrunchCube", function(cube) setCubeNA(cube, "no"))
+
+#' @rdname cube-missingness
+#' @export
+setMethod("showIfAny", "CrunchCube", function(cube) setCubeNA(cube, "ifany"))
+
+setCubeNA <- function(cube, value = c("always", "no", "ifany")){
+    match.arg(value)
```
nealrichardson: This should be `value <- match.arg(value)`, pedantically, though it doesn't matter in this case since you control all functions that call this one.

R/cube-result.R
```diff
@@ -13,9 +13,248 @@ setMethod("initialize", "CrunchCube", function (.Object, ...) {
...
+    }
+
+    translated_index <- translateCubeIndex(x, index, drop)
+
+    if (x@useNA != "always") {
+        translated_index <- skipMissingCategories(x, translated_index)
+    }
+
+    out <- x
+    out@arrays[] <- lapply(out@arrays, function(arr){
```
nealrichardson: Why is this not using `subsetCubeArray()`?

R/cube-subset.R
```diff
@@ -0,0 +1,235 @@
...
+#' status along the second dimension (`selected`, `not_selected`). When an MR
+#' variable is crossed with a categorical variable it creates a 3d array with
+#' the categorical variable's categories along one dimension and the MR
+#' dimensions on the other two.
+#'
+#' The complexity is that while the real MR cube includes two dimensions per MR,
+#' we only show the user one dimension per MR which represents only the
+#' `selected` items. This means that every cube has two different
+#' representations, the low dimensional user cube, and the higher dimensional
+#' programmatic cube. This function translates user cube subsets into the higher
```
jonkeane: `programmatic` here should probably be `real`

R/cube-subset.R
```diff
@@ -0,0 +1,235 @@
...
+    return(out)
+}
+
+
+#' Handle missing categories in CrunchCube
+#'
+#' By default we don't display missing categories. The result is that when the
+#' user subsets a cube with a missing category, we need to translate that subset
+#' to only refer to the non-missing categories. This only occurs if
+#' `cube@useMissing` is set to `"no"`. This function handles this behavior by
```
jonkeane: You should add to the sentence that starts "this only occurs if..." that it also could occur if `useNA = "ifAny"` and there aren't any non-0 missing categories.

tests/testthat/test-cube-subset.R
```diff
@@ -0,0 +1,297 @@
...
+    expect_identical(subsetArrayDimension(catarray_x_mr@dims[[2]], c(2:3)), expected)
+})
+
+test_that("translateCubeIndex", {
+    expect_identical(translateCubeIndex(cat_x_mr_x_mr, alist(1, ,), drop = FALSE),
+        alist(1, , TRUE, , TRUE))
+    expect_identical(translateCubeIndex(cat_x_mr_x_mr, alist(1:2, 1:2, 1:2), drop = FALSE),
+        alist(1:2, 1:2, TRUE, 1:2, TRUE))
+    expect_identical(translateCubeIndex(cat_x_mr_x_mr, alist(1:2, 1:2, 1:2), drop = FALSE),
+        alist(1:2, 1:2, TRUE, 1:2, TRUE))
```
jonkeane: I might be missing something subtle here, but are line 112-113 the same as 114-115?

tests/testthat/test-cube-subset.R
```diff
@@ -0,0 +1,297 @@
...
+    cube <- cat_x_mr_x_mr
+    expect_equal(length(dim(cube)), 3)
+    expect_error(cube[1,2,3,4,5,6],
+                 paste0("You must supply 3 dimensions to subset a 3 ",
+                        "dimensional cube; you supplied 6."))
+    err <- paste0(c("Invalid subset:",
+        "- At position 1 you tried to select element 3 when the dimension has 2 elements.",
+        "- At position 2 you tried to select element 4 when the dimension has 3 elements."),
+        collapse = "\n")
+    expect_error(cube[3, 4,1], err)
```
jonkeane: Suuuper minor: `4,1` should be `4, 1` here

tests/testthat/test-multitables.R
```diff
@@ -245,8 +246,9 @@ with_mock_crunch({
             test_that("tab book print methods", {
                 ## TODO: print method for TabBookResult
             })
+
             test_that("The first result in a MultitableResult has 3 dimensions", {
-                expect_identical(dim(book[[1]][[1]]), c(5L, 1L, 2L))
+                expect_equal(dim(showMissing(book[[1]][[1]])), c(5, 1, 2))
```
jonkeane: Same question as above: any reason to not use `expect_identical` and `c(5L, 1L, 2L)`?

vignettes/crunch-internals.Rmd
```diff
@@ -37,4 +37,65 @@ datasets(proj) <- ds
...
+We resolve this problem by allowing the user to select whether they want the cube to represent the user cube or the printed cube, and make all of the behavior of those cubes consistent. The `cube@useNA` controls whether the user wanted to interact with the printed cube or the user cube and they can `showMissing()` and `hideMissing()` functions to control which cube they are interacting with 
+
+```
+> dim(cube)
+[1] 5 6 8
+> dim(showMissing(cube))
+[1] 5 6 10
+```
+
+This makes subsetting a bit easier because we can enforce the dimensions of the printed or literal cube. For instance
```
jonkeane: Shouldn't this just be printed cube here?

vignettes/crunch-internals.Rmd
```diff
@@ -37,4 +37,65 @@ datasets(proj) <- ds
...
+
+```
+cube[5, 6, 10] 
+```
+ Errors because the user supplied the wrong number of dimensions while
+
+```
+showMissing(cube)[cube[5, 3, 6, 3, 10] 
+```
```
jonkeane: I think this line would still error, shouldn't it be `showMissing(cube)[5, 6, 10]`?

vignettes/crunch-internals.Rmd
```diff
@@ -37,4 +37,65 @@ datasets(proj) <- ds
...
+Returns the correct data they were looking for. 
+
+Subsetting the printed cube
+
+There are two cases for dropping dimensions from the printed and user cubes which need special handling. 
+
+_Dropping MR variable_
+
+The two MR dimensions are always dropped together. For instance if the first dimension of a cube is an MR and the user asks for `cube[1, ]` both MR dimensions are dropped and a one dimensional cube is returned. 
+
```
jonkeane: I assume you're talking about an MR x cat cube here, correct?

tests/testthat/test-cube-subset.R
GShotwell: You miss nothing :)

tests/testthat/test-multitables.R
```diff
@@ -214,8 +214,9 @@ with_mock_crunch({
             expect_output(print(book[[1]]), get_output(out))
             ## TODO: print method for TabBookResult
         })
+
         test_that("The first result in a MultitableResult has 2 dimensions", {
-            expect_identical(dim(book[[1]][[1]]), c(2L, 1L))
+            expect_equal(dim(book[[1]][[1]]), c(2, 1))
```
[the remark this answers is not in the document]
GShotwell: Oh, for some reason I thought expect_identical failed here, but it now seems to pass so I'll switch it back.

PR #222, opened 2018-03-21
Support for translating haven::labelled* objects into Crunch categoricals

jonkeane:


nealrichardson:
Reading http://haven.tidyverse.org/reference/labelled.html and http://haven.tidyverse.org/reference/labelled_spss.html, I suspect that this isn't right. For example, `labelled` says: 

```
# Unfortunately it's not possible to make as.factor work for labelled objects
# so instead use as_factor. This works for all types of labelled vectors.
```

yet this PR uses `as.factor`. (Presumably there's a real problem and not just the `_` vs. `.` shibboleth.)

That said, I'm not exactly sure what "right" would look like, and maybe this is better than the status quo and we should go with it and iterate?

PR #223, opened 2018-03-21
Index table support

jonkeane:
Currently it is limited to only 2 dimensional cubes. It also doesn't work with a specific column or vector (you can only index to the proportion of the margin). This works and is tested on cubes including multiple response.


R/cube-index-table.R
```diff
@@ -0,0 +1,50 @@
...
+#' Index tables are percentages of percentages. They take the percentage from
+#' `prop.table()` and divide that by the proportions of the other margin.
+#'
+#' `index.table()` is only implemented for 2 dimensional cubes. If you need to
+#' calculate indexes for a higher dimension Cube, please slice the cube first.
+#'
+#' @param x A CrunchCube to calculate index table for
+#' @param margin which margin to index against (1 for rows, 2 for columns)
+#'
+#' @return an array of percentages indexed to the margin provided
```
nealrichardson: But they appear not to be percentages. Should the function return `100 * out` instead?

jonkeane: We can do that. Or change the docs to an array of proportions (then it matches the order of magnitude of `prop.table` which may or may not be expected or necessary). I'm ok doing either, so I'll add the `100*` unless there are strong objections.

nealrichardson: The only application of this that I've seen is multiplied by 100 and rounded. I'd say `*100` but not round since you can't back out the rounding if you didn't want it but you can `/ 100` if you want proportions. But I don't feel strongly and it's not like we're getting high demand for this.

nealrichardson:
Merged, please update NEWS.md.

PR #224, opened 2018-03-22
Mr shiny gadget

GShotwell:
Shiny gadget which allows the user to select array subvariables and build `makeMR` and `makeArray` calls from them. After working through a few designs I think that this addin works best as a dropdown item from `tools::addins` rather than adding a shiny argument to `makeArray` and `makeMR` the reason for this is that adding a shiny argument would require editing the users code in place, which seems both a bit tricky and also somewhat surprising to the user. 

Some highlights of app: 
_Select Subvariables_
- Select which dataset you want to use
- Search subvariables using regular expressions
- Checked variables persist regardless of what's entered into the search bar so you can re-search without losing your selected variables
- `select all` and `clear all` buttons select and clear only the variables which match the search bar

_Build Array_
- Checks whether  selected subvariables can be combined into multiple response (they have at least one category in common). 
- Provides a checklist of selection categories which are shared by all subvariables.


GShotwell:
Jon thought of a good testing solution here which is to move the shiny code from this function to `crunchy` and run the testing on that build process. This package would then suggest `crunchy` and call the gadget functions in the same way that they are called here. This way we would be able to get the tested shiny apps without having to add phantom.js to our Jenkins process.

nealrichardson:
The idea is that we'd move any/all RStudio gadgets to the `crunchy` package? Sounds reasonable to me. I checked and it is acceptable to `check` to Suggest a package that depends on your package; Travis will confirm for us whether doing so gets us into some kind of cyclical dependency hell.

GShotwell:
Closing because I think it's simpler at this point to create a new branch for the bindings to the Crunchy gadget

PR #226, opened 2018-03-26
createDataset now automatically enables variable_folders in tests. 

jonkeane:
If the option `folders.enabled` or the environment variable `R_FOLDERS_ENABLED` are `TRUE`, so will interactively creating datasets.


R/new-dataset.R
```diff
@@ -69,7 +69,15 @@ createDataset <- function (name, body, ...) {
         body <- wrapEntity(name=name, ...)
     }
     dataset_url <- crPOST(sessionURL("datasets"), body=toJSON(body))
-    invisible(loadDataset(dataset_url))
+
+    ds <- loadDataset(dataset_url)
+
+    # TODO: remove this when the future folders API is default
+    if (envOrOption("folders.enabled") %||% FALSE) {
```
nealrichardson: `envOrOption` should take a `default` arg like `getOption` (if it doesn't already) so that you don't have to `%||%`.

jonkeane: I totally agree! I actually added it in the code that was [part of](https://github.com/Crunch-io/rcrunch/pull/218/files#diff-657870636200ce01f019ca1d50e527cfR173) PR #218

R/new-dataset.R
```diff
@@ -69,7 +69,7 @@ createDataset <- function (name, body, ...) {
         body <- wrapEntity(name=name, ...)
     }
     dataset_url <- crPOST(sessionURL("datasets"), body=toJSON(body))
-    invisible(loadDataset(dataset_url))
+    return(invisible(loadDataset(dataset_url)))
```
jonkeane: 😻

PR #242, opened 2018-04-19
Add method to extract subvariables from variable tuples

### LINKED ISSUE #215: Enrich the variable catalog generated by CrunchCube objects
`variables()` on a CrunchCube return a variable catalog with variable tuples which don't have all the same methods for accessing subvariables. 

We should work on making methods like `subvariables()`, `names()`, `aliases()` etc. work on variable tuples (when that information is available, eg variable tuples form CrunchCubes had subreferences with names and aliases, however variable tuples from the variable catalog doesn't).

- [x] `names()` for subvariables/subreferneces from a CrunchCube
- [x] `aliases()` for subvariables/subreferneces from a CrunchCube

jonkeane:
Closes #215


R/subvariables.R
```diff
@@ -34,27 +34,43 @@ setMethod("subvariables", "CategoricalArrayVariable", function (x) {
...
+            # if there is a subvariable element, we need to use that for ordering
+            names(subvars) <- paste0(absoluteURL(names(subvars), base=catalog_url), "/")
+            out <- Subvariables(
+                index=subvars[x$subvariables],
+                self=catalog_url
+            )
+        } else {
+            # if there is no subvariable element, this might be from a
+            # CrunchCube which doesn't have a subvariables member, though the
+            # order *might* not be deterministic, checking zz9 source still.
```
nealrichardson: You could inject a "subvariables" member in `cubeVarReferences()` based on the dimension.

R/cube-dims.R
```diff
@@ -39,6 +39,25 @@ cubeVarReferences <- function (x) {
...
+                                     function (x) x$alias %||% "",
+                                     character(1))
+
+        if (is.null(names(tuple$subreferences))) {
+            # if there are no names for the subvariable elements, fake urls from
+            # the aliases
+            names(tuple$subreferences) <- tuple$subvariables
+        }
+
+        # add a trailling slash to match how urls will look.
```
nealrichardson: Why is this necessary?

R/subvariables.R
```diff
@@ -34,27 +34,29 @@ setMethod("subvariables", "CategoricalArrayVariable", function (x) {
...
+    return(absoluteURL(unlist(x$subvariables), base=x@index_url))
+}
+
 #' @rdname Subvariables
 #' @export
 setMethod("subvariables", "VariableTuple", function (x) {
-    ## Return subvariable *urls* from a Tuple, properly formatted and absolute
-    return(absoluteURL(unlist(x$subvariables), base=x@index_url))
+    catalog_url <- absoluteURL(x$subvariables_catalog, base=x@index_url) %||% ""
+        subvars <- x$subreferences
```
nealrichardson: Is it just GitHub, or is the indentation inconsistent here?

jonkeane: You're right, the indentation was wrong here. I've fixed it.

R/cube-dims.R
jonkeane: The subvariables from a variable entity end with a `/` since they are full URLS (https://github.com/Crunch-io/rcrunch/blob/master/inst/app.crunch.io/api/datasets/1/variables/mymrset.json#L20-L24) 

With cube responses, here, we are faking the url from just the variable id, which lacks the trailing `/` (https://github.com/Crunch-io/rcrunch/blob/master/tests/testthat/cubes/mr-by-mr-different-mrs.json#L231-L234).

This allows the subvariable method for variable tuples to be more general and not special case cubes (at this point specifically: https://github.com/Crunch-io/rcrunch/pull/242/files#diff-d2a62bc0f07894f773e0ab8ecd51becaR57)

Issue #246, opened 2018-04-27
makeMRFromText should allow categorical origin variables


GShotwell:
Currently the server rejects the variable derivation generated from `makeMRFromText` when the origin variable is categorical. I patched the function to error when you try to provide it a categorical but it would be great if categorical variables were allowed. We need to determine a couple of things before fixing this:

1) Do users actually want to derive MR variables from delimited categorical variables?
2) Should this be moved to zz9 in order to have cross-platform support, and avoid the gnarly variable definition?

PR #247, opened 2018-04-27
makeMRFromText categorical bug

GShotwell:
Raise error when user tries to derive an MR from a delimited categorical. This is just a quick fix, and I raised an issue with next steps for this area https://github.com/Crunch-io/rcrunch/issues/246

### DIFF
--- a/R/make-array.R
+++ b/R/make-array.R
@@ -111,11 +111,12 @@ makeMRFromText <- function (var,
     if (missing(name)) {
         halt("Must supply a name for the new variable")
     }
-    if (is.Categorical(var) || is.Text(var)) {
+    if (is.Text(var)) {
         uniques <- names(table(var))
     } else {
-        halt(dQuote(substitute(var)),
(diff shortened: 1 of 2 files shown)


R/make-array.R
```diff
@@ -111,11 +111,12 @@ makeMRFromText <- function (var,
     if (missing(name)) {
         halt("Must supply a name for the new variable")
     }
-    if (is.Categorical(var) || is.Text(var)) {
+    if (is.Text(var)) {
         uniques <- names(table(var))
     } else {
-        halt(dQuote(substitute(var)),
-             " must be a Categorical or Text Crunch Variable.")
+        halt(dQuote(deparse(substitute(var))),
```
jonkeane: I'm curious what the `deparse` is doing here? I was under the impression that `substitute` returned a character-enough object for `halt`/`stop` to deal with. Was there an issue without it?

jonkeane: Aaaah, makes sense. TIL!

PR #248, opened 2018-05-01
Fixed a bug with changeCategoryID

jonkeane:
`changeCategoryID()` would break if there was an exclusion set, and there was at least one instance of the category that was being changed that had been excluded.

This includes tests for that circumstance as well as a fix that removes the exclusion and then re-sets it after `changeCategoryID()` is done.


jonkeane:
Yeah, I agree that unsetting/resetting the exclusion here multiple times could be problematic. 

You're right the error is [removing the old category](https://github.com/Crunch-io/rcrunch/blob/master/R/change-category-id.R#L68). I need to think a bit more about how to implement either of those two options. The tricky part is that we've already changed some of the metadata (and data), so erroring without unwinding the changes would leave the variable in a strange state.

jonkeane:
This is ready for re-review. I haven't been able to come up with a more clever way to unwind changes and warn, another way we could go about this would be to stop if there is an exclusion set at all and ask the caller to unset it. I don't like that because it is pretty unfriendly (though it won't leave the variable in a bad state which is good).

PR #249, opened 2018-05-03
Get cube dims

GShotwell:
This PR creates a helpful function to get the type of each dimension as a character vector. We need something like this because categorical arrays and multiple response variables both create unique cubes which need special handling, and it's useful to be able to ask whether a dimension. 

Longer term I think this information should be included in the cube which comes off the server because the code used to infer the dimension type feels a little fragile.


tests/testthat/test-cube-dims.R
```diff
@@ -45,32 +63,32 @@ with_mock_crunch({
...
+            c("cat_feeling", "dog_feeling"))
 
         mr_var <- variables(catarray_x_mr)[[3]]
         expect_equal(aliases(subvariables(mr_var)),
-                     c("food_opinion#", "rest_opinion#", "play_opinion#"))
+            c("food_opinion#", "rest_opinion#", "play_opinion#"))
         expect_equal(names(subvariables(mr_var)),
-                     c("food_opinion", "rest_opinion", "play_opinion"))
+            c("food_opinion", "rest_opinion", "play_opinion"))
     })
```
jonkeane: I thought we had settled on using the [tidy style](http://style.tidyverse.org/functions.html#long-lines-1) and RStudio's default of "Vertically align arguments in auto-indent" of aligning with the opening `(` on subsequent lines, but I might be wrong about that. I'm happy to turn that off in my IDE if we want to always go to the next indent point and not align with `(` like you have here.

R/cube-dims.R
```diff
@@ -199,3 +200,46 @@ is.selectedArrayDim <- function (dim) {
 
     return(FALSE)
 }
+
+
+#' Get dimension type
+#'
+#' This function returns the specific type of each cube dimension. This is useful
+#' when cubes contain categorical array or multiple response variables because it
+#' identifies the dimensions of the cube which refer to the differ parts of
```
jonkeane: "to the differ parts" -> "to the different parts"

R/cube-dims.R
```diff
@@ -199,3 +200,46 @@ is.selectedArrayDim <- function (dim) {
...
+            next
+        }
+        if (out[i] == "subvariable_items") {
+            out[i] <- "ca_items"
+            out[i + 1] <- "ca_categories"
+        }
+    }
+    names(out) <- names(vars)
+    return(out)
+}
```
jonkeane: This is nice, I wonder if it would be possible to unify this code with the code like https://github.com/Crunch-io/rcrunch/blob/master/R/cube-dims.R#L39-L41 and https://github.com/Crunch-io/rcrunch/blob/master/R/cube-dims.R#L167-L192 . It would be good, if possible, to handle all of this inspecting and labelling in one place.

R/cube-dims.R
```diff
@@ -170,3 +170,52 @@ is.selectedDimension <- function (dims) {
...
+#' - `mr_selections`: The selection status for a multiple response variable
+#'
+#' @return A character vector. This is identical to `types()` except that
+#' the array variable types are more specific.
+#' @param x a CrunchCube or CubeDims object
+#'
+#' @return a character vector of dimension types
+#' @export
+#' @keywords internal
+getDimType <-  function (x) {
```
GShotwell: I don't have a super strong opinions about this, but is there a reason to use this implementation instead of https://github.com/Crunch-io/crplyr/pull/2/files#diff-7797b6031d801b050ee3c519dc43bea6R87 did I miss something?

jonkeane: We chatted a bit in standup, but the major reason was to unify as much of the cube dimension typing as possible under one roof.

tests/testthat/test-cube-dims.R
jonkeane: I'm deferring this change for now to make the diffs here as clean as possible. I'm happy to re-align this file before merging to what we decided.

R/cube-dims.R
```diff
@@ -144,29 +144,84 @@ setMethod("[", "CubeDims", function (x, i, ...) {
...
+            array_cats <- categories(array_cat_dim[[1]])
+            
+            # if we meet these conditions, we are actually a multiple response
+            # and should label ourself as such.
+            # Unlike the strict is.3vl, this doesn't compare cat names because
+            # they've already been munged to TRUE/FALSE
+            is.MR <- length(array_cats) == 3 &&
+                setequal(ids(array_cats), c(-1, 0, 1)) &&
+                sum(is.selected(array_cats)) == 1 &&
+                sum(is.na(array_cats)) == 1
```
aminusfu: I believe `is.MR <- sum(is.selected(array_cats)) >= 1` should be sufficient.

nealrichardson: I _think_ if we did that, we wouldn't be able to query the MR as a categorical array, ignoring the "selected" bit.

jonkeane: Yeah, I shared the same suspicion, but haven't had an opportunity to investigate if it's true. At the very least, the (possibly overly)-specific version doesn't stop us from identifying correctly MRs in place now, so I'm inclined to keep it, at least for now.

aminusfu: You would, you'd just have to have wrap it in `select_categories(mr, [])` which would strip any "selected" flags. I believe that's a rare need (correct me if I'm wrong) so it seems better to me to wrap with an extra function for that than to make every other call more complex.

nealrichardson: We can (in a separate PR) switch to using `select_categories(mr, [])`, and then in principle we could simplify this check (the computational cost of which is not a problem), but we'd also have to teach the cube logic that there could be more than one "selected" category and aggregate accordingly. Definitely out of scope here though.

R/cube-dims.R
```diff
@@ -144,29 +144,84 @@ setMethod("[", "CubeDims", function (x, i, ...) {
...
-    # determine which dimensions are selected MR dimensions
-    selecteds <- mapply(is.it, x=index(vars), dim=dims@.Data,
-                        MoreArgs=list(MRaliases=MRaliases))
-    names(selecteds) <- dims@names
-    return(selecteds)
+  
+    vars <- variables(x)
+    array_aliases <- aliases(vars)[types(vars) == "subvariable_items"]  
+    # vapply becuase lapply on a VariableCatalog yields lists (not tuples)
+    out <- unlist(vapply(vars, what_dim_is_it, character(1), array_aliases))
```
nealrichardson: Why `unlist()` if you're returning a character vector from vapply (which also seems to call into question the comment on L222: it sounds like vapply is actually what you want anyway).

R/cube-residuals.R
```diff
@@ -54,12 +54,8 @@ setMethod('rstandard', 'CrunchCube', function (model) {
              "slice to evaluate.")
     }
 
-    # determine which dimensions are multiple response, and treat those special
-    # TODO: make this specific to multiple response rather than both cat array and MR
-    types <- vapply(dimensions(model), function (dim) dim$references$type, character(1))
-
-    if (any(types == 'subvariable_items')) {
-        return(standardizedMRResiduals(model, types))
+    if (any(grepl("^mr_", getDimTypes(model)))) {
```
nealrichardson: Stylistically, you could do `any(startsWith(getDimTypes(model), "mr_"))` and it would read nicely.

R/cube-subset.R
```diff
@@ -222,14 +223,14 @@ skipMissingCategories <- function(cube, index){
...
 }
 
-subsetArrayDimension <- function(dim, idx){
+subsetArrayDimension <- function(dim, idx, dim_type){
     dim$missing <- dim$missing[idx]
     dim$name <- dim$name[idx]
 
     # subset the category or item metadata
-    if (dim$references$type == "categorical") {
+    if (dim_type %in% c("categorical", "ca_categories")) {
```
nealrichardson: But not "mr_selections"? Aren't those categories too?

R/AllGenerics.R
```diff
@@ -35,7 +35,7 @@ function (x, value) standardGeneric("startDate<-"), signature="x")
 setGeneric("endDate", function (x) standardGeneric("endDate"))
 setGeneric("endDate<-",
     function (x, value) standardGeneric("endDate<-"), signature="x")
-setGeneric("alias")
+setGeneric("alias", function (object) standardGeneric("alias"))
```
nealrichardson: The other attribute getters are `function (x)` so I'd follow that for consistency.

R/tuple.R
```diff
@@ -102,6 +104,18 @@ setMethod("delete", "DatasetTuple", function (x, ...) {
 #' @export
 setMethod("name", "ShojiTuple", vget("name"))
 
+#' @rdname tuple-methods
+#' @export
+setMethod("alias", "ShojiTuple", function(object) vget("alias")(object))
```
nealrichardson: I'm not sure this makes sense as a ShojiTuple method since only VariableTuples would have them.

R/cube-dims.R
```diff
@@ -144,29 +144,84 @@ setMethod("[", "CubeDims", function (x, i, ...) {
...
+#' - `mr_selections`: The selection status for a multiple response variable
+#'
+#' @return A character vector. This is identical to `types()` except that
+#' the array variable types are more specific.
+#' @param x a CrunchCube or CubeDims object
+#'
+#' @return a character vector of dimension types
+#' @export
+#' @keywords internal
+getDimTypes <-  function (x) {
```
nealrichardson: Why does this function need to be exported?

R/cube-dims.R
```diff
@@ -144,29 +144,84 @@ setMethod("[", "CubeDims", function (x, i, ...) {
...
+#' This function returns the specific type of each cube dimension. This is useful
+#' when cubes contain categorical array or multiple response variables because it
+#' identifies the dimensions of the cube which refer to the different parts of
+#' array variable:
+#' - `ca_items`: Categorical array items
+#' - `ca_categories`: The categories of the categorical array
+#' - `mr_items`: Multiple response options or items
+#' - `mr_selections`: The selection status for a multiple response variable
+#'
+#' @return A character vector. This is identical to `types()` except that
```
nealrichardson: "identical" seems like too strong of a word here.

nealrichardson: Also, you have two `@return` statements.

R/AllGenerics.R
jonkeane: I changed, it, but then saw [the documentation](https://crunch.io/r/crunch/reference/describe.html) that said "Same as `x` but for the alias method, in order to match the generic from another package." and didn't want to wantonly over-rule that.

I don't think there's that much benefit in keeping `crunch::alias` in line with `stats::alias`, but that was the original reason. I'll harmonize this with the rest unless there's an objection and we should maintain it.

R/cube-dims.R
jonkeane: It doesn't really. There's a similar version being used in crplyr, that we anticipated removing in deference to this one. But we could always use `crunch:::getDimTypes` there since we control both packages.

R/cube-dims.R
jonkeane: Good catch, it was an unnecessary holdover from using `lapply` there before.

R/cube-subset.R
jonkeane: They are, I didn't include them originally because I don't think they should ever show up here since we never let someone subset the MR selections out of the cube. But I've added them just in case that ever does happen, it doesn't hurt to have them there and I was probably overthinking it the first time.

R/tuple.R
jonkeane: Ah, of course. I moved these into `VariableTuple` methods for the same effect (without the downside).

R/cube-dims.R
nealrichardson: We can't `:::` if we want it on CRAN, so I guess that means we need to export.

R/AllGenerics.R
nealrichardson: Oh I see. The benefit of keeping it in line with `stats::alias` is so that we aren't masking that function (whatever it is). So let's keep it as is. I guess I'm wondering now though, why did the generic have to change at all here?

jonkeane: It looks like it didn't, for some reason I was having difficulties with it earlier passing `CMD check` but it does now without any editing, so we're good.

R/variable.R
```diff
@@ -74,6 +74,20 @@ setMethod("digits<-", "CrunchVariable", function (x, value) {
     halt("digit specifications can only be set for numeric variables")
 })
 
+
```
nealrichardson: These should probably go up with the other CrunchVariable methods, and notice with those how you can just do `x$description` etc. because of how `$` is defined for tuples.

And since you're doing that, and since the CrunchVariable methods are all like `tuple(object)$alias`, you could make those be `alias(tuple(object))` and pass through the same method.

R/cube-dims.R
```diff
@@ -144,29 +144,83 @@ setMethod("[", "CubeDims", function (x, i, ...) {
...
+            # if we meet these conditions, we are actually a multiple response
+            # and should label ourself as such.
+            # Unlike the strict is.3vl, this doesn't compare cat names because
+            # they've already been munged to TRUE/FALSE
+            is.MR <- length(array_cats) == 3 &&
+                setequal(ids(array_cats), c(-1, 0, 1)) &&
+                sum(is.selected(array_cats)) == 1 &&
+                sum(is.na(array_cats)) == 1
+            
+            if (is.MR && dim_type == "subvariable_items") {
```
nealrichardson: This feels like it should be 

```
if (is.MR) { 
    if (dim_type == ...) {
        ...
    } else if ...
} else {
    ...
}
```

but maybe that's not any clearer.

R/variable.R
jonkeane: Moved and changed the execution path to all go through the VariableTuple for `description`, `alias`, `name`, and `notes`

PR #250, opened 2018-05-03
setIndexSlot should check for duplicate URLs

nealrichardson:
See https://www.pivotaltracker.com/story/show/157283138 for an error report and https://crunch-io.slack.com/archives/C8HLAUW23/p1525372333000744 for discussion. Basically, `ds <- hideVariables(ds, c("gender", "gender")` would make an invalid request. This fix adds validation that catches that and some other possible scenarios.

### DIFF
--- a/R/shoji-catalog.R
+++ b/R/shoji-catalog.R
@@ -27,7 +27,7 @@ setIndexSlot <- function (x, i, value, unique=FALSE) {
         ## unique. Make sure that the user isn't trying to assign different
         ## values for the duplicated entries, and if not, clean it up
         if (!scalar_value && !identical(duples, duplicated(paste(urls(x), value)))) {
-            halt("Can't update an index with duplicated entries")
+            halt("Can't update the same index item with more than one entry")
         }
         x <- x[!duples]
         if (!scalar_value) {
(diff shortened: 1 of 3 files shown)


R/shoji-catalog.R
```diff
@@ -16,8 +16,28 @@ getListSlot <- function (x, i, what=character(1), ifnot=NA_character_) {
...
+        stopifnot(length(x) == length(value))
+    }
+
+    duples <- duplicated(urls(x))
+    if (any(duples)) {
+        ## This is probably a mistake, a selection from a catalog that wasn't
+        ## unique. Make sure that the user isn't trying to assign different
+        ## values for the duplicated entries, and if not, clean it up
+        if (!scalar_value && !identical(duples, duplicated(paste(urls(x), value)))) {
+            halt("Can't update an index with duplicated entries")
```
jonkeane: I'm not sure that this error will be clear what to do to solve it. What about something like "Can't update the same index item with more than one entry" I really wish we could abstract away index and entry and talk about names/aliases but this could get called in other circumstances.

nealrichardson: Yeah I kinda punted on the error message, partly because it was hard to trigger. If you can think of better words, cool. Either way, i don't expect this error to get a lot of visibility. It's more a backstop to intercept what would be an even more opaque 400 Bad Request from the server.

PR #251, opened 2018-05-03
Remove nearly all vestiges of selected_array, any/none code

nealrichardson:
There still are two tab book fixtures that have selected_array in them; we'll need to replace them. There may be other vestiges lingering too.


R/cube-result.R
```diff
@@ -151,6 +143,7 @@ takeSelectedDimensions <- function (x, dims) {
     if (any(selecteds)) {
         drops <- lapply(selecteds, function (s) {
             ## For "Selected" dimensions, we only want to return "Selected", the 1st element
+            ## TODO: this could be brittle; consider an is.selected attr/vector
```
GShotwell: Not a big deal, but it might be worth adding the ways this might be brittle to the #TODO.

PR #252, opened 2018-05-03
Nag users once per session that they should switch to folders

nealrichardson:


nealrichardson:
Yeah I'm not sure there's a reason to limit this only to interactive sessions. I think at worst it gets ignored in the script logs, and at best someone sees it there and updates their behavior.

PR #255, opened 2018-05-09
[#157298702] - Hier __hidden__ group

dianaclarke:
See:

 - https://www.pivotaltracker.com/story/show/157298702


dianaclarke:
Oh right. Thanks for the reminder. I meant to circle back to the question of ordering.

PR #259, opened 2018-05-15
Fork merge documentation

GShotwell:
Fleshes out some of our fork-merge documentation


vignettes/fork-and-merge.Rmd
```diff
@@ -0,0 +1,157 @@
...
+library(crunch)
+```
+```{r, results='hide', echo=FALSE, message=FALSE}
+library(httptest)
+start_vignette("fork-and-merge")
+```
+
+```{r message=FALSE, results='hide'}
+ds <- newDataset(SO_survey, "stackoverflow_survey")
+```
```
jonkeane: I think it would look cleaner if these blocks were merged (even if we have to do the trick of calling all of the real ones silently, but then displaying `library(crunch); login(); ds <- newDataset`

Also, we should put `login()` there, it should be obvious, but just to be complete.

vignettes/fork-and-merge.Rmd
```diff
@@ -0,0 +1,157 @@
...
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Getting Started}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+One of the main benefits of Crunch is that it lets analysts and clients work with the same datasets. Instead of emailing datasets to clients, you can update the live dataset and ensure that they will see see the most up-to-date information. The potential problem with this setup is that it can become difficult to make provisional changes to the dataset without publishing it to the client. Sometimes an analyst wants to investigate or experiment with a dataset without the risk of sending incorrect or confusing information to the end user This is why we implemented a fork-edit-merge workflow for Crunch datasets. 
+
+"Fork" originates in computer version control systems and just means to take a copy of something with the intention of making some changes to the copy, and then incorporating those changes back into the original. To see how this works lets first upload a new dataset to Crunch.
```
jonkeane: Is it helpful to have something like 'A mnemonic is a path that forks off, but then rejoins the larger path later'?

vignettes/fork-and-merge.Rmd
```diff
@@ -0,0 +1,157 @@
...
+ds <- newDataset(SO_survey, "stackoverflow_survey")
+```
+
+Imagine that this dataset is shared with several users, and you want to update it without affecting their usage. You might also want to consult with other analysts or decision makers to make sure that the data is accurate before sharing it with clients. To do this you call `forkDataset()` to create a copy of the dataset. 
+
+```{r fork dataset}
+forked_ds <- forkDataset(ds)
+```
+
+You now have an copied dataset which is identical to the original, and are free to make changes without fear of disrupting the client's experience. You can add or remove variables, delete records, or change the dataset's organization. These changes will be isolated to your fork and won't be visible to the end user until you decide to merge the fork back into the original dataset. This lets you edit the dataset with confidence because your work is isolated.
```
jonkeane: "You now have an copied" -> "You now have a copied"

vignettes/fork-and-merge.Rmd
```diff
@@ -0,0 +1,157 @@
...
+ds <- mergeFork(ds, forked_ds)
+```
+
+After merging the original dataset includes the categorical array variable which we created on the fork. 
+
+```{r check successful merge}
+ds$ImportantHiringCA
+```
+
+It's possible to to make changes to a fork which can't be easily merged into the original dataset. For instance if, while we were working on this fork someone added another variable called `ImportantHiringCA` to the original dataset the merge might fail because there's no safe to reconcile the two forks. This is called a "merge conflict" and there are a couple best practices that you can follow to avoid this problem:
```
jonkeane: "there's no safe to" -> "there's no safe way to"

vignettes/fork-and-merge.Rmd
```diff
@@ -0,0 +1,157 @@
...
+
+`ds` now has the additional rows. 
+
+```{r}
+nrow(ds)
+```
+
+## Merging datasets
+
+Merging two datasets together can often be the source of bugs or data corruption so it's a good candidate for this workflow. Let's create a fake dataset with household size to merge onto the original one.
```
jonkeane: I think we should be a bit more descriptive and change "bugs or data corruption" to "overwriting or misaligning data"

PR #262, opened 2018-05-25
Insertions improvements (mainly for subtotals)

jonkeane:
This PR cleans up and extends some of the insertions code:

1. Insertions are now calculated for any dimension, not just rows. Doing this involved creating a function `applyAgainst()` It's intended to be internal, but I added roxygen documentation to show what it does. I'm frankly surprised there didn't appear to be a simpler way to do this, so please let me know if you know of one.
2. I made the subtotal execution less likely to present the user with an error, the subtotals are calculated inside of `try`s so that the worst that happens on error is the user gets the un-inserted cube. 
3. I fixed a couple of bugs in rendering and ordering of subtotals when the cube includes an MR of some type.
4. Expanded and reorganized tests for cubes with insertions (including tests for proportions and margins)


R/cube-residuals.R
```diff
@@ -57,7 +57,7 @@ setMethod('rstandard', 'CrunchCube', function (model) {
     if (any(startsWith(getDimTypes(model), "mr_"))) {
         return(standardizedMRResiduals(model))
     } else {
-        return(chisq.test(as.array(model))$stdres)
+        return(chisq.test(as.array(noTransforms(model)))$stdres)
```
GShotwell: This nested function call is getting a bit gnarly, it might be worth breaking it up into an intermediary object.

R/cube-transforms.R
```diff
@@ -86,37 +110,187 @@ setMethod("subtotalArray", "CrunchCube", function(x, headings = FALSE) {
...
+                             transforms_list = transforms(x),
+                             dims_list = dimensions(x),
+                             useNA = x@useNA,
+                             ...) {
+    trans_indices <- which(vapply(transforms_list, Negate(is.null), logical(1)))
+    
+    # Try to calculate the transforms for any dimensiton that have them, but
+    # fail silently if they aren't calcuable. We use a for loop so that failing
+    # one dimension doesn't break others. We could possibly use something like
+    # abind::abind here and bind together the insertion vectors in array form,
```
GShotwell: I think this is a good use of a for loop. Another dependency which would let us handle this situation a bit more cleanly is [purrr::safely](https://purrr.tidyverse.org/reference/safely.html). I think that purrr would help in a lot of other areas as well, so it might be a better dependency than `abind`. I guess the other way you could do with it is to use an `apply` function, and then return something which could be filtered out of the returned list, but I'm not sure that's any clearer than the loop.

R/transform-calculate.R
```diff
@@ -86,23 +86,33 @@ collateCats <- function (inserts, var_cats) {
...
+    # list because for insertions that have the same anchor, we want to maintain
+    # the order as it is stored in the API. To do this we insert the last one 
+    # first so that when we insert ones before that they are higher up (closer 
+    # to the category anchor)
+    # Categoies: A, B, C 
+    # Insertions: [{name = first, anchor = A}, {name = second, anchor = A}]
+    # Desired result: A, first, second, B, C
+    # Result if not `rev(inserts)`: A, second, first, B, C
+    for (insert in rev(inserts)) {
+        pos <- findInsertPosition(insert, cats_out, last_category)
```
GShotwell: Does this fail if `last_category` is `NULL`? If so are we sure that `ids(var_cats)` has entries?

tests/testthat/test-variable-transforms.R
```diff
@@ -49,12 +49,12 @@ insrts_list <- list(list(anchor = 0, name = "First one",
 insrts <- Insertions(data=insrts_list)
 
 test_that("findInsertPosition", {
```
GShotwell: It might be worth adding a couple of tests with some invalid `last_category` values like `NULL`. This is just to be defensive against future changes.

R/cube-transforms.R
```diff
@@ -86,37 +110,187 @@ setMethod("subtotalArray", "CrunchCube", function(x, headings = FALSE) {
...
+        array <- subsetTransformedCube(array, dims_list, useNA)
+    })
 
-    # TODO: calculate column transforms
     return(array)
 }
 
-subsetTransformedCube <- function (array, cube) {
+
+#' apply a function against a dimension
```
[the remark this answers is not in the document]
jonkeane: `apply` is basically the core of the helper function `applyAgainst`: https://github.com/Crunch-io/rcrunch/pull/262/files#diff-fc6c252cf7a4ec51064a3ca7fde28541R279 it's followed by an `aperm` to fix the ordering problem you noted. Almost all of the extra code there is to do the right thing for 1d (which the `apply` code doesn't) and ensure that names are good. 

The main reason I broke it off into a function is for readability, I think it's easier to read and grok what the function does without also necessarily having to think about subtotaling or the like.

R/transform-calculate.R
jonkeane: Good catch, it generally doesn't fail, unless the insertion was being inserted at the end. I've added a check to `collateCats` to confirm that there is at least 1 category before proceeding which should prevent this as a possibility (since all categories have to have an id).

tests/testthat/test-variable-transforms.R
jonkeane: I've added a check on the one calling function. Since `findInsertPosition` is intended to be pretty internal I don't think we need to have input checks that passing a `NULL` there works in most cases except for the one that it absolutely couldn't work in. I'm inclined to leave the test for catching the problem with `collateCats` and leave it be. Unless we need to expand the use of `findInsertPosition` elsewhere / export it.

R/transform-calculate.R
```diff
@@ -86,23 +86,37 @@ collateCats <- function (inserts, var_cats) {
     # setup an empty AbstractCategories object to collate into
     cats_out <- AbstractCategories()
     cats_out@.Data <- var_cats
+    
+    if (length(var_cats) < 1) {
+        halt("Can't collateCats with less than one categories.")
```
nealrichardson: I'm not a linguist, but "one categories" looks wrong to me. And "less" should be "fewer". But aren't you really trying to say "no categories"?

tests/testthat/test-variable-transforms.R
```diff
@@ -135,14 +141,27 @@ test_that("Converting insertions to subtypes works", {
 
     # collate and check
     collated <- collateCats(c(insrts_subtyped, insrt_heads_subtyped), cats)
-    expect_true(all(are.Subtotals(collated[c(2, 5, 8, 15, 16, 17)])))
-    expect_true(all(are.Headings(collated[c(1)])))
+    expect_true(all(are.Subtotals(collated[c(1, 5, 8, 15, 16, 17)])))
+    expect_true(all(are.Headings(collated[c(2)])))
```
nealrichardson: `c()` doesn't do anything here

R/cube-transforms.R
```diff
@@ -86,37 +118,187 @@ setMethod("subtotalArray", "CrunchCube", function(x, headings = FALSE) {
 #'
 #' @aliases subtotalArray
 #' @export
-applyTransforms <- function (x, array = cubeToArray(x), ...) {
-    # if any of the dimensions are subvariables, don't even attempt to calculate
-    # transforms or insertions.
-    if (any(getDimTypes(x) == "ca_items")) {
-        return(array)
-    }
+applyTransforms <- function (x,
```
nealrichardson: I've noticed several places where you've replaced `x <- as.array(cube)` with `x <- applyTransforms(x)`. Should `as.array.CrunchCube <- applyTransforms`?

R/cube-transforms.R
```diff
@@ -86,37 +118,187 @@ setMethod("subtotalArray", "CrunchCube", function(x, headings = FALSE) {
...
+                             ...) {
+    trans_indices <- which(vapply(transforms_list, Negate(is.null), logical(1)))
+    
+    # Try to calculate the transforms for any dimensiton that have them, but
+    # fail silently if they aren't calcuable. We use a for loop so that failing
+    # one dimension doesn't break others. We could possibly use something like
+    # abind::abind here and bind together the insertion vectors in array form,
+    # but that would add a dependency
+    for (d in trans_indices) {
+        try({
```
nealrichardson: Why so much `try()` in here?

R/cube-result.R
```diff
@@ -203,10 +203,12 @@ keepWithNA <- function (dimension, marginal, useNA) {
     } else {
         ## !always means either drop missing always, or only keep if there are any
         out <- !dimension$missing
+        names(out) <- dimension$name
         if (useNA == "ifany") {
             ## Compare against "marginal", the counts, to know which missing
-            ## elements have "any"
-            out <- out | marginal > 0
+            ## elements have "any". We need to subset by name here because the
+            ## margin might include insertions
```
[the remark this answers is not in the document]
jonkeane: This comes up when  `evalUseNA` is called at https://github.com/Crunch-io/rcrunch/pull/262/files#diff-fc6c252cf7a4ec51064a3ca7fde28541R311 after the subtotals have been inserted into the array.

R/cube-transforms.R
jonkeane: I started down that path in this PR, but found there were a bunch of little things I would have to change. I think that's probably a good idea, but I think we should defer it if possible and do it separately for ease of review.

R/cube-transforms.R
jonkeane: I added the `try()`s here to prevent a malformed (or unanticipated) transform from preventing the cube from being displayed or being accessible.

nealrichardson: I'm curious how the `try()`s might interact. You have a for loop and a try within each iteration, then a try around something else at the bottom. If one of my loop iterations errors, do they all error? Do they leave `array` in a bad state? Will the next `try` necessarily also catch an error? 

Are there real cases where you can trigger these errors that you're catching? Are they tested?

I'm wondering whether it might better to have one big tryCatch around everything that on error just returns `array` unmodified.

jonkeane: The `try()` inside the loop is an attempt to provide transforms for dims that have valid transforms. I've added a few tests for that condition now, and it does work as expected (and in the process caught a problem with the `prettyPrint` code that was preventing them from being formatted). 

I've removed the second try because I don't think it's necessary now. (It wasn't being triggered by failures above, at all.)

That being said, this is very much an edge case that I'm not sure anyone would hit, so maybe we do just want to wrap the whole thing in `tryCatch()` and call it a day.

PR #264, opened 2018-05-30
Export utility functions

GShotwell:
Export a couple of utility functions that we need from crplyr and crunchy


nealrichardson:
LGTM

PR #269, opened 2018-06-07
Mr builder gadget

GShotwell:
Adds binding to MR builder from crunchy.


inst/rstudio/addins.dcf
```diff
@@ -3,3 +3,7 @@ Description: Select Crunch Dataset
 Binding: listDatasets(shiny = TRUE)
 Interactive: true
 
+Name: Array Builder
```
nealrichardson: Should this be in `crunchy` instead of `crunch`?

R/make-array.R
```diff
@@ -80,6 +80,26 @@ makeMR <- function (subvariables, name, selections, ...) {
...
+#' use the gadget you must have at least one CrunchDataset loaded into the global
+#' environment.
+#'
+#' @return a valid call to `makeArray()` or `makeMR()`
+#' @export
+makeArrayGadget <- function(){
+    if ("makeArrayGadget" %in% ls(envir = asNamespace("crunchy"))) {
+        crunchy::makeArrayGadget()
+    } else {
+        stop("Please install the latest version of crunchy to access this function.")
```
nealrichardson: Use `halt`

R/make-array.R
```diff
@@ -80,6 +80,26 @@ makeMR <- function (subvariables, name, selections, ...) {
...
+#' construct without being able to investigate the available variables, and
+#' their categories. This shiny gadget lets you select subvariables from the
+#' dataset list, and ensures that those variables have consistent categories. To
+#' use the gadget you must have at least one CrunchDataset loaded into the global
+#' environment.
+#'
+#' @return a valid call to `makeArray()` or `makeMR()`
+#' @export
+makeArrayGadget <- function(){
+    if ("makeArrayGadget" %in% ls(envir = asNamespace("crunchy"))) {
```
nealrichardson: If you wanted to test this function, or at least the `stop` side of the `if` (and I'm not saying you need to), you could factor out this check to something like `has_function <- function (fun, pkg)`, and then mock `has_function` to return `FALSE`.

inst/rstudio/addins.dcf
GShotwell: It should be, I thought we'd do this in a couple of PRs to make it a bit easeir to id the PR which brought each gadget into the two packages.

R/make-array.R
GShotwell: Oh that's a good strategy. I think for this case it's not a test that adds a whole lot of value so I'd prefer to not include it. But still a good idea to add a function to check if a package has the right function.

PR #272, opened 2018-06-11
Resolves #257

### LINKED ISSUE #257: Numerics with any `NA`s return all NAs with `as.data.frame(..., force = TRUE)`
`as.data.frame()` without `force` works fine, but the `columnParser("Numeric")` produces all `NA`s when there are any missing responses in the variable. 

```
df <- data.frame('a'=c(1:9,NA), 'b'=factor(letters[1:10]))
ds <- newDataset(df, "Testing data frame exports")
df <- as.data.frame(ds)
class(df) 
table(df$a, useNA='always')

df_force <- as.data.frame(df, force=TRUE)
class(df_force)
table(df_force$a, useNA='always')
```

GShotwell:
Resolves bug whereby numeric columns with an NA value were coerced to all-NA.


R/as-data-frame.R
```diff
@@ -112,6 +112,11 @@ csvToDataFrame <- function (csv_df, crdf) {
             cp <- columnParser("categorical")
             sub_a <- aliases(subvariables(v))
             return(structure(lapply(csv_df[sub_a], cp, v, mode), .Names=sub_a))
+        } else if (is.Numeric(v)) {
+            # We use this to coerce the missing entries to NA, so that warning is
```
nealrichardson: This comment should be more explicit about what the data looks like such that coercion is needed.

PR #276, opened 2018-06-18
Index redux

jonkeane:
New, updated definition of index table (for column index): 
```
cube <- crtabs(~ x + y, ds)
uni_cube <- crtabs(~ x, ds)
index.table(cube, 2) == prop.table(cube, 2)/prop.table(uni_cube) * 100
# TRUE
```

To accomplish this, I implemented a function `collapse.dimensions` that collapses and removes whatever dimensions are given in `margin`. This returns a cube object that is mathematically and functionally equivalent (and ideally, indistinguishable from) requesting the lower-dimensional cube separately.

One place of possible improvement and extension is to make the `user2real` dimension mapping function more general and use it elsewhere in cube processing.


R/cube-result.R
```diff
@@ -388,6 +411,122 @@ setMethod("margin.table", "CrunchCube", function (x, margin=NULL) {
...
+    # grab the old cube structure and over-write it with new subset data.
+    out <- x
+    
+    # iterate through measures, collapsing the dimension(s) specified.
+    out@arrays[] <- lapply(out@arrays, function(arr){
+        # off_margins here are the margins that we want to keep, however because
+        # we treat items dimensions differently, we need to add them to the
+        # off_margins, and deal with them after we collapse by summing
+        off_margins <- margins_to_keep
+        if (any(collapsed_items)) {
```
GShotwell: I don't love having two conditional branches here and at 455, is it possible to combine the two blocks of conditional code by repeating the `array <- as.array(apply(arr, off_margins, sum))` call?  Personally I find repetative code with fewer branches a bit easier to follow.

R/cube-result.R
```diff
@@ -388,6 +411,122 @@ setMethod("margin.table", "CrunchCube", function (x, margin=NULL) {
...
+    margins_to_keep <- setdiff(seq_along(x@dims), margins_to_collapse)
+    
+    # if there are any _items in the margin to collapse, be wary!
+    collapsed_items <- endsWith(getDimTypes(x)[margins_to_collapse], "_items")
+
+    # grab the old cube structure and over-write it with new subset data.
+    out <- x
+    
+    # iterate through measures, collapsing the dimension(s) specified.
+    out@arrays[] <- lapply(out@arrays, function(arr){
```
GShotwell: This is a pretty big function to stick in a `lapply` and I found it a bit hard to follow. It might be easier to read if it were pulled into its own function, and purified (make the scoped arguments explicit arguments). That way you could pull out some of the motivation comments into function documentation.

R/cube-result.R
```diff
@@ -388,6 +411,122 @@ setMethod("margin.table", "CrunchCube", function (x, margin=NULL) {
...
+    # we needed to include the selected dim of MRs in off_margins above, but
+    # when subsetting dimensions we definitely do not want them to be included.
+    out@dims <- out@dims[margins_to_keep]
+    out$query$dimensions <- out$query$dimensions[margins_to_keep]
+    out$result$dimensions <- out$result$dimensions[margins_to_keep]
+
+    # Need to calculate the total missing to fill in, to do this, set all
+    # non-missing cells to zero, and then sum across all dimensions (with the
+    # exception of `_items` dimensions)
+    unweighted_counts <- as.array(out@arrays$.unweighted_counts)
```
GShotwell: What does `as.array` do here? Are there cases where `.unweighted_counts` is not an array?

R/cube-result.R
```diff
@@ -388,6 +411,122 @@ setMethod("margin.table", "CrunchCube", function (x, margin=NULL) {
...
+    non_missing <- evalUseNA(unweighted_counts, out@dims, "no")
+    all_missings <- do.call("[<-", c(list(x=unweighted_counts), non_missing, value = 0))
```
GShotwell: Something about this makes me a bit uncomfortable, maybe just because I don't associate assignment with summing, is there a way to accomplish this which doesn't rely on a `do.call`?

R/cube-result.R
jonkeane: Turns out, you don't need this first if since `collapsed_items` is a logical, so I've removed it.

R/cube-result.R
jonkeane: Good catch, this was left over from previous code that needed it.

R/cube-result.R
```diff
@@ -388,6 +411,122 @@ setMethod("margin.table", "CrunchCube", function (x, margin=NULL) {
...
+#' @param margin the margin or margins for the user cube to be translated
+#' @param dimTypes dimension types from `getDimTypes()` (by default:
+#'   `getDimTypes(cube)`)
+#' @param cube the cube to translate the margin for (optional if `dimTypes` is
+#'   explicitly supplied)
+#'
+#' @return margin or margins in the higher-dimension real cube
+#'
+#' @keywords internal
+user2real <- function(margin, dimTypes = getDimTypes(cube), cube) {
```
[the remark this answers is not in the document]
jonkeane: I was also skeptical the first time I saw it, but it works, with the caveat that sometimes there are odd namespace search issues (for example: if I had another argument called `getDimTypes` that would error), but in this case it works, and I think it is slightly more elegant than the conditional.

R/cube-result.R
```diff
@@ -388,6 +411,122 @@ setMethod("margin.table", "CrunchCube", function (x, margin=NULL) {
...
+    if (is.null(margin)) { 
+        # If margin is null, return null
+        return(NULL)
+    }
+    full_margins <- seq_along(dimTypes)
```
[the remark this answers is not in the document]
jonkeane: It turns out, no! I've removed it.

R/cube-index-table.R
```diff
@@ -1,13 +1,14 @@
...
 #' Index tables are percentages of percentages. They take the percentage from
-#' `prop.table()` and divide that by the proportions of the other margin.
+#' `prop.table(cube, margin)` and, by default, divide that by the proportions of the other margin. The `baseline` argument can be used to provide baseline proportions to compare against.
 #'
 #' `index.table()` is only implemented for 2 dimensional cubes. If you need to
 #' calculate indexes for a higher dimension Cube, please slice the cube first.
 #'
 #' @param x A CrunchCube to calculate index table for
 #' @param margin which margin to index against (1 for rows, 2 for columns)
+#' @param baseline an arbitrary set of proportions to compare the table given in `x` to. Useful for comparing two separate cubes. `baseline` must have the same length as the extent of the dimension given in `margin`.
```
nealrichardson: What's the use case for having this argument?

R/cube-result.R
```diff
@@ -434,3 +402,147 @@ setMethod("bases", "CrunchCube", function (x, margin=NULL) {
...
+#'
+#' @return margin or margins in the higher-dimension real cube
+#'
+#' @keywords internal
+#'
+#' @name margin-translation
+NULL
+
+#' @rdname margin-translation
+user2real <- function(margin, dimTypes = getDimTypes(cube), cube) {
```
nealrichardson: These function names are too ambiguous. They should say "margin" in them, I think.

R/cube-result.R
```diff
@@ -434,3 +402,147 @@ setMethod("bases", "CrunchCube", function (x, margin=NULL) {
...
+#' in the resulting table (array). The `bases` function takes a "margin"
+#' argument to work like `margin.table`, or with `margin=0` gives all cell
+#' counts.
+#'
+#' `collapse.dimensions` returns a cube that collapses the dimensions given in
+#' `margin`. This is useful if you want to get counts that are equivalent to a
+#' univariate cube from a multivariate cube. For example
+#' `collapse.dimensions(crtabs(~ fruit + pets, ds), 2)` will be equal to
+#' `crtabs(~ fruit, ds)` and `collapse.dimensions(crtabs(~ fruit + pets, ds),
+#' 1)` will be equal to `crtabs(~ pets, ds)`.
```
nealrichardson: Perhaps note that for the simple (non-MR) case, this is equivalent to `apply(x, margin, sum)`. Maybe that also helps with the function name, which I don't find particularly obvious. It's a bit cheeky, but the generalization of `rowSums()` and `colSums()` might be `dimSums()` ;)

Also, in what way are they "equal"? Are they identical cube objects? Have equivalent counts/margins/percentages/etc. based on the supported methods?

R/cube-collapse-dimensions.R
```diff
@@ -0,0 +1,127 @@
+#' @rdname cube-computing
+#' @export
+setMethod("collapse.dimensions", "CrunchCube", function (x, margin=NULL) {
```
nealrichardson: Unless you're planning on defining different `collapse.dimensions` methods for other objects, there's no reason to define this as a method instead of a function.

R/cube-collapse-dimensions.R
```diff
@@ -0,0 +1,127 @@
...
+    # with other measures below) then we would inflate the number of responses.
+    collapsed_items <- endsWith(getDimTypes(x)[margins_to_collapse], "_items")
+    
+    # grab the old cube structure and over-write it with new subset data.
+    out <- x
+    
+    # iterate through measures, collapsing the dimension(s) specified.
+    out@arrays[] <- lapply(
+        out@arrays,
+        collapse_with,
```
nealrichardson: This only works with measures that are counts. It's probably not worth implementing for mean et al. (and you probably can't for median et al.) since this function only exists to support `index.table`, and even then only to avoid another API request.

R/cube-collapse-dimensions.R
```diff
@@ -0,0 +1,127 @@
...
+        out@arrays,
+        collapse_with, 
+        margins_to_keep = margins_to_keep, 
+        margins_to_collapse = margins_to_collapse, 
+        collapsed_items = collapsed_items)
+    
+    # we needed to include the selected dim of MRs in off_margins above, but
+    # when subsetting dimensions we definitely do not want them to be included.
+    out@dims <- out@dims[margins_to_keep]
+    out$query$dimensions <- out$query$dimensions[margins_to_keep]
```
nealrichardson: I suspect that munging things that aren't in `@dims` and `@arrays` is not worth the effort to try to get right (maybe you have covered everything here, but I'm skeptical). IIRC we don't bother with this in the `[` method. I wouldn't want to cause an error for a user in this section while trying to update them, nor would I want to bother maintaining this logic when the API evolves. 

I'm pretty certain that all `CrunchCube` methods only look at dims and arrays and never reach into the query result. And in practice, all you care is that `prop.table` works so that you can use this in `index.table`.

R/cube-result.R
jonkeane: Hah, I like `dimSums()`, I'll `s/collapse.dimensions/dimSums/g` and coalesce the arguments if i can.

The cube objects are identical, though like you point out below, that's only really true for count measures.

R/cube-collapse-dimensions.R
jonkeane: Yeah, that was my plan. I'll add an error if a non-count measured cube is passed so that we don't even attempt this for other measures. We could decide later if we wanted to actually resort to querying for the other cube if someone ever wants it. But for now an error is probably just fine.

R/cube-collapse-dimensions.R
jonkeane: Originally this was to ensure I was not fooling myself in believing I could totally reconstruct the univariate cube, but I agree — it's probably more work than necessary to maintain, so I've removed it and now just test that `@arrays` and `@dims` are equal in the tests.

R/cube-index-table.R
jonkeane: The one I can think of now is if your cube's margin doesn't reflect the population you're interested in. The classic case we've been hearing is that you want to index to a nationally representative sample, so index with our default baseline will be that as long as the cube is weighted to reflect that sample. If the cube is unweighted or weighted to some other sample, the margin might not be the right baseline.

I also find it easier to reason about and experiment with to see what's going on to be able to provide my own baseline and see how that changes the values coming out.

jonkeane:
This is ready for another round of PR.

R/cube-collapse-dimensions.R
```diff
@@ -0,0 +1,117 @@
...
+    selecteds <- is.selectedDimension(x@dims)
+    check_margins(margin, selecteds)
+    
+    # ensure that the cube is a counts cube
+    only_count_cube(x)
+    
+    # translate from user-cube margins to real-cube margins and establish the
+    # two groups of margins: those to collapse and those to keep
+    margins_to_collapse <- user2realMargin(margin, cube = x)
+    margins_to_keep <- setdiff(seq_along(x@dims), margins_to_collapse)
```
nealrichardson: I don't think you need this argument. You can just send `margins_to_collapse` to the internal function and it can do `off_margins <- setdiff(seq_along(dim(array_in), margins_to_collapse)`.

jonkeane: Yup yup, I've removed the argument to `collapse_dims()` though I do use both `margins_to_keep` and `margins_to_collapse` in `dimSums` a few places each so I've kept their setting here.

In looking at this I also realized that with the name change, what margin refers to (the margins to keep or the margins to collapse) was at odds with how `rowSums` and `colSums` worked, so I've actually flipped it — which in hindsight I think is actually the much much more intuitive version answering the question: "I have this cube and I want to get a cube with only dimension X in it." much easier than having to figure out what the opposite set of margins is.

PR #280, opened 2018-06-19
Cube fixtures to inst

GShotwell:
This allows us to access cube fixtures from other packages and (I think) doesn't require any changes to test code.


inst/crunch-test.R
```diff
@@ -9,6 +9,11 @@ skip_locally <- function (...) {
 }
 
 loadCube <- function (filename) {
+    file <- sub('.*\\/', '', filename)
+    file <- system.file(paste0("cubes/", file), package = "crunch")
```
nealrichardson: You should just be able to pass `filename` to `system.file`, it calls `file.path` inside.

tests/testthat/test-misc.R
```diff
@@ -160,7 +160,7 @@ test_that("default date formater", {
 
 test_that("Cubify works with many dimensions", {
     # 1d
-    cube_json <- fromJSON(test_path("cubes/univariate-categorical.json"))
+    cube_json <- fromJSON(system.file("cubes/univariate-categorical.json", package = "crunch"))
```
nealrichardson: I know you're not changing this here, but I don't understand these tests. AFAIK you can get `$result$measures$count$data`, unmodified from the server response, from the CrunchCube without having to `fromJSON` it separately.

jonkeane: The original intent here was confirming the behavior of cubify with dimensions greater than 2. I wrote it this way to confirm that there was no re-arranging going on when we read cubes in. There isn't so we could change this to use the `$result$measures$count$data` if we wanted.

inst/crunch-test.R
```diff
@@ -9,7 +9,21 @@ skip_locally <- function (...) {
...
 
 loadCube <- function (filename) {
-    crunch:::CrunchCube(jsonlite::fromJSON(filename, simplifyVector=FALSE)$value)
+    file <- system.file(filename, package = "crunch")
+    if (nchar(file) > 0) {
+        filename <- file
+    }
+
+    cube_json <- jsonlite::fromJSON(filename, simplifyVector=FALSE)
+
```
nealrichardson: Why not `if ("value" %in% names(cube_json)) cube_json <- cube_json$value`? `tryCatch` makes it seem like we don't know the shape of data we're getting.

PR #290, opened 2018-07-13
Allow write.csv to write hidden variables.

GShotwell:
A user encountered a crplyr bug which prevented the export of hidden variables. In fixing that I discoverd that the include.hidden argument wasn't being properly passed down to `write.csv` this PR resolves that issue. 

Initial bug report: https://github.com/Crunch-io/crplyr/issues/6


tests/testthat/test-export-dataset.R
```diff
@@ -69,6 +69,22 @@ with_mock_crunch({
...
+
+        expect_POST(write.csv(ds, file = "", include.hidden = TRUE),
+            url,
+            post_request)
+
+        # ensure that include.hidden is passed down from as.data.frame to write.csv
+        expect_POST(as.data.frame(ds, include.hidden = TRUE, force = TRUE),
+            url,
+            post_request)
+
```
jonkeane: Could you add a test here (and it would probably also be good to have a similar integration test below) for what happens when you request a hidden variable explicitly by name without specifying `include.hidden`? According to the API docs this _should_ work without needing to send any additional parameters, but it would be good to be able to know that from tests.

R/crunch-data-frame.R
```diff
@@ -2,59 +2,66 @@
...
+
+    # When a user asks for a hidden variable with ds[, "hidden_var_name"] we should
+    # give it to them, and so overrride the include.hidden argument
+    allvars <- allVariables(dataset)
+    dsvars <- ShojiCatalog(crGET(self(allvars), query=list(relative="on")))
+    if (length(allvars) != length(dsvars) && # check that dataset has avariable subset
+            length(hiddenVariables(dataset)) > 0  # check that user is asking for hidden variable
+    ) {
+        include.hidden <- TRUE
     }
```
jonkeane: I'm not sure you actually need this machinery. It looks like what you're trying to do is see if the dataset object that's being passed is a subset of the variables available, and if it is, and the dataset has any hidden variables then you set `include.hidden` to `TRUE`, but that shouldn't be necessary here: I don't think we need to override `include.hidden` here. We can just include them in the promises made later by using line 43 below, and then we can decide to fetch them or not in `as.data.frame.CrunchDataFrame`

tests/testthat/test-as-data-frame.R
```diff
@@ -94,11 +95,12 @@ with_mock_crunch({
...
-        new_ds_df <- as.data.frame(new_ds)
-        expect_silent(
+        new_ds_df <- as.data.frame(new_ds, include.hidden = FALSE)
+        expect_warning(
             expect_equal(names(csvToDataFrame(csv_df, new_ds_df)),
-                c("gender", "location", "subvar2", "subvar1", "subvar3", "textVar",
-                    "starttime"))
+                c("birthyr", "gender", "location", "subvar2", "subvar1", "subvar3", "textVar",
+                    "starttime")),
+
```
jonkeane: What's the warning that we are expecting here? `"Variable birthyr is hidden"`?

tests/testthat/test-as-data-frame.R
```diff
@@ -228,6 +224,25 @@ with_test_authentication({
...
+    ds$hidden_var <- 1:20
+    ds <- hideVariables(ds, "hidden_var")
+
+    test_that("as.data.frame(force) retrieves hidden variables", {
+        skip_locally("Vagrant host doesn't serve files correctly")
+        expect_equal(hiddenVariables(ds), "hidden_var")
+
+        expect_warning(df <- as.data.frame(ds, force = TRUE),
+            "Variable hidden_var is hidden")
+        expect_equal(names(df), c("v1", "v2", "v3", "v4", "v5", "v6", "hidden_var"))
```
jonkeane: My expectation was that requesting the whole dataset here would return only the visible variables.

Here's what I thought we decided on for behavior:
`as.data.frame.CrunchDataset` returns promises of everything, but with `force=TRUE|as.data.frame.CrunchDataFrame()` if there’s no filter, you never get hidden variables in the export, if you explicitly ask for hiddens you get all of them (if there’s no filter), and if there is a filter you get what you filtered if it’s hidden or not?

tests/testthat/test-as-data-frame.R
```diff
@@ -228,6 +224,25 @@ with_test_authentication({
...
+
+        expect_warning(
+            df <- as.data.frame(ds[, c("v1", "hidden_var")], force = TRUE),
+            "Variable hidden_var is hidden"
+        )
+        expect_equal(names(df),
+            c("v1", "hidden_var"))
```
jonkeane: This looks like the right behavior / expectation to me: If I explicitly ask for a variable (that happens to be hidden) I get that variable (along with any other variables I might have requested)

tests/testthat/test-as-data-frame.R
jonkeane: Once you make this change, you'll probably want to duplicate this test with `include.hidden = TRUE` in `as.data.frame()` and confirm that you do get everything (including hiddens) when you request it that way.

tests/testthat/test-export-dataset.R
```diff
@@ -69,6 +69,43 @@ with_mock_crunch({
         expect_error(write.csv(ds[c("gender", "birthyr", "gender", "gender", "birthyr")], file=""),
             "Duplicate variable references: gender and birthyr")
     })
+
+    url <- 'https://app.crunch.io/api/datasets/3/export/csv/'
+    post_request <- '{"filter":null,"where":{"function":"select","args":[{"map":{"66ae9881e3524f7db84970d556c34552":{"variable":"https://app.crunch.io/api/datasets/3/variables/gender/"},"f78ca47313144b57adfb495893968e70":{"variable":"https://app.crunch.io/api/datasets/3/variables/birthyr/"},"d7c21314ca9e453c93069168681a285c":{"variable":"https://app.crunch.io/api/datasets/3/variables/starttime/"}}}]'
```
jonkeane: Could you add line breaks here?

tests/testthat/test-export-dataset.R
```diff
@@ -158,4 +195,21 @@ with_test_authentication({
...
+
+    test_that("We can export hidden variables", {
+        skip_locally("Vagrant host doesn't serve files correctly")
+        ds$hidden_var <- 1:20
+        ds <- hideVariables(ds, "hidden_var")
+        filename <- tempfile()
+
+        write.csv(ds, file = filename, include.hidden = TRUE)
+        df <- read.csv(filename)
+        expect_true("hidden_var" %in% names(df))
```
jonkeane: I think it would be good to be explicit here about what all variables are in the df, not just that the hidden one is there.

tests/testthat/test-export-dataset.R
```diff
@@ -158,4 +195,21 @@ with_test_authentication({
...
+
+        # When variable is specified but include.hidden is ommitted
+        write.csv(ds[, "hidden_var"], file = filename)
+        df <- read.csv(filename)
+        expect_true("hidden_var" %in% names(df))
```
jonkeane: Likewise here, it would be good to check that `names(df) == "hidden_var"` and _doesn't_ include other vars.

tests/testthat/test-export-dataset.R
```diff
@@ -158,4 +195,21 @@ with_test_authentication({
         expect_identical(names(df2), c("v1", "v2", "v4"))
         expect_identical(is.na(df2$v1), is.na(df$v1))
     })
+
+    test_that("We can export hidden variables", {
+        skip_locally("Vagrant host doesn't serve files correctly")
+        ds$hidden_var <- 1:20
+        ds <- hideVariables(ds, "hidden_var")
+        filename <- tempfile()
+
```
jonkeane: We should also have a test in this pattern that `write.csv(ds, file = filename)` (ie `include.hidden = FALSE`) returns no hidden variables.

R/crunch-data-frame.R
```diff
@@ -2,59 +2,66 @@
...
 #' if only one row is left.
 #'
 #' @name CrunchDataFrame
 NULL
 
 CrunchDataFrame <- function (dataset, row.order = NULL,
-                             categorical.mode = "factor",
-                             include.hidden = FALSE) {
+    categorical.mode = "factor",
+    include.hidden = TRUE) {
```
jonkeane: Also, I don't think we should change the default argument here to be `TRUE`, it should remain `FALSE` if possible.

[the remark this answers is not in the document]
GShotwell: Sorry neglected to update my settings.

R/crunch-data-frame.R
GShotwell: Ah, look at that. Thanks!

tests/testthat/test-as-data-frame.R
```diff
@@ -94,19 +95,21 @@ with_mock_crunch({
...
         new_ds_df <- as.data.frame(new_ds, include.hidden = TRUE)
         expect_warning(
             expect_equal(names(csvToDataFrame(csv_df, new_ds_df)),
-                c("birthyr", "gender", "location", "subvar2", "subvar1", "subvar3", "textVar",
-                    "starttime")),
-            paste0("Variable birthyr is hidden"))
+                         c("birthyr", "gender", "location", "subvar2", "subvar1", "subvar3", "textVar",
+                           "starttime")),
+            "Variable birthyr is hidden")
     })
```
jonkeane: Could we also have a test here showing what happens when `new_ds_df <- as.data.frame(new_ds); (names(csvToDataFrame(csv_df, new_ds_df)) ...`? I might be missing this test somewhere else, but if not we should include it to cover all of our bases

tests/testthat/test-export-dataset.R
```diff
@@ -69,6 +69,60 @@ with_mock_crunch({
...
+        # the export API. We work around this by asking for hidden variables when
+        # the user specifies them in the variable subset. We don't currently have a
+        # good way to do this when the user subsets the dataset with all the variable
+        # names. This problem could also go away if the export endpoint allowed us
+        # to pass hidden variable names like regular variable names.
+        expect_POST(
+            write.csv(ds[, c("gender", "birthyr", "starttime")], file = ""),
+            url,
+            post_request)
+    })
```
jonkeane: This test should pass already, no? Do we need to skip it?

R/as-data-frame.R
```diff
@@ -45,7 +45,7 @@
 #'  Crunch Dataset order will be used.
 #' @param categorical.mode what mode should categoricals be pulled as? One of
 #' factor, numeric, id (default: factor)
-#' @param include.hidden logical: should hidden variables be included? (default: `FALSE`)
+#' @param include.hidden logical: should hidden variables be included?
```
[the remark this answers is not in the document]
GShotwell: Ah right, I forgot to change it back after restoring the default.

tests/testthat/test-as-data-frame.R
GShotwell: Now that I think about it I think we can just remove this test entirely. Since we always have hidden variables in the `names(cdf)` `csvToDataFrame` is going to do the same thing regardless of what is passed to include.hidden.

tests/testthat/test-export-dataset.R
GShotwell: I don't think it does. The reason is that we don't have a way to tell that the user has specified all the column names instead of selecting the whole dataset.

PR #296, opened 2018-07-31
Z score redux

jonkeane:
Allow for comparing one column/row against another column/row and get a z-score. Wrap the pairwise comparison of all columns/rows in a table to a given baseline.


R/cube-residuals.R
```diff
@@ -1,66 +1,246 @@
...
-#' @references
-#' Agresti, A. (2007)
-#' \emph{An Introduction to Categorical Data Analysis, 2nd ed.},
-#' New York: John Wiley & Sons. Page 38.
-#' @importFrom stats chisq.test
-#' @name cube-residuals
-#' @aliases rstandard,CrunchCube-method cube-residuals
+#' @return the z-score for the column or row given in `x`
+#' 
+#' @name dimension-comparison
```
jonkeane: test

R/cube-residuals.R
```diff
@@ -1,66 +1,246 @@
...
+
+#' @rdname dimension-comparison
+#' @export
+compareDims <- function (cube, dim = c("cols", "rows"), baseline, x) {
+    dim <- match.arg(dim)
+    
+    # grab the names of extents for each dimensions
+    # TODO: we shouldn't need to do this if we can just pass baseline/x into the
+    # subsetting function when we can subset by name _or_ index natively
+    if (dim == "cols") {
```
slobodan-ilic: I don't know if it's me coming from py or expecting the insane functional style from r, but it's too much if/elses for my taste. I'd strive to keep this a one-liner, if there are no downsides to it in r (such as falsely increasing the test coverage, e.g.).

Would it be possible to write something like this:
`names <- colnames(cube) if(dim == "cols") else rownames(cube)`? Is this a big no-no because of some reason I can't see?

slobodan-ilic: Yeah I've seen the `ifelse`, it's awful, hence not mentioned in my comment. Sorry for the blunt pre assumption that the thing found on stack overflow would just work...

R/cube-residuals.R
```diff
@@ -23,46 +67,185 @@ standardizedMRResiduals <- function(cube) {
...
+        names <- rownames(as.array(cube))
     }
 
-    if (any(startsWith(getDimTypes(model), "mr_"))) {
-        return(standardizedMRResiduals(model))
+    # convert to numeric indices
+    # TODO: remove when we can accept either characters or indices and pass them
+    # to subsetting unmolested
+    if (is.character(baseline)) {
+        baseline_ind <- which(names == baseline)
```
GShotwell: Will this cause problems for duplicated or out of order character vectors? It might be easier to error until we merge in the cube-subset enhancement PR

R/cube-residuals.R
```diff
@@ -23,46 +67,185 @@ standardizedMRResiduals <- function(cube) {
...
+#' Given a single baseline column compare each other row or column against this
+#' baseline. Internally this function uses `compareDims()` iteratively.
+#'
+#' *Warning* since there is more than one comparison being made against each
+#' baseline the z-scores, and especially the p-values derived from these
+#' z-scores should be interpreted with caution. Using standard p-value cutoffs
+#' will result in anti-conservative interpretations because of the multiple
+#' comparisons problem. Adjustments to p-value cut offs (e.g. Bonferonni
+#' correction) should be used when interpreting z-scores from the
+#' `compare[Rows|Cols|Dims]Pairwise()` family of functions.
```
GShotwell: It might be good to include a reference (maybe a wikipedia article or the like) here. This paragraph is probably only meaningful to people who have a very good understanding of p-values, which is not that many people.

R/cube-residuals.R
```diff
@@ -23,46 +67,185 @@ standardizedMRResiduals <- function(cube) {
...
+#' @param baseline a character, the column to use as a baseline to compare
+#'   against all other columns
+#' @param ... arguments passed from `compareRowsPairwise()` or
+#'   `compareColsPairwise()` to `compareDimsPairwise()` (i.e. `baseline`)
+#'
+#' @return an array of z-score for the all the columns or rows compared to
+#'   `baseline`. The `baseline` column is all 0s
+#'
+#' @name dimension-comparison-pairwise
+NULL
```
GShotwell: I'm not sure we need to do this documentation hack anymore. I think if you stick this above the generic the documentation will generate correctly. You might have to swap `rdname` for `name` (I can never remember which goes where)

R/cube-residuals.R
GShotwell: `ifelse` isn't awful, it's just vectorized. Most of the time when doing condition handling in the R package you want the `if .... else....` construct, but it's very useful operating on vectors for instance when say creating a new variable in a data frame: 
```
 `mtcars$mpg_recode <- ifelse(mtcars$mpg > 200, "large", "small")
```

R/cube-residuals.R
jonkeane: I believe in this case we need it because the things that are being documented are functions and not methods.

R/cube-residuals.R
jonkeane: These shouldn't be vectors, so I've added in `length == 1` here (and added tests), in which case duplicated / out of order doesn't matter.

jonkeane:
Ok, I've just pushed the minor change and merged master and hashed out all of the style changes. I'm inclined to merge this now and follow up with the TODOs once the subsetting enhancements are merged in.

PR #308, opened 2018-08-16
Add 'weight' argument to crunchBox and to its API request

nealrichardson:
(and rebuild some docs?)


nealrichardson:
Currently, the API accepts but ignores the "weight" argument (see https://www.pivotaltracker.com/story/show/158676228). In a next step, the server will actually respond to the weight that is provided (https://www.pivotaltracker.com/story/show/158675720), and finally the server will stop magically applying a weight if one is not specified (https://www.pivotaltracker.com/story/show/158676482). So, the situation you describe where an invalid variable is sent as weight will be resolved very soon on the server.

PR #311, opened 2018-08-22
Look for a Warning response header and raise a warning

nealrichardson:


R/api.R
```diff
@@ -56,6 +56,17 @@ crDELETE <- function(...) crunchAPI("DELETE", ...)
 #' @importFrom httr content http_status
 #' @keywords internal
 handleAPIresponse <- function(response, special.statuses = list()) {
+    deprecation_warning <- get_header("Warning", response$headers)
```
jonkeane: It doesn't appear that there are any conflicts currently but I wonder if this should check for a more specific string in the message (and possibly change the backend to include that more specific message), eg `deprecated`. Alternatively, we might rely on the 299 status code, though digging around seems to suggest that's a very generic status. That way we could still use a `Warning` header for other kinds of non-deprecation warnings we might need (one that comes to mind is: https://github.com/Crunch-io/design/pull/90/files#diff-dceb1a4226338bbed6775ab26d42d369R51)

PR #312, opened 2018-08-24
Allow conditional transforms with LHSes and RHSes that don't have Crunch expressions

jonkeane:
I've loosened the validation, and added some tests. I had to do a bit more work to get the dataset object (it now checks the RHSes, LHSes, and data arugments — one of those must have at least one dataset reference in it — I doubt anyone will run into this because doing `conditionalTransform(TRUE~"foo", name = "my new all foo variable")` is a lot more work than `ds$all_foo <- "foo"` but in case someone finds themselves there).

### DIFF
--- a/tests/testthat/test-conditional-transform.R
+++ b/tests/testthat/test-conditional-transform.R
@@ -13,9 +13,28 @@ with_mock_crunch({
                 " .*formulas.* argument, or through .*....*"
             )
         )
+               
+        expect_error(
+            conditionalTransform(TRUE ~ "foo"),
+            paste0(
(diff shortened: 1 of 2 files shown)


tests/testthat/test-conditional-transform.R
```diff
@@ -13,9 +13,28 @@ with_mock_crunch({
                 " .*formulas.* argument, or through .*....*"
             )
         )
+               
+        expect_error(
```
GShotwell: Do we maybe need an integration test to ensure that the new type of transform goes through okay?

PR #314, opened 2018-08-28
Release cleanup

jonkeane:
Compress cube fixture storage


hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+#
+# An example hook script to verify what is about to be committed.
```
nealrichardson: Update this comment--looks like it's left over from a template.

PR #320, opened 2018-09-07
Decks

GShotwell:
This is a fair chunk of code letting people interact with decks. The initial idea behind this is to get some basic functionality into the package and improve it going forward, but there turned out to be more work than anticipated to get that basic functionality. Decks have six levels of objects: 

- Deck Catalogs 
- CrunchDecks
- Slide Catalogs
- Crunch Slides
- AnalysisCatalogs
- Analyses

While all of these are implemented in this PR, the user will mostly only interact with Deck Catalogs, Decks, and Slides. Things we've decided to put off: 
- Copying decks between datasets
- Extract formula queries from analyses
- `displaySettings` helpers (we want to provide things which tell people what display settings they can change and what kind of values they should provide)


R/decks.R
```diff
@@ -0,0 +1,517 @@
...
+              payload <- value@body[c("title", "subtitle")]
+              payload$analyses <- analysesToQueryList(analyses(value))
+              payload <- wrapEntity(body = payload)
+              crPOST(self(x), body = toJSON(payload))
+
+              if (i < n_slides) {
+                  # You can't modify the contents of a slide by patching it
+                  # so we need to add the new slide, delete the original slide,
+                  # and reorder the slideCatalog.
+                  new_order <- moveLastElement(seq_len(n_slides + 1), i)
```
GShotwell: Is there a better way to accomplish this? The only other method I could think of was updating the title and subtitle of the slide, then pulling the analysis catalog and updating that, but that seems even clunkier.

R/decks.R
```diff
@@ -0,0 +1,517 @@
...
+})
+
+setMethod("cubes", "AnalysisCatalog", function(x) {
+    lapply(seq_along(x@index), function(i) cube(x[[i]]))
+})
+
+setMethod("displaySettings", "AnalysisCatalog", function(x){
+    analyses <- lapply(seq_along(length(x)), function(i) x[[i]])
+    if (length(x) > 1) {
+        warning("Slide has multiple analyses, returning display settings for the first analysis")
```
GShotwell: Should this return a list of displaySettings? I think that's probably the most consistent, but since multi-analysis slides are pretty rare I'm not sure if it's more confusing than beneficial.

R/decks.R
```diff
@@ -0,0 +1,482 @@
...
+    }
+    index <- which(matches)
+    if (sum(matches) > 1) {
+        warning(dQuote(i), " does not uniquely identify elements. Returning the first match")
+        index <- index[1]
+    }
+    getEntity(x, index, CrunchDeck, ...)
+})
+
+setMethod("show", "DeckCatalog", function(object){
```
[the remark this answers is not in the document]
GShotwell: The `as.data.frame` method works on deck catalogs, I think the reason I defined a different show method was just to not print the `creation_time` and `owner_id` fields which take up a lot of screen real estate. But maybe it's better to just print everything instead of doing that. What do you think?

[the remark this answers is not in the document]
nealrichardson: I don't have a strong opinion either way. For some other options/patterns, see https://github.com/Crunch-io/rcrunch/blob/master/R/show.R (and search for "Catalog").

R/slides.R
```diff
@@ -0,0 +1,331 @@
...
+                     display_settings = list(),
+                     title = "",
+                     subtitle = "",
+                     ...) {
+    ds <- loadDataset(datasetReference(deck))
+    if (inherits(query, "formula")) {
+        query <- list(query)
+    }
+
+    settings <- modifyList(display_settings, DEFAULT_DISPLAY_SETTINGS)
```
jonkeane: Are the arguments to `modifyList` flipped here? Wouldn't this make the defaults always be used no matter if they were specified above?

```
> modifyList(list(a=1), list(a=4, b=2))
$a
[1] 4

$b
[1] 2
```

R/slides.R
```diff
@@ -0,0 +1,331 @@
...
+    AnalysisCatalog(crGET(shojiURL(x, "catalogs", "analyses")))
+})
+
+setMethod("analysis", "CrunchSlide", function (x) {
+    out <- AnalysisCatalog(crGET(shojiURL(x, "catalogs", "analyses")))
+    return(out[[1]])
+})
+
+setMethod("analysis<-", c("CrunchSlide", "formula"), function(x, value){
+    analyses[[1]] <- value
```
jonkeane: Do we need to return (invisibly) `x` here?

R/slides.R
```diff
@@ -0,0 +1,331 @@
...
+    analysis <- analyses(x)[[1]]
+    query(analysis) <- value
+    return(invisible(x))
+})
+
+setMethod("cubes", "CrunchSlide", function(x) cubes(analyses(x)))
+setMethod("cube", "CrunchSlide", function(x) cube(analyses(x)[[1]]))
+setMethod("displaySettings", "CrunchSlide", function(x) displaySettings(analyses(x)))
+setMethod("displaySettings<-", "CrunchSlide", function(x, value) {
+    displaySettings(analyses(x)) <- value
```
jonkeane: Do we need to return (invisibly) `x` here?

R/slides.R
```diff
@@ -0,0 +1,331 @@
...
+})
+
+# AnalysisCatalog --------------------------------------------------------------
+
+setMethod("[[", "AnalysisCatalog", function (x, i, ...) {
+  getEntity(x, i, Analysis)
```
jonkeane: Obviously very minor: this indent looks like two spaces?

PR #326, opened 2018-09-26
Cube subset enhancements

GShotwell:
This PR improves cube setting by supporting two additional features:

1. Allows users to reorder cubes by subsetting, i.e. `cube[c(2,1), ]` will reorder the first dimension and put the second index before the first. 
2. Support character subsetting

The big challenge for this PR was to figure out how to properly account for hidden or invisible categories when reordering. I decided to do this by setting `useNA` to `"always"`, translate the index that the user supplied, then subset the full cube. I think this works better than the previous mechanism because it separates the hidden category translation from the user-cube to real-cube translation process.


R/cube-subset.R
```diff
@@ -103,6 +128,34 @@ setMethod("[", "CrunchCube", function(x, i, j, ..., drop = TRUE) {
...
+        halt("Index is not unique. Cube subetting is only supported for unique indices.")
+    }
+
+    if (is.character(idx)) {
+        visible_categories <- cat_names[visible]
+        not_categories <- !(idx %in% visible_categories)
+        if (any(not_categories)) {
+            halt("Invalid categories: ", serialPaste(idx[not_categories]))
+        }
+        return(match(idx, visible_categories))
```
jonkeane: Although it _shouldn't_ be the case there could be edge cases where there is more than one of the same name (currently insertions+category names aren't required to be unique). I wonder if `which(idx %in% visible_categories)` would be safer here in case we run into that particular edge.

R/cube-subset.R
```diff
@@ -195,40 +242,89 @@ translateCubeIndex <- function(x, subset, drop) {
...
-                out <- rep(FALSE, length(visible))
-                out[visible][sub] <- rep(TRUE, length(sub))
+                out <- translateHidden(idx, missing, drop, vis)
             }
             return(out)
         },
-        visible = visible_cats, sub = index, SIMPLIFY = FALSE
+        missing = missing_list, idx = index, vis = visible, SIMPLIFY = FALSE
     )
 }
```
jonkeane: Are we missing `return(out)` here?

tests/testthat/test-cube-subset.R
```diff
@@ -129,11 +185,7 @@ test_that("translateCubeIndex", {
         translateCubeIndex(cat_x_mr_x_mr, alist(1:2, 1, 2), drop = FALSE),
         alist(1:2, 1, TRUE, 2, TRUE)
     )
-    # MR selection entries are set to index 1 when the indicator is dropped
-    expect_identical(
-        translateCubeIndex(cat_x_mr_x_mr, alist(1:2, 1, 2), drop = TRUE),
-        alist(1:2, 1, "mr_select_drop", 2, "mr_select_drop")
-    )
```
jonkeane: Was this test deletion intentional?

tests/testthat/test-cube-transforms.R
```diff
@@ -884,26 +884,27 @@ test_that("subtotals after cube subsetting", {
...
+            22, 22, 0
         ),
         dims = list(
             "feelings" =
                 c(
                     "extremely happy", "somewhat happy", "happy", "neutral",
                     "somewhat unhappy", "extremely unhappy", "unhappy"
                 ),
-            "animals" = c("dogs", "felines", "both")
+            "animals" = c("dogs", "both", "felines")
```
jonkeane: This change looks good: before we must not have been dropping the category appropriately so the `felines` insertion was being collated ahead of the `both` insertion, even though by the rules of ordering all anchored-to-non-existant insertions should be at the bottom.

```
$animals
An object of class  “Transforms” 
$insertions
  anchor    name     func    args
1      1 felines subtotal       1
2      2    both subtotal 1 and 2
```

and:
```
> categories(variables(pet_feeling_both)[[2]])
  id    name value missing
1  1    cats     1   FALSE
2  2    dogs     2   FALSE
3 -1 No Data    NA    TRUE
```

tests/testthat/test-cube-subset.R
```diff
@@ -299,6 +352,27 @@ test_that("[ method for cat array cubes", {
...
+        as.array(catarray_x_mr)[c(2, 1), , ]
+        )
+    expect_equal(
+        as.array(catarray_x_mr[, , c(3, 1, 2)]),
+        as.array(catarray_x_mr)[, , c(3, 1, 2)]
+        )
+    expect_equal(
+        as.array(catarray_x_mr[, 5:2, ]),
+        as.array(catarray_x_mr)[, 5:2, ]
+        )
```
jonkeane: Since these tests ought to be cheap, do you think we should add at least one with subsetting two dimensions at the same time? It would also be very good to test the useNA behavior is doing what we expect it to, and not accidentally leaking `usaNA = "always"`.

tests/testthat/test-cube-transforms.R
GShotwell: Yeah I think it was caused by an error of mine in using the `nargs() = 2` to identify cases where only `i` was supplied as an index and not `j` or `...`, this worked in most cases but of course if you provided `cube[1, drop = FALSE]` you were sending three arguments, but only one index.

tests/testthat/test-cube-subset.R
GShotwell: It was but on reflection I think it might still be useful so I put it back.

tests/testthat/test-cube-subset.R
GShotwell: I added a few more tests including some, I think the useNA behaviour is probably tested sufficiently with all the other cube-subset tests, but I added a few just in case.

R/cube-subset.R
GShotwell: Whoops, sorry about that

PR #327, opened 2018-09-26
Delete code on the other side of the feature flag for project folders

nealrichardson:
* Delete logic that checks for the `old_projects_order` feature flag
* Delete the temporary `mv`/`mkdir` methods written to work with project dataset orders
* Make the `ordering<-` method error for projects
* Remove the old `CrunchProject` and `ProjectCatalog` classes and methods. They're all `ProjectFolder`s now
* Merge the "alt" test fixtures for project folders with the existing fixtures in `inst/`
* Update tests; delete those for the temporary mv/mkdir methods

We've done the migration and flipped the flag in production, so it's not possible for this code to get hit anymore. That said, there's no rush to delete.


R/get-datasets.R
```diff
@@ -169,6 +169,33 @@ loadDatasetFromURL <- function(url) {
     return(dataset)
 }
 
+datasetURLFromPath <- function(path) {
```
nealrichardson: This should be removed entirely, but I've left it (moved here from a file that was otherwise deleted) for the moment.

PR #328, opened 2018-09-26
Start deprecating variable order methods

nealrichardson:


tests/testthat/test-variable-order.R
```diff
@@ -712,134 +667,35 @@ with_mock_crunch({
...
+                    "        Gender",
+                    sep = "\n"
+                ),
+                fixed = TRUE
+            )
         )
     })
+
     test_that("intersect_entities", {
         expect_prints(intersect_entities(ord, ds[c("birthyr", "starttime")]),
```
GShotwell: This is an example of where the vdiffr package might be helpful because it makes it easier to verify that the current printing expectations are correct.

PR #330, opened 2018-09-28
Optimize large CrunchCube insertions

jonkeane:
Persephone reported that insertions on large cubes were taking quite a long time, even if there was just one dimension with one insertion. It turns out this is largely due to remapping what insertions needed to go where for each 1d segment of the cube that the insertions were calculated for, as well as some other metadata overhead. This PR alters how insertions are calculated to instead, for each dimension where insertions are extant, create a list of closures that will, given the input 1d segment, return the output 1d segment with insertions. This way we figure out how to calculate each insertion once rather than as many times as there are extents in the other dimension.

On a real 4d cube with one insertion we go from: 
```
   user  system elapsed 
 29.091   0.935  30.145 
```
to: 
```
   user  system elapsed 
  1.252   0.088   1.336 
```


R/cube-transforms.R
```diff
@@ -187,6 +198,86 @@ applyTransforms <- function(x,
 }
 
 
+makeInsertionFunctions <- function(var_cats, transforms, cats_in_array = NULL, ...) {
+    ### Insertions
+    # collate insertions with categories for rearranging and calculation purposes
+    # we need the categories to know what order the the cube cells should be in
+    # and where to insert insertions (as they are `anchor`ed) to a category id.
+    dots <- list(...)
+    if (!"include" %in% names(dots)) {
```
GShotwell: Nitpicky but can we swap the conditionals here to avoid the `!`?  I.e. 
```
if ("include" %in% names(dots)){
    includes <- dots$include
}...
```
Alternately maybe some brackets around the `%in%` statement, I always  find negated `%in% statements a bit difficult to read.

jonkeane:
I used profvis pretty extensively which helped a lot (profvis rating: 10/10; would do again).

Although these are not super helpful because they aren't interactive, here is the before:
image

and the after:
image

It's possible there might be places for more optimization (the one that stands out is the check that the category referenced is in the array), but the returns would be relatively minor even for a large cube like this.

PR #334, opened 2018-10-16
Remove the ability to send multiple analyses to NewSlide

GShotwell:
The way we were creating multi-analysis slides was resulting in an unpredictable analysis catalog order. Since this functionality is not that useful to most users we decided to take it out.


jonkeane:
LGTM this can merge whenever.

PR #336, opened 2018-10-24
Teach R how to set the magic streaming bits

jonkeane:


R/dataset-stream.R
```diff
@@ -51,3 +51,37 @@ appendStream <- function(ds) {
...
+setGeneric("streaming<-", function(x, value) standardGeneric("streaming<-"))
+
+#' @rdname streaming
+#' @export
+setMethod("streaming", "CrunchDataset", function(x) x@body$streaming)
+
+
+#' @rdname streaming
+#' @export
+setMethod("streaming<-", "CrunchDataset", function(x, value) {
```
GShotwell: Is there a reason we're not using the match.call pattern here? Does that not work with setters?

nealrichardson: Using `match.call` would be more straightforward to do if this were a function instead of an S4 method. I'd recommend that, in fact: S4 machinery is unnecessary when we only are defining a method for one class. You could make a function and `stopifnot(is.dataset(x))`. There are some functions like this in the package.

PR #338, opened 2018-11-02
teach filters about teams

jonkeane:


tests/testthat/test-filters.R
```diff
@@ -71,6 +72,24 @@ with_mock_crunch({
...
+        )
+        
+        # can also just use a url
+        expect_PATCH(
+            team(f_2) <- "https://app.crunch.io/api/teams/team1/",
+            "https://app.crunch.io/api/datasets/1/filters/filter2/",
+            '{"team":"https://app.crunch.io/api/teams/team1/"}'
+        )
+    })
+
```
nealrichardson: * Add a test that no request is made if you set with the same thing
* Add a test of unsetting the team (sending NULL)
* Probably should have an integration test creating a filter, setting a team on it, and then setting NULL to remove it. I doubt we have e2e test coverage for this behavior anywhere else.

jonkeane: Done, done, and done. (Also 😂to "I doubt we have e2e test coverage for this behavior anywhere else.")

PR #340, opened 2018-11-04
Teach multitables to be shared by teams

jonkeane:
A bit of a refactor as well to keep the `team` methods together (for documentation purposes)


R/teams.R
```diff
@@ -95,3 +95,68 @@ setMethod("delete", "CrunchTeam", function(x, ...) {
...
+#' to see what teams you are on.
+#' 
+#' @param x a `CrunchFilter` or `Multitable`
+#' @param value a `CrunchTeam` or url for a Crunch team
+#' 
+#' @return a `CrunchTeam` that the asset is shared with.
+#' 
+#' @rdname team-sharing
+#' @export
+setGeneric("team", function(x) standardGeneric("team"))
```
GShotwell: These might not need to be methods, we could instead do something like
```
team <- function(x) {
    stopifnot(class(x) %in% c("CrunchFilter", "Multitable"))
   ....
}
```

This would both remove some code and also make it easier to add new assets to the team structure once they get added.

jonkeane: I'm inclined to leave them as functions since we want the dispatch based on the type of `value` (I know we could add an if/else inside the function, but I think this is a bit more elegant). I have collapsed them into single helper functions which saves a few characters.

PR #342, opened 2018-11-14
Cube calculations

GShotwell:
This creates a new object type `CrunchCubeCalculation` which we need to make tibbles out of the output of prop.table and margin.table. The intention is that we can create additional calculations as needed and hopefully they should just work. Most of the diff here is just calling `as.array` wherever we currently call `prop.table` or `margin.table` is called in the tests.


nealrichardson:
If it helps, feel free to make an `expect_cube` or whatever expectation function that wraps the `as.array` stuff.

PR #346, opened 2018-11-29
Benchmarks for cube transform computation

jonkeane:


jonkeane:
Should make these more traditional benchmarks where they are run a few times and then the median times are used.

PR #352, opened 2019-01-03
Pairwise comparisons

malecki:


R/cube-comparisons.R
```diff
@@ -0,0 +1,299 @@
...
+    
+    # ensure that the extents given are in the cube
+    # TODO: remove when we can defer this check to the subsetting methods
+    not_in_cube <- !(c(baseline, x) %in% names)
+    if (any(not_in_cube)) {
+        stop(c(baseline, x)[not_in_cube], " is not a column or row in the cube")
+    }
+    
+    # ensure that there are not MRs on the comparison direction
+    dim_types <- crunch:::getDimTypes(cube)
```
nealrichardson: Don't `:::` within the same package

R/cube-comparisons.R
```diff
@@ -0,0 +1,299 @@
...
+#' \dontrun{
+#' # TODO: include Hirotsu occupation-illness example cube and output
+#' illness_occupation <- crunch_example_data(hirotsu)
+#' # P-value for joint null hypothesis that each
+#' # column is equal to each other column.
+#' 
+#' wishartPvalues(illness_occupation)
+#' }
+wishartPvalues <- function(cube, dim=c("cols", "rows")){
+    X2 <- wishartMatrix(cube, dim)
```
nealrichardson: If rootWishart is `Suggests`, don't import the functions, call them with `rootWishart::fun()`, and also check that `rootWishart` is installed (see https://github.com/Crunch-io/rcrunch/blob/master/R/misc.R#L288)

tests/testthat/test-cube-pairwise.R
```diff
@@ -0,0 +1,166 @@
...
+    expect_equal(out, out_t)
+})
+
+test_that("a matrix where each row and column is the same", {
+    referenceChisq <- structure(rep(0, 16),
+                           .Dim = c(4L, 4L),
+                           .Dimnames = list(c("a", "b", "c", "d"), c("a", "b", "c", "d")))
+    referencePvals <- structure(rep(1, 16),
+                                .Dim = c(4L, 4L),
+                                .Dimnames = list(c("a", "b", "c", "d"), c("a", "b", "c", "d")))
```
jonkeane: These could be re-written as
`cubify(0, dims = list(c("a", "b", "c", "d"), c("a", "b", "c", "d")))`
and 
`cubify(1, dims = list(c("a", "b", "c", "d"), c("a", "b", "c", "d")))`
respectively (and most or all other `structure(...)` calls could be as well)

tests/testthat/test-cube-residuals-zed-scores.R
```diff
@@ -182,74 +182,55 @@ test_that("residuals for catarray", {
...
-    )
+    expected_chisq = structure(5.74120651376478, .Names = "X-squared")
     expect_equal(
         compareCols(
             gender_x_ideology,
             baseline = "Very liberal",
             x = "Very Conservative"
         ),
-        expected_zScores
+        expected_chisq
```
jonkeane: Am I reading this right that we're changing the behavior of `compareCols()` from returning z-scores for each cell (like `zScores()` does) to returning the return value from `chisq.test`? Is there any reason not to include `stdres` in the allowed `value` parameters so that it can return what we have been already returning? 

My suspicion from talking to a few of the offices about this is that the p-value derived from those z-scores are what people are thinking they want when they compare one column against another (and the same in the pair-wise compare all columns case).

R/cube-comparisons.R
```diff
@@ -0,0 +1,306 @@
...
+#' for the p-value of the test of the pair of columns/rows.
+#'
+#' @return A symmetric square matrix of all column-comparison or row-comparison
+#' \eqn{\chi^2} statistics. Typical element \eqn{i,j} is the test statistic equivalent
+#' to \code{chisq.test} subsetting to just columns i and j (for dim="cols").
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' some_cube <- crunch_example_data(cat_by_cat)
```
jonkeane:
```suggestion
#' some_cube <- crtabs(~ educ + gender, ds)
```

R/cube-comparisons.R
```diff
@@ -0,0 +1,306 @@
...
+    )
+    return(simplify2array(out))
+}
+
+#' Matrix of Chi-Squared Statistics for all rows or columns
+#' 
+#' Use the alternative Wishart method of forming the matrix of column- or row-wise
+#' comparison Chi-squared test statistics for a categorical-by-categorical 
+#' contingency table.
+#'
```
jonkeane: To add:

The null hypothesis is that all of the rows (or columns) are equal to each other. The test statistic matrix that is returned when requested is a measure of closeness between the pair of rows (or columns) given by their names. The p-value matrix that is returned are similarly the probabilities of finding the observed or more extreme results while the null hypothesis is true for each pair of rows (or columns).

R/cube-comparisons.R
```diff
@@ -0,0 +1,306 @@
...
+    names(out) <- names
+    
+    return(out)
+}
+
+#' Matrix of Chi-Squared Statistics for all rows or columns
+#' 
+#' Generate a matrix of pairwise comparisons of rows or columns, each against 
+#' the others.
+#'
```
jonkeane: To add:

The null hypothesis is that for each pair of rows (or columns) those two specific rows (or columns) in the pair are equal to each other. The test statistic matrix that is returned when requested is a measure of closeness between the pair of rows (or columns) given by their names. The p-value matrix that is returned are similarly the probabilities of finding the observed or more extreme results while the null hypothesis is true for each pair of rows (or columns).

tests/testthat/test-cube-residuals-zed-scores.R
jonkeane: Also, if we are transitioning away from z-scores to just providing the test statistic we should match that behavior elsewhere. I'm not certain we actually want to make that transition, however.

tests/testthat/test-cube-pairwise.R
malecki: *should* they? is `cubify` preferred for developer readability? As only an occasional contributor I wasn’t clear on what `cubify` did and `structure` was clearer to me ”this is the reference object”. Can change if you like.

jonkeane: I find cubify to be a bit more readable, and it saves arguments — it's not _necessary_ of course but those structures stood out to me and I had to copy/paste them and run them to confirm they were doing what they were doing.

It's not relevant for these two, but for the ones that are not a single value it also would be nice if the line breaks matched the rows breaks to make it easier to read. I found myself having to mentally find the diagonal and shift forward/back from there to see where I was in the matrix.

PR #356, opened 2019-01-08
Rework user-agent string creation and export set_crunch_config()

nealrichardson:


R/api.R
```diff
@@ -184,19 +191,45 @@ set_crunch_config <- function(cfg = c(
...
+        # And any extra bits provided
+        ...
     )
-    ua <- paste0(names(versions), "/", versions, collapse = " ")
-    if (!missing(x)) ua <- paste(ua, x)
-    return(ua)
+    return(paste(ua, collapse = " "))
+}
+
+ua_packages <- c(
```
jonkeane: Do we think we will need to mock/change this? If no (and if there's no other reason) it would be a tiny bit easier to read and not hold in memory / remember where that thing is defined if it were included inside of `crunch_user_agent` above.

nealrichardson: I put it outside because one way you could extend this list would be by `trace()`ing `crunch_user_agent()` and doing like `ua_packages <- c(ua_packages, crnewpackage="crunewpackage")`. But that's a bit much, really. So if you'd rather it be inside the `crunch_user_agent` function, that's fine.

PR #358, opened 2019-01-20
Don't rely on `margin.table`'s class attributes when calculating `prop.table(margin=NULL)`

jonkeane:


nealrichardson:
If this was caused by a report that the print method was printing undesired things, should there be a test for print? Or is asserting CrunchCubeCalculation class sufficient because there are tests elsewhere for its print method?

PR #361, opened 2019-02-15
Triple-S metadata support

jonkeane:
Support for Triple-S metadata imports


tests/testthat/test-new-dataset.R
```diff
@@ -98,6 +98,21 @@ with_mock_crunch({
             '{"element":"shoji:entity","body":{"name":"helper.R"}}'
         )
     })
+    test_that("newDataset with a schema posts to sources", {
+        expect_POST(
+            newDataset(x = "helper.R", schema = "helper.R"),
+            "https://app.crunch.io/api/sources/",
+            'list\\(uploaded_file = list\\(path = .*helper.R',
```
nealrichardson: Cool. I was going to warn you that this assertion might not be platform independent (https://github.com/nealrichardson/httptest/blob/master/tests/testthat/test-mock-api.R#L103), but seems you've found a way!

PR #362, opened 2019-02-20
Pptx deck export

jonkeane:


tests/testthat/test-decks.R
```diff
@@ -113,6 +113,10 @@ with_mock_crunch({
             exportDeck(main_deck, format = "xlsx"),
             "https://app.crunch.io/api/datasets/1/decks/8ad8/export/"
         )
+        expect_POST(
```
nealrichardson: For completeness, you should also wrap this in `expect_header(..., "Accept: application/vnd.openxmlformats-officedocument.presentationml.presentation")`

PR #364, opened 2019-02-26
Use travis to build on windows

jonkeane:
This implementation should be temporary — it looks like https://github.com/travis-ci/travis-build/pull/1538 would let us just specify windows, but until then, this could be more reliable than appveyor


.travis.yml
```diff
@@ -1,14 +1,58 @@
...
+    env:
+      global: 
+        - _R_CHECK_LENGTH_1_LOGIC2_=true
+        - secure: BWvMPmOiTzFNfFz3Ok+WHk6FjNNjK8qV5QohBhS0bgsMgeCrx4Zu3IXp42cXaEzzCe/VEnnEyYYTkZROS3mdLRAlOHnEHvCJMtETOODOK6EFGUFDv+C7AXZyqbdoNPrGrW5QKppHAVbTuL/A9z+tnBaeyG68GhOmOFBk4mciHNc=
+  - os: linux
+    r: release
+    env:
+      global: 
+        - _R_CHECK_LENGTH_1_LOGIC2_=true
+        - secure: BWvMPmOiTzFNfFz3Ok+WHk6FjNNjK8qV5QohBhS0bgsMgeCrx4Zu3IXp42cXaEzzCe/VEnnEyYYTkZROS3mdLRAlOHnEHvCJMtETOODOK6EFGUFDv+C7AXZyqbdoNPrGrW5QKppHAVbTuL/A9z+tnBaeyG68GhOmOFBk4mciHNc=
```
nealrichardson: This secure token should only be needed for the `release` version because it's used to build the pkgdown site and only on "release": https://github.com/Crunch-io/rcrunch/blob/master/pkgdown.sh#L5

.travis.yml
```diff
@@ -1,14 +1,58 @@
...
+  - os: windows
+    language: shell
+    env:
+      global: _R_CHECK_LENGTH_1_LOGIC2_=true
+    before_install:
+      - choco install -y r.project
```
nealrichardson: Can this all be shoved into a shell script, just so the cargo cult can be contained out of sight?

.travis.yml
```diff
@@ -1,14 +1,58 @@
...
+      - export R_VERSION=`ls 'C:\Program Files\R\'`
+      - export PATH=$PATH:';C:\Program Files\R\'$R_VERSION'\bin\x64'
+      - echo 'options(repos = c("https://cloud.r-project.org"))' > ~/.Rprofile.site
```
nealrichardson: Not that I'm keen on debugging this code, but if we're going to have it, let's make it unoffensive to us. Like, no `c()` around a single string, and put a space after the `;` in the `Rscript.exe -e` lines.

This also looks a lot like what our Makefile does; idk if their windows hosts have make, but if so, maybe we could call that instead.

PR #369, opened 2019-03-22
Hidden variable folder

nealrichardson:
This PR updates how we manage hidden variables, preferring the "hidden folder" over the "discarded" attribute in most places. It's a necessary step toward deprecating the discarded attribute, and it's the model we'll want to follow to implement "secure" variables.


tests/testthat/test-as-data-frame.R
```diff
@@ -144,7 +144,6 @@ with_mock_crunch({
     test_that("as.data.frame() works with hidden variables", {
         new_ds <- loadDataset("test ds")
         new_ds$gender@tuple[["discarded"]] <- TRUE
```
jonkeane: Is this discarded attribute being driven by membership in the hidden folder on the backend now? I _assume_ so, but I wanted to make sure to know that this was actually flexing what we needed for this test.

nealrichardson: This test hack still works because the identification of variables to include in the `select` is based off of the local copy of the variables catalog, which includes hidden variables and indicates them as hidden with the legacy "discarded" property. `new_ds$gender@tuple[["discarded"]] <- TRUE` doesn't actually move the variable to be hidden, and since `hiddenVariables()` checks the hidden folder now, that's why I deleted the assertion that follows. But as.data.frame/exportDataset still uses the "discarded" attrs to decide which variables to request.

PR #370, opened 2019-03-28
API now requires/ensures No Data categories/reasons

aminusfu:


tests/testthat/test-update-with-missing.R
```diff
@@ -142,10 +142,8 @@ with_test_authentication({
             ),
             name = "All B"
         )
-        expect_equal(ids(categories(ds$cat)), 1)
-        ds$cat[5] <- NA
         expect_equal(ids(categories(ds$cat)), c(1, -1))
-        expect_equal(as.vector(ds$cat[3:7], mode = "id"), c(1, 1, -1, 1, 1))
+        expect_equal(as.vector(ds$cat[3:7], mode = "id"), c(1, 1, 1, 1, 1))
```
nealrichardson: Consider deleting these tests entirely. Presumably you have tests on the server side where variables are created without No Data explicitly and it gets added automatically, so these would be redundant. They were testing behavior on the R side to make sure that No Data is added when `NA` is set, and that's not being exercised anymore.

tests/testthat/test-derive-array.R
```diff
@@ -185,7 +185,7 @@ with_test_authentication({
         ## No Data is added to the categories here
         expect_identical(
             names(categories(ds$metapetloc)),
-            c("Cat", "Dog", "Bird", "Skipped", "Not Asked", "one", "two", "No Data")
+            c("Cat", "Dog", "Bird", "Skipped", "Not Asked", "No Data", "one", "two")
```
jonkeane: Is the change in the order because before "No Data" was being added, but only after the two arrays were merged?

tests/testthat/test-interact.R
```diff
@@ -91,12 +91,16 @@ with_test_authentication({
...
+            c(
+                apply(
+                    expand.grid(
+                        c("Cat", "Dog", "Bird", "Skipped", "Not Asked", "No Data"),
+                        c("Argentina", "Australia", "Austria", "Belgium", "Brazil", "No Data")
+                    ),
+                    1, paste,
+                    collapse = ":"
+                ),
+                "No Data"
```
jonkeane: Is the `c()` wrapping necessary here?

jonkeane: Oh I see now what's going on, nevermind.

tests/testthat/test-update-with-missing.R
jonkeane: ☝️

tests/testthat/test-derive-array.R
aminusfu: Correct.

PR #375, opened 2019-05-07
Upgrade tests to work with master or zoom require-no-data-164939686

aminusfu:


tests/testthat/helper-expectations.R
```diff
@@ -52,8 +52,27 @@ expect_valid_apidocs_import <- function(ds) {
...
+    )
+}
+
+expect_equal_temp_nodata <- function(actual, expected) {
+    # Newer versions of the Crunch API will automatically add a "No Data" category if not present.
+    # This helper transitions us to that future, and can be removed when all API nodes
+    # have moved to the new behavior.
+    expect_true(
+        isTRUE(all.equal(actual, expected)) ||
+        isTRUE(all.equal(actual, expected[expected != -1]))
```
jonkeane: Did we _just so happen_ to only use `expect_equal` when testing IDs and `expect_equivalent` when testing names?

tests/testthat/test-cube-nway.R
```diff
@@ -70,12 +70,13 @@ with_test_authentication({
                     3, 4,
                     3, 6,
                     4, 3,
-                    5, 1
+                    5, 1,
+                    0, 0
                 ),
-                dim = c(2L, 5L),
+                dim = c(2L, 6L),
```
jonkeane: Won't this still be `5L` in the current world?

tests/testthat/test-cubes.R
```diff
@@ -218,20 +218,40 @@ with_test_authentication({
...
+                    0, 0
+                ),
+                dim = c(2L, 4L),
+                dimnames = list(
+                    v8 = c("1955-11-05", "1955-11-06"),
+                    v7 = c(LETTERS[3:5], "No Data")
+                )
+                ),
+                check.attributes = FALSE
+            ))
```
jonkeane: Would it be too much to have the eventually-will-be-true test alongside here so we don't have to swim through git history to remember how we constructed the `expect_equivalent` (and that we used `equivalent`). Even just commenting in the `expect_equivalent(...` inline with the non-legacy test here would be 💯

tests/testthat/test-interact.R
```diff
@@ -79,25 +79,53 @@ with_test_authentication({
...
+                    missing = TRUE
+                )
+            ),
+            error = function(e) expect_json_equivalent(
+                test_cases[[10]],
+                list(
+                    expression = ds$q1 == "Not Asked" & ds$country == "Australia",
+                    name = "Not Asked:Australia",
+                    missing = TRUE
+                )
```
jonkeane: This isn't absolutely necessary since this works, but a slightly more straight-forward way to do this would be `if`/`else` conditioned on `length(test_cases) == 11`. It reads a bit more of the intention there.

tests/testthat/helper-expectations.R
aminusfu: Mostly, yes. It doesn't apply universally.

tests/testthat/test-cube-nway.R
aminusfu: No; for this test I decided to explicitly add the "No Data" category to the fixtutre (see line 5).

tests/testthat/test-cubes.R
aminusfu: I added comments to all the complicated ones describing what to do when the other ticket goes.

tests/testthat/test-interact.R
```diff
@@ -79,25 +79,56 @@ with_test_authentication({
                 missing = FALSE
             )
         )
-        expect_json_equivalent(
-            test_cases[[10]],
-            list(
-                expression = ds$q1 == "Not Asked" & ds$country == "Australia",
-                name = "Not Asked:Australia",
-                missing = TRUE
+        tryCatch(
```
jonkeane:
```suggestion
        # Replace this `tryCatch()`
        # construction with the tried bit once the "default values"
        # ticket https://www.pivotaltracker.com/story/show/164939686 is released.
        tryCatch(
```

PR #377, opened 2019-05-21
Remove calls to the defunct "release" endpoint.

aminusfu:


tests/testthat/helper.R
```diff
@@ -52,11 +52,6 @@ newDatasetFromFixture <- function(filename) {
     )))
 }
 
-releaseAndReload <- function(dataset) {
-    .releaseDataset(dataset)
-    return(refresh(dataset))
-}
-
```
jonkeane: If any of the tests fail, you might need to replace calls to `releaseAndReload` with `refresh`.

aminusfu: None fail.

PR #378, opened 2019-05-29
Batches Deprecations API changes

amol-:


tests/testthat/test-appending.R
```diff
@@ -77,14 +77,4 @@ with_test_authentication({
...
-
-    try(crDELETE(urls(batches(out))[2]))
-    out <- refresh(out)
-    test_that("deleting a batch drops its rows", {
-        expect_true(is.dataset(out))
-        expect_length(batches(out), 2)
-        expect_identical(dim(out), dim(df))
-        expect_identical(categories(out$v4), cats)
-        expect_equivalent(as.vector(out$v3), df$v3)
-    })
```
amol-: It is no longer possible to DELETE a batch that successfully appended.
That's fine because our users were never deleting batch that appended, only deleting those that failed, which is still possible.

PR #383, opened 2019-06-26
also filter out secure variables

malecki:


malecki:
This mainly needs tests updated to be aware of the secure flag on variable tuples. This is *not* reading them from the secure folder. N.B. the secure flag is readonly.

PR #385, opened 2019-10-04
Consider that folder trees don't revert on dataset restore

jjdelc:


tests/testthat/helper-expectations.R
```diff
@@ -41,6 +41,39 @@ expect_valid_df_import <- function(ds) {
     expect_identical(names(versions(ds)), "initial import")
 }
 
+expect_valid_df_revert <- function(ds) {
```
jjdelc: This is almost a copy of `expect_valid_df_import` but unlike that, we cannot assume that the hierarchy state is the same as when originally imported. So I had to remove that assertion and also modify the name matching because they can be on a different order (`names()` reads from var order)

PR #395, opened 2020-02-04
Hidden variables only use folder structure

gergness:
I think we're good to go now! This has been a meandering prcoess, so let's review some time tomorrow @malecki


gergness:
superseded by #462

PR #397, opened 2020-02-13
some more examples for deck slides

malecki:


R/slides.R
```diff
@@ -194,7 +194,12 @@ DEFAULT_DISPLAY_SETTINGS <- list(
 #' @param deck A Crunch Deck
 #' @param query A formula definition of a query to be used by the slide. This is
 #' similar to CrunchCube query
-#' @param display_settings (optional) A list of display settings, if om
+#' @param display_settings (optional) A list of display settings. If omitted,
+#' slide will be a table of column percentages with hypothesis test highlighting
+#' enabled. These plot types that can be set here as `vizType`:
```
gergness: I'd replace 
"These plot types that can be set here as `vizType`:

with:
"The most common setting used is `vizType`, which can be:"

PR #407, opened 2020-03-09
Fix R-dev stringsAsFactors changes

gergness:
Fix for upcoming changes to stringsAsFactors argument.


tests/testthat/test-as-data-frame.R
```diff
@@ -14,8 +14,8 @@ ca.values <- data.frame(
     levels = c("A", "B"), class = "factor"
     ),
     subvar3 = structure(c(
-        1L, NA, 1L, 2L, NA, 1L, 2L, NA, 1L, NA, 2L, 1L, 1L, 2L,
-        1L, 1L, 2L, 1L, NA, 1L, 2L, 1L, 1L, 1L, 2L
```
gergness: add comment explaining why a failure with no B's would happen.

PR #409, opened 2020-03-12
Add possibility to run tests avoiding ssl self-signed certificates against local environments.

raulcd:


R/api.R
```diff
@@ -16,7 +16,8 @@ crunchAPI <- function(http.verb, url, config = list(), status.handlers = list(),
         if (!is.null(payload)) try(cat("\n", payload, "\n"), silent = TRUE)
     }
     FUN <- get(http.verb, envir = asNamespace("httpcache"))
-    x <- FUN(url, ..., config = c(get_crunch_config(), config))
+    no_verify_ssl <- getOption("crunch.verify_ssl")=="FALSE"
+    x <- FUN(url, ..., config = c(get_crunch_config(), config, config(ssl_verifyhost = !no_verify_ssl, ssl_verifypeer=!no_verify_ssl)))
```
malecki: Would these be better set in `set_crunch_config` based on the env var, and then they'd be picked up here automatically? Is there any reason to prefer this way or that?

gergness: I think my preference is to add:

```
if (crunch::envOrOption("test.verify.ssl") == FALSE) {
    crunch::set_crunch_config(c(ssl_verifyhost = FALSE, ssl_verifypeer=FALSE), update = TRUE)
}
```

here: https://github.com/Crunch-io/rcrunch/blob/master/inst/crunch-test.R#L4

This way it's more similar to the other test options that are set

raulcd: @gergness let me know if you want me to follow/change anything or test something locally. Right now I'll leave this to you unless you ask me to do something if that's ok with you.

gergness: Yep sounds good @raulcd, Mike and I will take it from here (though we may ask you to confirm our changes continue to work for you, I have a vagrant set up on my machine, so don't need this).

Thank you!

raulcd: Sounds like a plan, let me know once you want me to test and I'll try it out!

gergness: @raulcd - can you try with this branch now?

raulcd: hi @gergness it fails with the following error:
```
$ R_TEST_VERIFY_SSL=FALSE R_TEST_USER=captain@crunch.io R_TEST_PW=asdfasdf R_TEST_API=https://local.crunch.io:28443/api/ make test INTEGRATION=TRUE file=variable-summary
................
Error in f(init, x[[i]]) : is.request(y) is not TRUE
Calls: system.time ... <Anonymous> -> c -> c.request -> Reduce -> f -> stopifnot
Timing stopped at: 1.546 0.046 1.836
Execution halted
make: *** [Makefile:9: test] Error 1
```
if I go to the previous commit:
```
$ git checkout HEAD~1
$ R_TEST_VERIFY_SSL=FALSE R_TEST_USER=captain@crunch.io R_TEST_PW=asdfasdf R_TEST_API=https://local.crunch.io:28443/api/ make test INTEGRATION=TRUE file=variable-summary
........
Variable summaries: ........................................................

══ DONE ════════════════════════════════════════════════════════════════════════
Total teardown: Time difference of 0.8389423 secs
   user  system elapsed 
  4.289   0.110  14.301 
```

gergness: Sorry @raulcd - can you try again? There may be errors about slide titles now, but these are because this branch is stale

raulcd: no worries @gergness .
It seems is not really picking up the changes now no matter whether I use `R_TEST_VERIFY_SSL=FALSE` or `R_TEST_VERIFY_SSL=TRUE` it fails with the certificate failure:
```
`$ R_TEST_VERIFY_SSL=FALSE R_TEST_USER=captain@crunch.io R_TEST_PW=asdfasdf R_TEST_API=https://local.crunch.io:28443/api/ make test INTEGRATION=TRUE file=variable-summary
...
Error in curl::curl_fetch_memory(url, handle = handle) : 
  SSL certificate problem: self signed certificate
Error: Test failed: 'There are no test code errors'
* `stop(e$message)` threw an error.
Message: SSL certificate problem: self signed certificate
Class:   simpleError/error/condition
Backtrace:
 1. testthat::expect_error(stop(e$message), NA)
 2. testthat:::quasi_capture(...)
 5. rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
Timing stopped at: 1.094 0.043 1.154
Execution halted
make: *** [Makefile:9: test] Error 1
```
You can see the last local commit is the correct one:
```
 $ git log -1
commit 4eb4cd8e4ceb93f88c42ee4d55aff6c1610299e8 (HEAD -> enable-local-https-tests, origin/enable-local-https-tests)
Author: Greg Freedman Ellis <greg@crunch.io>
Date:   Mon Mar 16 11:07:03 2020 -0500

    needs to be a httr config object
```

gergness: Ugh, okay. Pretty sure I got it now. Thanks for being patient

raulcd: Ey! Worked now :)

PR #418, opened 2020-03-19
Add filter argument to tabbook

gergness:
Adds ability to set override active filter on tabbook export.

~As I pushed, realized we might want to allow CrunchExpr in the filter argument, worth it?~
No, because CrunchExprs didn't work in old code.

### DIFF
--- a/R/tab-book.R
+++ b/R/tab-book.R
@@ -18,7 +18,7 @@
 #' @param file character local filename to write to. A default filename will be
 #' generated from the `multitable`'s name if one is not supplied and the
 #' "xlsx" format is requested. Not required for "json" format export.
-#' @param filter a Crunch `filter` object or a vector of names 
+#' @param filter a Crunch `filter` object or a vector of names
 #' of \code{\link{filters}} defined in the dataset.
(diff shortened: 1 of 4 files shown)


gergness:
add link to tabbook API Docs

R/tab-book.R
```diff
@@ -29,31 +34,52 @@
...
             ## Generate a reasonable filename in the current working dir
-            file <- paste(name(multitable), f, sep = ".")
+            file <- paste(name(multitable), fmt, sep = ".")
         }
     }
 
     if (!is.null(weight)) {
         weight <- self(weight)
     }
+    flt <- filter
```
malecki:
```suggestion
```

R/tab-book.R
```diff
@@ -29,31 +34,52 @@
...
+    if (!is.null(filter) & all(is.character(filter))) {
```
malecki: should probably raise user errors here for some other conditions

R/tab-book.R
malecki: remnant of debug

gergness:
Awesome, thank you for finishing!

PR #431, opened 2020-04-01
Tabbook new endpoint fixes #420

### LINKED ISSUE #420: should tabook move to the export endpoint?
Also testing is light on tabbook

gergness:
This includes #425 because I messed up the git. 

The tests check the shape of the request now (you had added logic about filters, I added selected variables, options in the new endpoint, and shape + warning of deprecated endpoint).

### DIFF
--- a/R/tab-book.R
+++ b/R/tab-book.R
@@ -13,16 +13,23 @@
 #' @param weight a CrunchVariable that has been designated as a potential
 #' weight variable for `dataset`, or `NULL` for unweighted results.
 #' Default is the currently applied [`weight`].
-#' @param format character export format: currently supported values are "json"
+#' @param output_format character export format: currently supported values are "json"
 #' (default) and "xlsx".
 #' @param file character local filename to write to. A default filename will be
 #' generated from the `multitable`'s name if one is not supplied and the
 #' "xlsx" format is requested. Not required for "json" format export.
 #' @param filter a Crunch `filter` object or a vector of names
 #' of \code{\link{filters}} defined in the dataset.
+#' @param use_legacy_endpoint Logical, indicating whether to use a 'legacy'
+#' endpoint for compatibility (this endpoint will be removed in the future).
+#' Defaults to `FALSE`, but can be set in the function, or with the environment
+#' variable `R_USE_LEGACY_TABBOOK_ENDPOINT` or R option
+#' `use.legacy.tabbook.endpoint`.
 #' @param ... Additional "options" passed to the tab book POST request.
 #' More details can be found
 #' [in the crunch API documentation](
+#' https://docs.crunch.io/endpoint-reference/endpoint-multitable.html#options)
+#' or [for the legacy endpoint](
 #' https://docs.crunch.io/endpoint-reference/endpoint-tabbook.html#options)
 #' @return If "json" format is requested, the function returns an object of
 #' class `TabBookResult`, containing a list of `MultitableResult`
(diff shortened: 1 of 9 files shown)


R/tab-book.R
```diff
@@ -121,6 +121,35 @@ extToContentType <- function(ext) {
     return(mapping[[ext]])
 }
 
+# Possibly went a little overboard allowing different filter options in tabbook
+# extract out the logic here
+standardize_tabbok_filter <- function(dataset, filter) {
```
malecki, line 126:
```suggestion
standardize_tabbook_filter <- function(dataset, filter) {
```
Great idea extracting this!
Definitely the right way to go but to get the real benefit means some mildly tedious test rewriting (it's still the right thing to do, since we rushed the first version of it out):
all of the test cases should be taken out of `tabBook` — probably wrap them in `toJSON` and just move those json bits over to new unit tests.

R/tab-book.R
```diff
@@ -58,39 +77,25 @@ tabBook <- function(multitable, dataset, weight = crunch::weight(dataset),
...
-    if (is.CrunchExpr(expr_filter)) {
-        expr_filter <- list(c(zcl(expr_filter), name = formatExpression(expr_filter)))
-    }
-
-    if(length(filter) > 0 && !is.null(expr_filter)) {
-        filter <- unname(c(filter, expr_filter))
-    } else if (!is.null(expr_filter)) {
-        filter <- expr_filter
-    }
+    filter <- standardize_tabbok_filter(dataset, filter)
```
malecki, line 80:
```suggestion
    filter <- standardize_tabbook_filter(dataset, filter)
```

tests/testthat/test-multitables.R
```diff
@@ -363,16 +374,59 @@ with_mock_crunch({
...
+                    '{\"filter\":null,\"weight\":null,\"options\":[]}'
+                ),
+                "The legacy tabbook endpoint has been deprecated and will be removed in the future."
+            )
+        })
+    })
 
     with_POST("https://app.crunch.io/api/datasets/1/multitables/apidocs-tabbook/", {
-        book <- tabBook(mults[[1]], data = ds, format = "json")
+        book <- tabBook(mults[[1]], data = ds, output_format = "json")
```
malecki, line 429: shouldn’t we leave one of these intact to ensure that `format="json"` doesn’t break? (and ensure coverage for the branch of code that checks for it?)

tests/testthat/test-multitables.R
```diff
@@ -363,16 +374,59 @@ with_mock_crunch({
...
+        expect_POST(
+            tabBook(mults[[1]],
+                    data = ds, output_format = "json", format = list(pval_colors = TRUE))
+            ,
+            "https://app.crunch.io/api/datasets/1/multitables/ed30c4/export/",
+            '{\"filter\":null,\"weight\":null,\"options\":{"format":{"pval_colors":true}}}'
+        )
+    })
+
+    test_that("tabBook warning when using format argument", {
```
gergness, line 400: @malecki test for deprecation warning for `format="json"`. Currently only on new endpoint, mostly because I wasn't sure how testthat handles two warnings (and there will be one for old endpoint and another for old argument). Do you think it's worth adding that case?

tests/testthat/test-multitables.R
gergness: that was a separate test, see above

malecki: oops, thanks

tests/testthat/test-multitables.R
malecki: nah. Let’s merge it and then fix https://github.com/Crunch-io/crunchtabs/blob/master/R/tabBooks.R#L11 to use `output_format`
cc @1beb — this should be mostly transparent to you; expect deprecation warnings in crunchtabs and a PR to fix them. Maybe we should consider pinning an rcrunch version in https://github.com/Crunch-io/crunchtabs/blob/master/DESCRIPTION#L19?

1beb:
@malecki @gergness I'm still collecting examples from internal teams. I'll have more tomorrow. Can we hold off on this until tomorrow afternoon?

tests/testthat/test-multitables.R
1beb: I am indeed getting a tonne of the deprecation warnings but everything appears to be working fine for PDF outputs. I've tried about 5 different clients stuff so far and no failures. I still need to test ExcelTabs because that is used heavily by YouGov Blue. I have examples collated so the remaining step is a bit of comparative code and we should be good to go.

1beb: To confirm is it only the `format` argument that needs to be changed to `output_format`. We are working with two different branches so I will need to update both. Has the option stayed the same "json"?

gergness: Yep, `"json"` and `"xlsx"` are still the two options, but we had to rename the argument to avoid a conflict with options in the new endpoint.

That's the only thing you should have to change.

PR #439, opened 2020-04-20
adding Nexus R proxy

gjergjidojce:


.travis.win.install.sh
```diff
@@ -3,7 +3,7 @@
 choco install -y r.project &&
 export R_VERSION=`ls 'C:\Program Files\R\'` &&
 export PATH=$PATH:';C:\Program Files\R\'$R_VERSION'\bin\x64' &&
-echo 'options(repos = "https://cloud.r-project.org", install.packages.compile.from.source = "never")' > ~/.Rprofile.site &&
+echo 'options(repos = "https://ui.nexus.crint.net/repository/rcrunch/", install.packages.compile.from.source = "never")' > ~/.Rprofile.site &&
```
gergness, line 6: As discussed on slack, let's not use nexus here, because it only affects windows and we want windows & linux to match on travis

PR #442, opened 2020-04-23
add weight & filter methods to crunchdecks

gergness:
I'm only checking the first request, do you think that's enough?

I think I would need to do some complicated error recovery mechanism to check multiple requests.


NEWS.md
```diff
@@ -1,4 +1,5 @@
 # crunch 1.27.5 (Development version)
+* You can set the `weight` and `filter` to be the same on each slide of a deck.
```
malecki, line 2:
```suggestion
* You can set the weight and filter to the same value for every slide in a deck by assigning a `weight` or `filter` on the deck object.
```

malecki (commented):
suggestion, is this clearer for news?

PR #445, opened 2020-05-05
Hidden+Private v3

gergness:
Supersedes #412 (which was painful to rebase)

Detection of hidden variables now is entirely from folder structure of hidden variable folder.

Also adds ability to interact with "private" (/"secure") variables with an analagous API to hidden.


gergness:
included in #462

PR #446, opened 2020-05-05
Use header based authentication instead of cookies

gergness:
PR that supersedes #428 because it uses the same branch as is being tracked in pivotal


gergness:
included in #462

PR #452, opened 2020-05-11
Gh actions

gergness:
trying out github actions


gergness:
TODO: 
- [ ] Get codecov working
- [ ] Add action to update pkgdown on tag (requires encrypting a token with access to tada)

PR #453, opened 2020-05-19
[WIP] Add derived subvar

gergness:
Definitely not ready for primetime yet, but wanted to get some notes down.

Weird things:
- `absolutifyVariables` (which modifies the derivation expression before user sees it) is wrong in some cases, likely having to do with array variables. It seems to assume that the variable is given by relative URL, but within `map` zcl function, variables are given by ID only. I have gotten around this by not converting to absolute paths, but I'd like to actually fix.
- For subvariables that are created when the array is first derived, they get their metadata automatically filled in, but we don't get that when adding new ones via derivation. Is it possible to get this? Or a limitation of the API?
- Haven't gotten the MRs to work yet because they have an extra layer of nesting in zcl. Need a better way to do the derivation modification that fails more gracefully and is flexible for at least these 2 cases.

```r
library(crunch)
set.seed(2020-05-18)

levels <- c("Good", "Okay", "Bad")
data <- data.frame(
    x1 = factor(sample(c(levels, NA), 10, TRUE), levels),
    x2 = factor(sample(c(levels, NA), 10, TRUE), levels),
    x3 = factor(sample(c(levels, NA), 10, TRUE), levels)
)

ds <- newDataset(data, name = "combine into array test")
ds$xcat <- deriveArray(ds[1:2], name = "x cat")
ds$xmr <- deriveArray(ds[1:2], name = "x mr", selections = "Good")


# Works in POC
ds$xcat <- addSubvariable(ds$xcat, ds["x3"])

# Does not yet work 
ds$xmr <- addSubvariable(ds$xmr, ds["x3"])
```

### DIFF
--- a/R/add-subvariable.R
+++ b/R/add-subvariable.R
@@ -14,24 +14,64 @@
 #' @export
 addSubvariable <- function(variable, subvariable) {
     stopifnot(is.Array(variable))
-    new.urls <- addSubvarDef(variable, subvariable)
+    if (!is.derived(variable)) {
+        new.urls <- addSubvarDef(variable, subvariable)
 
-    ## Store these for post workaround
-    subvar.urls <- subvariableURLs(tuple(variable))
+        ## Store these for post workaround
+        subvar.urls <- subvariableURLs(tuple(variable))
 
-    ## Do the adding
-    crPATCH(shojiURL(variable, "catalogs", "subvariables"),
-        body = toJSON(sapply(new.urls, emptyObject, simplify = FALSE))
-    )
+        ## Do the adding
+        crPATCH(shojiURL(variable, "catalogs", "subvariables"),
+                body = toJSON(sapply(new.urls, emptyObject, simplify = FALSE))
+        )
 
-    ## Workaround because apparently bind/rebind isn't retaining the order
-    crPATCH(self(variable),
-        body = toJSON(list(subvariables = I(c(subvar.urls, new.urls))))
-    )
+        ## Workaround because apparently bind/rebind isn't retaining the order
+        crPATCH(self(variable),
+                body = toJSON(list(subvariables = I(c(subvar.urls, new.urls))))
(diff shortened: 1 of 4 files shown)


R/add-subvariable.R
```diff
@@ -67,3 +75,128 @@ addSubvarDef <- function(var, subvar) {
...
+    } else if (isSelectCatDerivation(old_deriv)) {
+        new_deriv <- addToSelectCatDerivation(old_deriv, subvariable, categories(variable))
+    } else {
+        halt("Could not add subvariable because did not recognize variable derivation structure")
+    }
+
+    derivation(variable) <- new_deriv
+    # We don't get metadata from original variable like we would if we were creating
+    # subvariable during original derivation...
+    # TODO: It would be nice to update name in the same POST as updating derivation
```
gergness, line 102: I think a derivation can have a `subreferences` field? Is that the best way?

R/add-subvariable.R
```diff
@@ -67,3 +75,128 @@ addSubvarDef <- function(var, subvar) {
...
+    # to ensure consistency, but I don't see how yet
+    # TODO: Also, the alias is really ugly for newly created subvariables
```
gergness, line 104: Is there a way to get the server's naming logic? I tried just adding the alias explicitly, hoping that the server would dedup for me, but this resulted in a failure.

tests/testthat/test-add-subvariable.R
```diff
@@ -91,3 +145,61 @@ with_test_authentication({
         )
     })
 })
+
+with_test_authentication({
+    # adding to derived array
+    # TODO: Use existing fixture? Add this to fixtures?
```
gergness, line 151: This simple example works well for these tests, but we usually use `newDatasetFromFixture()`, worth changing to that (either by using existing fixture or adding this one there)?

PR #461, opened 2020-06-09
keyring take 2

gergness:
Sorry @mainwaringb I think making that keyring branch screwed up github, so I couldn't push my commits anymore. I've made a new PR over here with my changes.


gergness:
supersedes #448

gergness:
Thanks @mainwaringb !

PR #463, opened 2020-06-14
Keyring

mainwaringb:
Completed tests. Also made minor changes to `with_mock_auth_settings` functions: making sure the fake `key_list` function mirrored the behaviour of the real function, and making sure the `test_with` command ran correctly.

Worth noting an ssue around CI on Linux/Secret Service that I found mentioned in the [Keyring developer notes](https://github.com/r-lib/keyring/blob/master/inst/development-notes.md#testing-the-package). I'm hoping that this wasn't the root of the initial problem, although I think some of the Windows tests also had issues:

> libsecret uses libglib, and libglib does not allow unloading the package. More precisely, if you unload the keyring package on Linux, then some libglib threads will stay around. If you then reload the package, R will crash.
> 
> This limitation is especially annoying for development with devtools, because devtools::load_all(), devtools::test(), etc. try to unload and reload the package, which results in crashes. A workaround is to run the tests from the command line:
> 
> R -e 'devtools::test()'
> 

We should be aware that the mock test won't catch any change to the data format returned by `key_get`. It seems exceptionally unlikely that the RStudio team would intentionally change this. However, it is conceivable that future OS updates might change how password storage is handled, in a way that requires keyring to make changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ matrix:
       - _R_CHECK_TESTS_NLINES_=0
     apt_packages:
        - libsodium-dev
+    before_script:
+       - Rscript -e 'devtools::install_github("rstudio/rmarkdown")'
   - os: linux
     r: oldrel
     env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ matrix:
     env:
       - _R_CHECK_LENGTH_1_LOGIC2_=true
       - _R_CHECK_TESTS_NLINES_=0
+    apt_packages:
+       - libsodium-dev
     before_script:
        - Rscript -e 'devtools::install_github("rstudio/rmarkdown")'
   - os: linux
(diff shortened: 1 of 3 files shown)


gergness:
Awesome!

Thanks, I don't think that was the problem, but didn't investigate it a ton. Also good point on the fact that we won't catch changes to keyring.

I think these tradeoffs are worthwhile, though, because the rcrunch tests need to run in a lot of different environments (On Travis which is publicly visible, on a Jenkins instance that the Crunch team supports (so that we can test rcrunch against development versions of the main Crunch service) and locally for not just R developers, but Crunch backend developers who we don't want to bother with a ton of setup, because usually they are making small infrequent changes when they modify behavior of the server.

mainwaringb:
It looks like the remaining issue is with Linux libsodium on Travis, are we good delete this from your perspective?

mainwaringb:
@gergness Just bumping this! Not sure if the Travis issues have been fixed, or if Travis is down - however, hoping for the former.

gergness:
Hey, so sorry! The only thing blocking is that I need to make sure this works with crunch's internal CI system which is completely invisible to you and not something I know a ton about and so have been procrastinating

PR #479, opened 2020-08-10
case when integration tests

gergness:
add integration tests for case when

waiting for https://www.pivotaltracker.com/n/projects/2172644/stories/174198825 to ship


gergness:
API has been fixed so this can be merged now

malecki:
I think we really should have a realistic / user-understandable example here that we could point a user to, use in doc, perhaps share with Automation team

PR #482, opened 2020-08-14
refactored text from expropriate to reassign

1mikegrn:


NEWS.md
```diff
@@ -224,7 +224,7 @@ and multiple response variables.
 # crunch 1.18.4
 * All catalog objects now have an `as.data.frame()` method.
 * The list of dataset "weight variables" can now be set with `modifyWeightVariables()`, `weightVariables(ds) <- ds$newWeight` or `is.weightVariables(ds$var) <- TRUE`
-* Users with account admin privileges can now `expropriateUser()` to transfer datasets, projects, and other objects owned by one user to another, as when that user has left your organization.
```
malecki, line 227: editing past news isn't a great way to communicate this change. It should receive a new entry at the top of the file (current version) so we can inform users that the function has changed names.

The few users who were aware of and used this function may be confused by its sudden disappearance; we might even want to keep the old function name and have it throw an error to inform them.

R/user.R
```diff
@@ -94,30 +94,30 @@ setMethod("[[", c("UserCatalog", "character"), function(x, i, ...) {
...
-#' Expropriate all Crunch objects from a user
+#' Reassign all Crunch objects from a user
 #'
 #' If you want to transfer all teams, projects, and datasets owned by one user
-#' to another you can with `expropriateUser`. To have permission to use
-#' `expropriateUser` you must be an account admin and be from the same account
-#' as the user who is being expropriated. This is useful if a user leaves your
+#' to another you can with `reassignUser`. To have permission to use
+#' `reassignUser` you must be an account admin and be from the same account
+#' as the user who is being reassignd. This is useful if a user leaves your
```
malecki, line 102: typo

PR #492, opened 2020-08-24
fix: branch build for numeric arrays

slobodan-ilic:
* When creating arrays (by binding) from already exisging variables, we
  used to rely on the fact that we _only_ have categorical arrays. This
  has now changed, and if we create an array from numerical variables,
  it will result in a numerical array. Therefore, if we want to create a
  categorical array, we have to manually cast the constituent variables
  to be of type "categorical", prior to creating the array.


slobodan-ilic:
Close as irrelevant, since we're not requiring the explicit cast for each binding of numeric variables to categorical array.

PR #494, opened 2020-08-31
work on httptest, especially for child packages

### LINKED ISSUE #450: revisit httptest
- learn from https://ropensci.org/technotes/2019/12/11/http-testing/ ?

- Make `gsub_request()` more targeted (had problem with crunchtabs because it trimmed a UUID from a POST)

gergness:
Closes #450 

- Adds exported `with_api_fixture()` for child packages to use when testing
- Adds a request modifier for httptest. I believe this solves the biggest pain point @1beb and I have had, where the request during a test doesn't match the captured request. I've tested in an example package that with this new function I can use the full link of a dataset (with the full dataset ID). There's a decent chance @1beb will need to rerun his fixtures to get it working after this though.
- Adds a vignette with my accumulated knowledge of testing child packages from seeing `crunchtabs` (would love feedback from @1beb)


vignettes/test-children.Rmd
```diff
@@ -0,0 +1,165 @@
...
+3) Either creating the dataset in the script, or at least starting from 
+   a dataset in a project stored in a place that everyone who works on
+   the package have access to.
+
+Putting that altogether, here's an example of a script. It could be saved
+for example in a file `"tests/testthat/captured/generate_1.R"`.
+```{r, eval = FALSE}
+library(crunch)
+library(httptest)
+
```
1beb, line 78: Let's add an explicit `rm(list=ls())` here. I've found weird stuff happens if I don't clear my workspace appropriately. We should also make note that a restart of R + clearing workspace is recommended before capturing a test.

vignettes/test-children.Rmd
```diff
@@ -0,0 +1,165 @@
...
+directly to developing for rcrunch itself 
+([more details at bottom of this document](#rcrunch-specific)). 
+
+When testing code that relies on an API, there are two broad categories of testing:
+
+- **Integration testing** - The tests call to the server each time they are run. 
+  While this is nice in theory, it can cause problems because test code can
+  overload the servers because of how frequently they need to run which will not only
+  cause your tests to fail, but negatively affect all users of the API.
+- **Mock testing** - Rather than actually call the server, the tests use "mocks"
```
1beb, line 29: There is another type of testing here that's built into testthat called "with_mock" that can be extremely useful and in many cases better than trying to create fixtures. I have made some examples internally for the innovations team but I think they belong here as well. 

image

1beb: Generally as well the guidance to separate I/O from logic is a good idea. Many on our team are loading a ds than going directly to a data.frame (do not pass go ,do not collect $200). These calls are often mixed in with other logic that could be tested independently if they were indepdent of crunch.

1beb (approved):
A couple of small notes about problems that I have experienced and patterns I have seen across teams. It's good to go as is but recommend adding information about: 

1. with_mock from testthat
2. Note to restart and clear workspace before capturing tests

vignettes/test-children.Rmd
```diff
@@ -0,0 +1,181 @@
...
+
+## Mock testing strategy
+Mock tests for the crunch package are tested using the `httptest` package, which was 
+written while the author worked at Crunch.io and so is tailored to our situation quite
+well. Before continuing, I highly recommend reading the vignettes for that package:
+
+- [`httptest`: A Test Environment for HTTP Requests in R](https://enpiar.com/r/httptest/articles/httptest.html)
+- [Redacting and Modifying Recorded Requests](https://enpiar.com/r/httptest/articles/redacting.html)
+
+Another strategy for generating mock tests is to use the
```
gergness, line 49: @1beb - added a short description of `with_mock()` here. In an ideal world I think you'd use `httptest` because it gives rcrunch more flexibility, though I understand it may not feel like it is worth the hassle to you.

vignettes/test-children.Rmd
```diff
@@ -0,0 +1,181 @@
...
+   folder, at the level above where the fixtures are saved.
+2) Deleting existing mock files if they already exist.
+3) Either creating the dataset in the script, or at least starting from 
+   a dataset in a project stored in a place that everyone who works on
+   the package have access to.
+
+Putting that altogether, here's an example of a script. It could be saved
+for example in a file `"tests/testthat/captured/generate_1.R"`.
+
+Before running, it's generally best to start from a clean R session
```
gergness, line 86: @1beb - couldn't bring myself to directly recommend `rm(list = ls())`, but mention it in text here

vignettes/test-children.Rmd
1beb: I think the suggestion that httptest is any better than with_mock is inappropriate. In both cases a real change to the API means you have to update your tests/code/fixture. Both are equally divorced :safety_pin:

vignettes/test-children.Rmd
1beb: Mulligan!

vignettes/test-children.Rmd
```diff
@@ -52,9 +52,11 @@ in `testthat` itself, but will be deprecated in the next version of `testthat` a
 the time of writing this). 
 
 The `with_mock()` approach has the advantage that there are fewer moving parts to go
-wrong. However, this approach has the disadvantage that it allows your mocks to be 
-completely divorced from the crunch API and rcrunch package. Therefore it will be harder
-for your tests to get updated in the event that the behavior of rcrunch changes.
+wrong. However, following the steps in this vignette to create your mocks will allow
```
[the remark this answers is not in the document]
1beb: :+1:

1beb (approved):

PR #513, opened 2020-10-08
Another take at multiweight tabbook

gergness:
Another take at multiweight tabbooks.

I think the arguments and return value of `tabBook()` is compatible with your PR, though I have substantially changed (and renamed) the weight helper function. 

There is a continuum of differences from just stylistic preference  to actual improvements. I find it hard to concretely define them, but I'll try to focus on some changes that I made deliberately that are closer to the "improvement" side.

1) If we are going to export the weight helper function (which I think is a good idea), then it needs to be understandable on its own. Renaming the `ind` and `index` columns was a start, but there's still a keep column that doesn't mean that some rows are dropped, an `order` column that only matters if you know what the individual tabbooks look like, and a `position` column that seems like it might also mean the same thing as `order`, but doesn't. In the new implementation, the data.frame is two columns `weight` and `alias` (and the order of the data.frame determines the order). 

This is important for two reasons, because it's clearer to whoever has to work on this part of the codebase next, and also because, as I was reminded today "there are downstream dependencies here" (even for features that haven't yet passed code review). By including such implementation details in the public interface, it locks us into them, making it harder to maintain the code and preserve backwards compatibility when things change. As an example, what if the multitable export server side API changes to allow you to specify multiple weights in a single export. Then the `order` column is totally meaningless, but to help my downstream dependencies, I would need to keep it.

I also decided to be strict that there can only be these two columns, because this allows us to expand in the future. I kind of think there needs to be a "name" column that would allow you to set the name of the cubes (eg "q1 - likely voter", "q1 - registered voters"), but I'm not sure where that name can be stored in the tabbook result and it sounds like you plan to do that for yourself in crunchtabs, so there's no rush. If I allowed arbitrary columns, then using the name column like this would potentially be a breaking change.

2) All of the work about picking the weights is done in the `tabBookWeightSpec()` function now because it's much easier and quicker to test that function than the `tabBook` one. This is why I'm able to achieve (almost) complete test coverage of the new code with just one new fixture. Not only is saving fixtures a pain, but each test of `tabBookWeightSpec()` is like 10 times faster than the tabbook export, so I'm able to test more edge cases without slowing down testing (and thus development time).

3) At your suggestion, I made the mock generator script totally reproducible so that it does all of that file moving and renaming that we were having to do manually before. I'm trying to document more things.


R/tab-book.R
```diff
@@ -48,7 +51,7 @@
 #' @importFrom jsonlite fromJSON
 #' @export
 tabBook <- function(multitable, dataset, weight = crunch::weight(dataset),
-                    output_format = c("json", "xlsx"), file, filter = NULL,
+                    output_format = c("json", "xlsx"), file = NULL, filter = NULL,
```
1beb, line 54: There was some code that I'm unfamiliar with the functionality of used previously for the `file` parameter. I flipped this to NULL instead of empty (with a subsequent test for `missing()`). The effects of this are unclear to me but the code appeared equivalent.

R/tab-book.R
```diff
@@ -112,6 +129,161 @@ tabBook <- function(multitable, dataset, weight = crunch::weight(dataset),
...
+
+    # stitch together
+    # Most of the objects should be the same because they come from the same multitable
+    # But the analyses object contain an item per variable with the weight included
+    # and then the pages section contains the cube results
+    analyses <- mapply(
+        weight = weight_spec$weight,
+        page_num = weight_spec$page_num,
+        FUN = function(weight, page_num) {
+            books[[which(names(books) == weight)]]@.Data[[1]]$analyses[[page_num]]
```
1beb, line 191: This might need to be `[page_num]` in single square brackets to match the original format of a TabBookResult.

R/tab-book.R
```diff
@@ -112,6 +129,161 @@ tabBook <- function(multitable, dataset, weight = crunch::weight(dataset),
...
+        },
+        SIMPLIFY = FALSE
+    )
+    pages <- mapply(
+        weight = weight_spec$weight,
+        page_num = weight_spec$page_num,
+        FUN = function(weight, page_num) {
+            books[[which(names(books) == weight)]]@.Data[[2]][[page_num]]
```
1beb, line 199: This one stays as double square brackets :+1:

dev-misc/capture-multi-weight-tabbook-mock.R
```diff
@@ -0,0 +1,89 @@
+library(httptest)
```
1beb, line 1: This is great. Can be used for docs :+1: this process is wildly forgettable otherwise.

tests/testthat/test-multitables.R
```diff
@@ -653,22 +364,6 @@ with_test_authentication({
...
-    test_that("We can get an json tab book", {
-        skip("multitables and multiple response need more work.")
-        skip_on_local_backend("Vagrant host doesn't serve files correctly")
-        book <- tabBook(mult, data = ds)
-        expect_is(book, "TabBookResult")
-        expect_identical(dim(book), c(ncol(ds), 3L))
-        expect_identical(names(book), names(variables(ds)))
-    })
-
     test_that("team-sharing of multitables", {
```
1beb, line 367: It looks like a bunch of tests were removed here?

R/tab-book.R
```diff
@@ -62,15 +65,29 @@ tabBook <- function(multitable, dataset, weight = crunch::weight(dataset),
...
-        } else {
-            ## Generate a reasonable filename in the current working dir
-            file <- paste(name(multitable), fmt, sep = ".")
-        }
+    if (is.null(weight) | is.variable(weight)) {
+        tabBookSingle(multitable, dataset, weight, fmt, file, filter, use_legacy_endpoint, dots)
+    } else if (is.list(weight)) {
+        tabBookMulti(multitable, dataset, weight, fmt, file, filter,  use_legacy_endpoint, dots)
+    } else {
+        stop("weight must be NULL, a CrunchVariable or a list indicating a multi-weight spec")
```
1beb, line 73: Based on the updated tabWeightSpec it could also be a data.frame

1beb:
Downstream I have a functional branch based off of [ed2e1ee](https://github.com/Crunch-io/rcrunch/pull/513/commits/ed2e1eef3320b0868496aab7f57e1f55bc4e1ceb)

R/tab-book.R
gergness: Yeah, it's hard to pass missing arguments like the old API to other functions (like `tabBookMulti()`/`tabBookSingle()` so I agreed with your decision to change to `NULL`.

R/tab-book.R
gergness: Decided against this, this way it better matches single tabBooks to have double.

tests/testthat/test-multitables.R
gergness: Just moved to a file dedicated to `tabBook`

R/tab-book.R
gergness: Yeah, `is.list()` returns `TRUE` for `data.frame`s, but I added a check for data.frame so that the code is clearer

1beb:
Some 500's coming off of body$where in tabBookSingle with the example dataset.

1beb (commented):
I had to pull this into crunchtabs with some modifications so that I could deliver to my users. I'm not sure how or if we should bother trying to bring this back into `rcrunch` given the limited users (Innovations team). 

Would be happy to revisit but if this is something you want to integrate please follow the functionality embedded in crunchtabs.

PR #525, opened 2020-12-07
176040966 replace links old api docs

dcarr178:


gergness (approved):
LGTM - I fixed a few lines that crept over 100 characters and ran `make doc` to generate the .Rd files in `man/`. Gonna make sure I caught all the lint mistakes and then will merge

PR #531, opened 2021-01-15
Num array basic usage

gergness:
The bulk of the non-CrunchCube related work for Numeric Arrays. They can be created (both as derived variables and non-derived), tests had to be updated to make sure they weren't depending on coercing numerics to categorical when binding into an array. (builds off #530)


R/make-array.R
```diff
@@ -15,6 +15,10 @@
 #' @param selections character, for `makeMR` and `deriveArray` the names of the
 #' categories to mark as the dichotomous selections. Required for
 #' `makeMR`; optional for `deriveArray`; ignored in `makeArray`.
+#' @param numeric Logical indicating whether the array should be a numeric
+#' array or categorical array. `NULL` the default will guess numeric if
+#' all variables are known to be numeric and categorical if all are
+#' categorical. It cannot guess for expression subvariables.
```
gergness, line 21: `It cannot guess if the subvariables are expressions rather than true Crunch Variables` or maybe something better?

gergness (commented):
Also should there be vignette language about how to use a Numeric Array?

PR #537, opened 2021-02-01
Num array cube

gergness:
A simple way to handle numeric arrays in cubes, when the first measure can be identified as a numeric array (the measure has metadata with object type that has subvariables, matching what the cr.cube does) we inflate a single "measure axis" that comes with most measures of numeric arrays and adds it as an equal to the other cube dimensions.

I see 2 issues with this approach:
1) (probably not a big deal): This approach can get confused with complex cubes with some measures based on a numeric array and others aren't.
2) (probably will need a different strategy when as we continue to support num arrays): Not all measures based on numeric arrays have one dimension, eg covariance matrix has 2, so I think we need a separate the concepts of "measure axis" vs "cube dimension" and keep track of which is which)

### DIFF
--- a/R/cube-dims.R
+++ b/R/cube-dims.R
@@ -15,6 +15,8 @@ cubeDims <- function(cube) {
             references = tuple
         ))
     })
+    dimnames <- c(dimnames, inflateNumArray(cube))
+
     names(dimnames) <- vapply(
         dimnames, function(x) x$references$alias,
         character(1)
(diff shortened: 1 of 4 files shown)


R/cube-dims.R
```diff
@@ -82,6 +84,30 @@ elementName <- function(el) {
     return(out)
 }
 
+inflateNumArray <- function(cube) {
+    ## Numeric arrays have pseudo-dimensions stored in their results
+    ## that needs to be "inflated" up to a real dimension  here.
+    ## We'll look only at the first cube for this, because in most cases
+    ## this okay, but it does mean that cubes that mix numeric arrays and
+    ## not will have invalid dimensions.
+    first_metadata <- cube$result$measures[[1]]$metadata
```
gergness, line 93: Do I need to use defaultMeasure here?

gergness: Not available because it's not a cube yet. This approach already isn't perfect, I think this'll be okay until we implement measure axes.

PR #541, opened 2021-02-03
Scorecard query

gergness:
Can now do:

```r
slide <- newSlide(deck, ~scorecard(ds$enjoy_mr, ds$another_mr), title = "scorecard slide")
```

I continued to stretch the formula notation, at some point this may break, but for now I think it's nice because it means you can use the scorecard query other places that expect a query too (eg in `crtabs`).

Left to do:
documentation (not sure where to put it)

### DIFF
--- a/R/cube-query.R
+++ b/R/cube-query.R
@@ -1,13 +1,34 @@
 #' Crunch xtabs: Crosstab and otherwise aggregate variables in a Crunch Dataset
 #'
 #' Create a contingency table or other aggregation from cross-classifying
-#' variables in a CrunchDataset.
+#' variables in a CrunchDataset, expanding on the notation allowed in
+#' [`stats::xtabs()`] to tailor to the kinds of calculations available in crunch.
 #'
-#' @param formula an object of class 'formula' object with the
-#' cross-classifying variables separated by '+' on the right side of the
-#' "~". If aggregating by functions other than counts, include the aggregation
-#' expression on the left-hand side.
-#' Compare to [stats::xtabs()].
+#' There are 3 types of queries supported:
+#'
+#' - Crosstabs: Share the most in common with [`stats::xtabs()`], are defined by
+#' a formula with only a right hand side, with each dimension specified on the
+#' right-hand side, separated by a `+`. A dimension are generally variables, but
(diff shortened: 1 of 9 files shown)


R/cube-query.R
```diff
@@ -1,13 +1,34 @@
...
-#' expression on the left-hand side.
-#' Compare to [stats::xtabs()].
+#' There are 3 types of queries supported:
+#'
+#' - Crosstabs: Share the most in common with [`stats::xtabs()`], are defined by
+#' a formula with only a right hand side, with each dimension specified on the
+#' right-hand side, separated by a `+`. A dimension are generally variables, but
+#' categorical array variables contribute 2 dimensions, "categories" and "subvariables".
+#' If you just use the categorical array variable directly, the "subvariables" dimensions
+#' will be added first and the "categories" second, but you can choose their order by
```
malecki, line 14: Probably better? How many quotes appear in our docs? [Official guidance](https://developer.r-project.org/Rds.html) is to use this stuff, and “sparingly”

```suggestion
#' categorical array variables contribute 2 dimensions, \dQuote{categories} and 
#' \dQuote{subvariables}.
#' If you just use the categorical array variable directly, the subvariables dimensions
#' will be added first and the categories second, but you can choose their order by
```

R/cube-query.R
```diff
@@ -1,13 +1,34 @@
...
+#' measures are `mean(var)`, `n()` (the same as a crosstab), `min(var)`, `max(var)`,
+#'  `sd(var)`, `sum(var)` and `median(var)` (where `var` is a CrunchVariable).
+#'
+#' - Scorecards: When you want to compare multiple MR variables with the same
+#' subvariables, you can use a scorecard to create a tabulation where they are
+#' lined up. Scorecard queries cannot be combined with the other types. Use
+#' the `scorecard(..., vars = NULL)` (where `...` is a set of MR variables or
+#' `vars` is a list of them).
+#'
+#' @param formula an object of class 'formula' object, that specifies that query
```
malecki, line 30: maybe overkill?
```suggestion
#' @param formula a [stats::formula] object that specifies that query
```

PR #545, opened 2021-02-11
Adding option to retry for manipulating wait outside of code

1beb:


gergness:
Hm, not sure this is the right way to expose this - I worry that users will wrap eg `tabBook()` in the `retry()` which would send the POST multiple times, when really you only want to retry on the download.

Do you think we need to expose `retry()`? I can see how the `wait` option is useful, but maybe we keep `retry` private, and document `crunch_retry_option` in `?http-methods`

Would that be enough, or do you think you need to expose retry?

gergness:
Oops, I was wrong - I saw the roxygen and assumed it was exported, but it's not. Thanks!

PR #555, opened 2021-03-23
Script dryrun

gergness:
- Allows `...` in `runCrunchAutomatio()` so we can use the `dry` argument (and later `dry_run` when the API is updated)
- Allows passing multiple scripts, which are then concatenated. I think this is particularly important for dry runs because the validation would need to be cumulative.
- I couldn't resist subtracting out the line numbers for multiple scripts, we can undo it if you really want, but I don't think it's that hard to avoid the off by 1 error, and being able to jump right to errors in their original file when running `runCrunchAutomation(list.files(), dry_run = TRUE)` is a huge UX improvement.
Screen Shot 2021-03-23 at 4 18 23 PM

### DIFF
--- a/R/automation.R
+++ b/R/automation.R
@@ -231,6 +231,11 @@ crunchAutomationErrorHandler <- function(response) {
             }
         )
 
+        # Sometimes no further information is provided
+        if (length(errors) == 0) {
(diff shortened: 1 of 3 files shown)


gergness:
On hold while crunch automation changes how multi-file support works

PR #556, opened 2021-03-24
Lazy variables catalog

gergness:
Lazily load variables catalog (including hidden/private which require folder traversal).

To avoid having to read huge JSON files repeatedly, adds a new layer of cache into httpcache that stores the R object variables catalog, so I think we're avoiding most situations that would be slower than the status quo, so while I don't think it's ever slower, it is a little more unpredictable. For example, things like RStudio's autocomplete of `ds$` rely on the variables catalog, so there can be kind of a weird stutter.

It takes branch builds down from 30 to 23 minutes, so it's significant speed up for developers, and it will speed up certain situations for regular use quite a bit too (though it can be hard to avoid loading the variables catalog).


1beb:
Tested against main branch of crunchtabs. No problems to report. 
image

1beb:
Another note. There's seems to be a multiplied lag. Let's say I have a list of 15 or more datasets. After the 15th dataset we see an exponentially increasing lag occur. 

```
lapply(ds_list, names)
```

gergness:
Hm, yeah, seems like that can happen in some situations with lots of datasets. Added option `options(crunch.lazy.varcat=FALSE)` so this behavior can be changed

R/dataset.R
```diff
@@ -1,38 +1,127 @@
...
-        if (!is.null(private_dir)) {
-            private_vars <- variablesBelowFolder(private_dir, "alias")
-            .Object@privateVariables <- .Object@variables[aliases(.Object@variables) %in% private_vars] #nolint
-        }
+
+    # This is only NULL when instantiating a fresh dataset object. If
+    # subclassing an existing dataset object, the variable catalog will
+    # already be populated (and due to subsetting, may not be identical
+    # to a fresh pull from the API)
+    if (is.unforcedVarCat(.Object@variables) && !use_lazy_var_cat()) {
```
gergness, line 8: replace cat with catalog

R/dataset.R
```diff
@@ -1,38 +1,127 @@
...
+#' because there is a limit to the number of datasets your user can hold open at the
+#' same time and so at some point the server will have to unload and then reload the
+#' datasets. However, it's probably even faster if you are able to alter your code so
+#' that it operates on datasets sequentially.
+#'
+#' @param x A crunch dataset
+#'
+#' @return A dataset with it's variable catalogs filled in
+#' @export
+forceVarCat <- function(x) {
```
gergness, line 47: Spell out `forceVariableCatalog()`

PR #558, opened 2021-03-31
Download numeric array

persephonet:
numeric arrays wouldn't download in as.vector or as.data.frame so I updated the code. There aren't tests written because numeric arrays don't seem to be part of the test fixtures atm.

### DIFF
--- a/R/as-vector.R
+++ b/R/as-vector.R
@@ -135,6 +135,22 @@ parse_column <- list(
         }
         return(out)
     },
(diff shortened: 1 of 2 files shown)


gergness (approved):
thanks!

gergness:
Going to add tests and change how the `include.hidden` is fixed in a PR that builds off this

gergness:
included in #560, but a rebase made it not automatically close

Issue #306, opened 2018-08-14
Harmonise the behavior of `as.data.frame()` with respect to hidden variables


jonkeane:
When we merged of #290 it looks like we had a slight discrepancy in the behavior of `as.data.frame()` with force on.

The policy we agreed on was: 
• if `ds` is not a subset of variables, `as.data.frame(force=TRUE)` returns only visible variables unless `include.hidden=TRUE`
• if it is a subset of variables, `as.data.frame(force=TRUE)` should return whatever variables are specified in the subset, regardless of whether any are hidden, which in effect ignores `include.hidden=TRUE`

The integration test were written expecting this behavior, but when the PR above was merged, it turns out that's not the behavior that actually happened. I changed the test expectation in https://github.com/Crunch-io/rcrunch/commit/31e96991d62617a49e6cfe4feff4fa6d9a00b332 just to get the build process flowing. That change shouldn't be needed and should be reverted when the behavior above is harmonised.

I wonder if it's possible that this discrepancy was from the fact that we allowed hidden variables to have promises made for them with `as.data.frame()` sans `force = TRUE`. This behavior is fine, and probably good, since these promises are cheap and only get evaluated if someone tries to use that variable. This behavior, however, isn't a requirement as far as I know and we could stop doing it and make both do _exactly_ the same thing if that's the only way forward.

Finally, it looks like the unit tests that would have caught this were taken out — I couldn't find any discussion why they were removed, so it's possible that there's a good reason, but if at all possible we should add them back to catch this without having to run integration tests (against a non-local backend!).

gergness:
review after #412 is merged

PR #560, opened 2021-04-08
as.vector & as.data.frame improvements

### LINKED ISSUE #516: include.hidden = FALSE is ignored in as.data.frame(force = TRUE)
``` r
library(crunch)

login()
#> Logged into crunch.io as Greg@crunch.io

# fct_levels <- c("Yes", "No")
# 
# ds <- newDataset(
#     data.frame(
#         x1 = factor(c("Yes", "No", "Yes"), fct_levels), 
#         x2 = factor(c("No", "Yes", "No"), fct_levels), 
#         y_hidden = factor(c("Yes", "No", "No"), fct_levels)
#     ),
#     "test hide export"
# )
# 
# hide(ds$y_hidden)
ds <- loadDataset("test hide export")
```

``` r
as.data.frame(ds, include.hidden = FALSE, force = TRUE)
#> Warning: Variable y_hidden is hidden
#> Warning in format.data.frame(if (omit) x[seq_len(n0), , drop = FALSE] else x, :
#> corrupt data frame: columns will be truncated or padded with NAs
#>    x1  x2 y_hidden
#> 1 Yes  No     <NA>
#> 2  No Yes     <NA>
#> 3 Yes  No     <NA>
```

gergness:
- Includes #558 to allow `as.vector()` on numeric arrays
- Adds support for numeric arrays in `as.data.frame()`
- Fixes #516 & #306 by having `as.data.frame(force = TRUE)` respect `include.hidden` (for backwards compatibility, `CrunchDataFrame`s still have access to the hidden and secure variables)
- Moves tests of `as.vector` and `as.data.frame` to use the vegetables example so that they would have numeric arrays to test (which ended up leading to lots of reorganization of mocks)


mocks/app.crunch.io/api/datasets/by_name/Vegetables%20example.json
```diff
@@ -23,8 +23,11 @@
             },
             "size": {
                 "columns": 23,
+                "derivations": 3,
                 "rows": 210,
-                "unfiltered_rows": 210
+                "source_columns": 20,
+                "unfiltered_rows": 210,
+                "variables": 23
```
malecki, line 30: oops, sorry about this surprise recent change.
It’s really good to see it reflected in an updated fixture here though actually.

malecki (approved):

mocks/app.crunch.io/api/datasets/by_name/Vegetables%20example.json
gergness: Oh yeah, I see this as a major benefit of the scriptable regenerating .json files. If we needed to develop tests for a feature that depended on these, we would already have fixtures in place.

PR #566, opened 2021-04-20
Dry run only

gergness:
Pull out the dry run part out of #555 so we can merge it without the multi-file support

And add handling of async validation errors and encoding specification


R/api.R
```diff
@@ -51,11 +57,17 @@ crDELETE <- function(...) crunchAPI("DELETE", ...)
...
+#' during an async request
 #' @return The full HTTP response object, just the content, or any other
 #' status-specific action
 #' @importFrom httr content http_status
 #' @keywords internal
-handleAPIresponse <- function(response, special.statuses = list()) {
+handleAPIresponse <- function(
+    response,
+    special.statuses = list(),
+    progress.handler = NULL
```
malecki, line 69: makes sense

R/api.R
```diff
@@ -51,11 +57,17 @@ crDELETE <- function(...) crunchAPI("DELETE", ...)
 #' Do the right thing with the HTTP response
 #' @param response an httr response object
 #' @param special.statuses an optional named list of functions by status code.
+#' @param progress.handler an optional function that resolves errors raised
+#' during an async request
```
malecki, line 61:
```suggestion
#' @param progress.handler an optional function to handle errors reported by
#' a progress result. Default NULL prints the string \code{message}; other
#' functions required to handle non-string messages in progress responses.
```

R/automation.R
```diff
@@ -102,6 +102,10 @@ setMethod("scriptSavepoint", "Script", function(x) {
 #' @param is_file The default guesses whether a file or string was
 #' used in the `script` argument, but you can override the heuristics
 #' by specifying `TRUE` for a file, and `FALSE` for a string.
+#' @param encoding The encoding the script file is saved as on your
+#' machine. Defaults to UTF-8.
```
malecki, line 106:
```suggestion
#' @param encoding Optional encoding to convert \strong{from}: the API accepts only UTF-8.
```

gergness: I did these locally so that I could re-roxygen. Thanks!

PR #586, opened 2022-01-11
Api token 180398987

gergness:
New api token workflow


R/api.R
```diff
@@ -217,6 +226,22 @@ locationHeader <- function(response) {
 
 get_crunch_config <- function() getOption("crunch.httr_config")
 
+get_crunch_auth_config <- function(url) {
+    # --- Don't send token outside of api host (aws downloads fail if you try)
+    api_hostname <- parse_url_for_domain(envOrOption("crunch.api"))
```
gergness, line 231: Add check that it is always of the form `https://X.crunch.io/api/

PR #590, opened 2022-01-28
reduce if not eliminate the use of the word project

malecki:
help fix this up; probably needs more work but we want to reduce if not eliminate the word project


malecki:
see also https://github.com/Crunch-io/rcrunch/blob/master/vignettes/projects.Rmd . sorrryyyy

PR #591, opened 2022-02-15
Tests use token 181268334

gergness:
This combines a few disparate things because I was running around working on all of them today:

1) This removes all uses of `login()` from integration tests. This may likely lower test coverage, but I think it's worth it so that I don't have to rush to remove them if we want to stop supporting login.

2) Removes flaky test about string truncation and increases the time that it is willing wait for an append test that's been flaky.

3) Adds `envOrOption` to the pkgdown site's reference page


malecki (approved):
LGTM

PR #597, opened 2022-04-25
[181987759]: Error on password auth

gergness:


gergness:
Now get error `"Using the crunch API with a username and password is no longer supported. See `help('crunch-api-key')` to learn how to authenticate using an API key."` when trying to use `login`, `logout` and `resetPassword`

PR #602, opened 2022-10-17
[181049738]: copyFolders uses public folder not root

gergness:
Missed a place where we relied on the old style of public variables in the root of the variables directory instead of using the public folder.

I also inlined the logic of `rootVariableFolder()` because I don't think we ever want to use it directly so we can avoid these bugs.


malecki (approved):
LGTM

PR #615, opened 2023-03-07
[184615453]: Add ability to run Crunch Automation scripts on folders

sluga:
Previously, the runCrunchAutomation function only supported running CA scripts on datasets.

### DIFF
--- a/R/AllGenerics.R
+++ b/R/AllGenerics.R
@@ -479,7 +479,7 @@ setGeneric("palettes", function(x) standardGeneric("palettes"))
 #' @export
 setGeneric("defaultPalette", function(x, ...) standardGeneric("defaultPalette"))
 
-# for ggplot to not copmlain when given crunchdata
+# for ggplot to not complain when given crunchdata
 #' Fortify crunch objects for use with ggplot
 #'
 #' @param model model or other R object to convert to data frame
@@ -493,3 +493,5 @@ fortify.CrunchDataFrame <- function(model, data, ...) model
 #' @rdname fortify
 #' @export fortify.CrunchDataset
 fortify.CrunchDataset <- function(model, data, ...) model
+
+setGeneric("sendCrunchAutomationScript", function(x, ...) standardGeneric("sendCrunchAutomationScript"))
--- a/R/automation.R
+++ b/R/automation.R
@@ -91,24 +91,28 @@ setMethod("scriptSavepoint", "Script", function(x) {
 
 #' Run a crunch automation script
 #'
-#' Crunch automation is a custom scripting syntax that allows you to
-#' concisely describe the metadata of your data when importing. The
-#' syntax is described [in the crunch API documentation](
-#' https://help.crunch.io/hc/en-us/categories/360004356012-Crunch-Automation)
+#' Crunch Automation is a custom scripting language that allows you to
+#' execute common Crunch commands. The syntax is described in the
+#' [Crunch API documentation](
+#' https://help.crunch.io/hc/en-us/categories/360004356012-Crunch-Automation).
 #'
-#' @param dataset A crunch dataset
-#' @param script A path to a text file with crunch automation syntax
-#' or a string the syntax loaded in R.
+#' If a character vector with length bigger than 1 is passed to `script`,
+#' it's converted to a string by concatenating its elements together using
+#' line breaks.
+#'
+#' @param x A crunch dataset or project folder (for backwards compatibility,
+#' `dataset` is also accepted)
+#' @param script A path to a text file containing a Crunch Automation script
+#' or a character vector of length 1 or more with Crunch Automation commands (see `Details`)
 #' @param is_file The default guesses whether a file or string was
 #' used in the `script` argument, but you can override the heuristics
 #' by specifying `TRUE` for a file, and `FALSE` for a string.
 #' @param encoding Optional encoding to convert **from**, defaults to UTF-8.
 #' The API accepts only UTF-8, so all text will be converted to UTF-8 before
 #' being sent to the server.
-#' @param ... Additional options, such as `dry_run = TRUE` passed on
-#' to the API
-#'
-#' @return For `runCrunchAutomation()`: an updated dataset (invisibly),
+#' @param ... Additional options, such as `dry_run = TRUE`, passed on
+#' to the API if x is a dataset (if x is a project folder, an error is thrown)
+#' @return For `runCrunchAutomation()`: an updated dataset/project folder (invisibly),
 #' For `showScriptErrors()`, when run after a failure, a list with two items:
 #' `script`: that contains the script string sent to the server and `errors` which is a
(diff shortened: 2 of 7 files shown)


R/AllGenerics.R
```diff
@@ -493,3 +493,7 @@ fortify.CrunchDataFrame <- function(model, data, ...) model
 #' @rdname fortify
 #' @export fortify.CrunchDataset
 fortify.CrunchDataset <- function(model, data, ...) model
+
+setGeneric("sendCrunchAutomationRequest", function(x, ...){
```
gergness, line 497: I worry `Request` is an overloaded with meaning when it comes to web traffic. Maybe `sendCrunchAutomationScript()`?

R/automation.R
```diff
@@ -128,45 +132,108 @@ setMethod("scriptSavepoint", "Script", function(x) {
...
+            'While using `dataset` will continue to work, it will emit this warning.'
+        )
+        target <- list(x = args$dataset)
+        args$dataset <- NULL
+        dots$dataset <- NULL
+        do.call(
+            'runCrunchAutomation', c(target, args, dots),
+            quote = FALSE, envir = parent.frame()
+        )
+        return(invisible(NULL))
```
gergness, line 172: why not return result of the ultimate `runCrunchAutomation()`?

R/automation.R
```diff
@@ -128,45 +132,108 @@ setMethod("scriptSavepoint", "Script", function(x) {
...
+    }
+    script <- paste(script, collapse = '\n')
+    reset_automation_error_env()
+    automation_error_env$file                  <- filename
     automation_error_env$last_attempted_script <- script
 
+    sendCrunchAutomationRequest(x = x, script = script, ...)
+
+    # Provide feedback for dry_run success so that user is confident it was successful
+    if (isTRUE(dots$dry_run)) message("Script dry run was successful")
```
gergness, line 188: I think this precedes your code, but probably should skip the refresh for `dry_run` and just return x

R/automation.R
```diff
@@ -128,45 +132,108 @@ setMethod("scriptSavepoint", "Script", function(x) {
...
+        ...
+){
+    # project folders include a slot views with element execute,
+    # which gives us the URL to hit;
+    # but the account ('top-level folder', what you get from: `projects()`)
+    # is also of class ProjectFolder, but doesn't include this info;
+    # running CA scripts on the account is not supported currently
+    if (!is.crunchURL(x@views$execute)){
+        stop(
+            'not implemented yet'
```
gergness, line 223: Let's provide more details in the error message. Something like `"This object does not support crunch automation scripts at this time."`

Ideally it'd say "Top level project folder", but I think it'd be too much fuss to get a good description that works in all cases.


Also, I'm probably not consistent about this, but house style is to use `halt()` instead of `stop()`. Let's continue doing that even though the main feature of adding `call. = FALSE` is becoming less useful with RStudio improvements to error traces.

R/automation.R
```diff
@@ -175,6 +242,10 @@ string_is_file_like <- function(x) {
         (grepl("\\.[[:alnum:]]+$", x) || file.exists(x))
 }
 
+wrapView <- function(value, ...) {
```
gergness, line 245: Weird, almost wrote that I didn't think you'd need this but looks like scrunch has it too. Can you move this function to `shoji.R`, next to the other `wrap*` functions?

R/automation.R
```diff
@@ -128,45 +132,108 @@ setMethod("scriptSavepoint", "Script", function(x) {
...
+    # running CA scripts on the account is not supported currently
+    if (!is.crunchURL(x@views$execute)){
+        stop(
+            'not implemented yet'
+        )
+    }
+    
+    crPOST(
+        shojiURL(x, "views", "execute"),
+        body = toJSON(wrapView(value = script)),
```
gergness, line 229: I'm not sure where the `...` should be, but ideally we'd be able to have `...` passed to the API somehow.

R/automation.R
gergness: Silently ignores `...`?

R/automation.R
```diff
@@ -128,45 +132,108 @@ setMethod("scriptSavepoint", "Script", function(x) {
...
+    args      <- as.list(sys.call())
+    args[[1]] <- NULL
+    if ('dataset' %in% names(args)){
+        warning(
+            'The argument `dataset` has been renamed to `x` because the function ',
+            'now supports more generic operations. ',
+            'While using `dataset` will continue to work, it will emit this warning.'
+        )
+        target <- list(x = args$dataset)
+        args$dataset <- NULL
```
gergness, line 166: Realized I don't understand how `dots` and `args` are different, can you explain?

gergness (changes requested):
Looks good! A few comments.

I can't get myself worked up about style stuff, but we may want to restart using the `styler` package (or convince Mike that conflicting styles is okay). As you may have noticed from the github actions comments, I've let some linter warnings sneak in, and I think your white space around assignment will add even more.

I think people recommend using automatic linters because then noone has to talk about styling ever again, which sounds nice to me.

R/automation.R
sluga: OK

R/automation.R
sluga: I was basing this part on: https://help.crunch.io/hc/en-us/articles/10521594535309-Getting-started-with-folder-scripts

Where it says:

```
POST /api/projects/<project_id>/run/
{
  Element: shoji:view,
  Value: """script contents"""
}
```

So I'm also unsure where to put the extras. I'll investigate.

R/automation.R
sluga: Well, args captures everything whereas dots captures just the extra arguments. But in this code the extras get duplicated, so this is a bug. Will fix & also make sure the test covers this.

R/automation.R
gergness: Looking if scrunch does anything

R/automation.R
gergness: Okay, this may not actually matter, but please change to look more like the `wrapEntity()` function:

```
wrapView <- function(..., value = list(...)) {
    list(element = "shoji:view", value = value)
}
```

Then because you're passing the value as a named parameter already, it will still send a string, but if parameters are added and `value` is allowed to be an object instead of a string it'll be easier to support.

R/automation.R
gergness: Based in conversation in automation channel, let's throw an error if `...` are not empty on a folder script.

R/automation.R
```diff
@@ -128,45 +132,108 @@ setMethod("scriptSavepoint", "Script", function(x) {
...
+        dots$dataset <- NULL
+        do.call(
```
gergness, line 168: I can tell this is hard so may be just missing something, but wonder if the code would be cleaner if you fixed up the arguments so they were consistent, and then continued here instead of using `do.call` to go back to the beginning of `runCrunchAutomation`

PR #618, opened 2023-03-28
Use getShinyOption to access options (if ...)

sluga:
When using crunch from Shiny apps on a server, we typically don't want the users to share options, we want the options to be session-specific. shiny's shinyOptions and getShinyOption accomplish this when called from within the server function.

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -230,16 +230,22 @@ vectorOrList <- function(obj, type) {
 #' @export
 # nolint end
 envOrOption <- function(opt, default = NULL, expect_lgl = FALSE, expect_num = FALSE) {
-    ## First look in CRUNCH_OPTIONS environment
+
+    # First look in CRUNCH_OPTIONS environment
     crunch_opt <- get_crunch_opt(opt)
     if (is_crunch_opt_default(crunch_opt)) {
         return(default)
+    } else if (is.function(crunch_opt)){
+        crunch_opt <- crunch_opt()
+        if (!is.null(crunch_opt)){
+            return(crunch_opt)
+        }
     } else if (!is.null(crunch_opt)) {
         attributes(crunch_opt) <- NULL
         return(crunch_opt)
     }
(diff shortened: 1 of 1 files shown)


R/misc.R
```diff
@@ -230,7 +230,25 @@ vectorOrList <- function(obj, type) {
...
+        }
+        shiny_option <- shiny::getShinyOption(opt)
+        if (!is.null(shiny_option)){
+            return(shiny_option)
+        }
+    }
+    
+    ## look in CRUNCH_OPTIONS environment
     crunch_opt <- get_crunch_opt(opt)
     if (is_crunch_opt_default(crunch_opt)) {
```
gergness, line 253: Hm, what do you think about checking if `crunch_opt` is a function here, and then crunchy could have this logic in it. Right now this, this `__SHINY_SERVER_CONTEXT__` feels a little out of place in the rcrunch package to me.

gergness: More explicitly I'm suggesting `rcrunch` would only change so that there's something like:
```
else if (is.function(crunch_opt)) {
  crunch_opt <- crunch_opt()
  if (!is.null(crunch_opt)) return(crunch_opt)
}
```

And then during our crunchy settings we could have:

```
token_func <- function() {
    if (identical(Sys.getenv('__SHINY_SERVER_CONTEXT__'), 'TRUE')){
         if (!require(shiny)){
             stop(
                 '__SHINY_SERVER_CONTEXT__ set to TRUE, ',
                 'but the shiny package cannot be loaded'
             )
         }
         shiny_option <- shiny::getShinyOption(opt)
         if (!is.null(shiny_option)){
             return(shiny_option)
         }
     }
}

set_crunch_opt(crunch.api.key = token_func)
```

sluga: Oh yeah, your suggestion is much better, let's go with that.

gergness (approved):

PR #620, opened 2023-04-12
Modify newDatasetFromFile so that it correctly handles a JSON schema

sluga:
This way we can upload to Crunch using a parquet or csv file for the data and a JSON file for the fully specified metadata.


tests/testthat/test-new-dataset.R
```diff
@@ -93,19 +93,47 @@ with_mock_crunch({
             '{"element":"shoji:entity","body":{"name":"helper.R"}}'
         )
     })
-    test_that("newDataset with a schema posts to sources", {
+    test_that("newDataset with an .sss schema posts to sources", {
         expect_POST(
-            newDataset(x = "helper.R", schema = "helper.R"),
+            newDataset(x = "helper.R", schema = "schema.sss"),
```
gergness, line 98: I have a slight preference for doing this with a tempfile (`tempfile("schema", fileext=".sss")`) rather than making 2 new files to further clutter up the repository.

(Obvious counterpoint, there's already a ton of junk, but I think I'd rather stash it somewhere in the `inst/` folder if you think it should be permanent.)

tests/testthat/schema.json
```diff
@@ -0,0 +1 @@
...ble","metadata":{"some_var1":{"name":"Some variable 1","alias":"some_var1","description":"What do you choose?","type":"categorical","categories":[{"name":"Response 1","id":1,"numeric_value":1,"missing":false},{"name":"Response 2","id":2,"numeric_value":2,"missing":false},{"name":"Response 3","id":3,"numeric_value":3,"missing":false},{"name":"Skipped","id":8,"numeric_value":8,"missing":true},{"name":"Not Asked","id":9,"numeric_value":9,"missing":true}]},"some_var2":{"name":"Some variable 2","alias":"some_var1","description":"What do you choose now?","type":"categorical","categories":[{"name":"Response 1","id":1,"numeric_value":1,"missing":false},{"name":"Response 2","id":2,"numeric_value":2,"missing":false},{"name":"Response 3","id":3,"numeric_value":3,"missing":false},{"name":"Skipped","id":8,"numeric_value":8,"missing":true},{"name":"Not Asked","id":9,"numeric_value":9,"missing":true}]},"hidden_var":{"name":"Secret dream","alias":"hidden_var","description":"What is your secret dream?","type":"text","missing_rules":{"Not asked":{"value":""}}}},"hidden":["hidden_var"],"order":[{"group":"One variable","entities":["some_var1"]},{"group":"Another variable","entities":["some_var2"]}]}}}
```
gergness, line 1: Is it possible to use `inst/example-datasets/pets.json` for your tests? If so, I'd prefer that over a new file.

gergness (commented):
Some complaints about the new files in the testthat folder but I don't feel too strongly.

If possible, I'd like to wait until Mike gets back so I can get some background on why xml/sss schema are sent with each batch but json ones aren't (we may be assuming that sss/xml are only for spss files for example). If it's better to merge now, go for it and I can talk to him after.

sluga:
Fixed the files.

No problem waiting on Mike, we're in no hurry.

gergness:
Noticed right before merging that the windows failure was real - was a simple fix though

PR #621, opened 2023-04-17
[184790111]: Redirect Crunch Automation on root project folder

sluga:


R/automation.R
```diff
@@ -206,18 +206,21 @@ setMethod("sendCrunchAutomationScript", "ProjectFolder", function(x,
     # which gives us the URL to hit;
     # but the account ('top-level folder', what you get from: `projects()`)
     # is also of class ProjectFolder, but doesn't include this info;
-    # running CA scripts on the account is not supported currently
+    # previously, we function raised an error in this case;
+    # now, following what frontend did, we execute the script on the account
     if (!is.crunchURL(x@views$execute)) {
```
gergness, line 211: I think we should do some heuristics to only redirect when at the root and keep the error if we're not and there's no execute view (I don't know if it currently works this way, but I can imagine the API not letting you run scripts on certain folders because of permissions by not including the execute view, and we wouldn't want to silently redirect to the account in this case).

Is there an obvious way to tell? If not, we can ask Jose what the frontend does here: https://github.com/Crunch-io/whaam/pull/8006/files#diff-ed5a738f637f797570efe80b7388b58ef741de4cb66bf7615caf498b3b35978bR27

tests/testthat/app.crunch.io/api/account.json
```diff
@@ -0,0 +1,6 @@
+{
```
gergness, line 1: Yeah, this is how I was imagining getting around the error in `GET` before the `expect_POST`. It seems like it works for me, do you still get errors? (If so, did you run `make compress-fixtures`?)

PR #622, opened 2023-04-28
Stop relying on n_missing in cube results

gergness:
1) Changes how R figures out `na.rm` for the direct numeric functions like `mean`, `sd`, etc.

2) Open up the possibility of using `na.rm` in `crtabs()` like `crtabs(mean(ds$age, na.rm = FALSE) ~ 1, ds)` though don't change the default here even though this isn't consistent with base R, preserve backwards compatibility even though it means that `na.rm` defaults to `TRUE` in `crtabs()` (and decks and tabbooks).


R/univariate.R
```diff
@@ -28,11 +28,15 @@ setGeneric("median")
...
     cube <- CrunchCube(crGET(cubeURL(x), query = query))
-    if (!na.rm && cube$result$measures[[measure_name]][["n_missing"]] > 0) {
-        return(NA_real_)
-    } else {
-        return(as.array(cube))
+    out <- as.array(cube)
+    # --- Convert from NaN to NA because the zz9 doesn't distinguish
+    # --- Try to do it minimally so that if it's really NaN we preserve it
+    # --- but in reality we don't have enough information to know for sure
+    if (!na.rm && any(vapply(out, is.nan, logical(1)))) {
```
sluga, line 35: 1. Why do  we need vapply here? Instead of just:

```
if (!na.rm && any(is.nan(out))) {
    out[is.nan(out)] <- NA
}
```

... simplifying further to:

```
if (!na.rm) {
    out[is.nan(out)] <- NA
}
```

2. What do you mean by doing it 'minimally'?

gergness: Ha, I think that's python making me forget about vectorized functions 😢 

Meaning if you don't  do `na.rm` we're not setting to `NA`, so that some `NaN`s are preserved. I updated the comment.

PR #630, opened 2023-10-26
Troubleshoot new deck API endpoint

gergness:


R/api.R
```diff
@@ -50,17 +50,31 @@ crunchAPI <- function(
 crGET <- function(...) crunchAPI("GET", ...)
 #' @rdname http-methods
 #' @export
-crPUT <- function(...) crunchAPI("PUT", ...)
+crPUT <- function(url, config = list(), ..., body) {
```
gergness, line 53: arg order is weird here, but trying to match `crunchAPI()` so that I don't make any breaking changes

sluga: Does body need to be named as opposed to a part of ...?

R/api.R
```diff
@@ -39,7 +39,7 @@ crunchAPI <- function(
 #' These methods let you communicate with the Crunch API, for more background
 #' see [Crunch Internals](https://crunch.io/r/crunch/articles/crunch-internals.html).
 #'
-#' @param ... see [`crunchAPI`] for details. `url` is the first
+#' @param url,config,body,... see [`crunchAPI`] for details. `url` is the first
```
sluga, line 42: The help page now looks a bit odd (compare GET/DELETE & PUT/PATCH/POST, they all get url/config, but it's explicit only in the latter group).

R/api.R
```diff
@@ -191,22 +205,37 @@ handleAPIfailure <- function(code, response) {
...
     }
     msg <- http_status(response)$message
+    msg2 <- NULL
     if (code == 404) {
         # Add the URL that was "not found" (there isn't going to be any
         # useful response content message)
         msg2 <- response$url
     } else {
-        msg2 <- try(content(response)$message, silent = TRUE)
+        err_content <- try(content(response), silent = TRUE)
```
sluga, line 214: Are both of the checks needed in the top-level if statement? I'd think `is.list` is enough ...

Do we lose anything with a more radical simplification such as:

```
if (is.list(err_content)){
    msg2 <- coalesce(err_content$message, err_content$description)
}
```

... adding coalesce somewhere:

```
coalesce <- function(...){
    for (i in list(...)) if (!is.null(i)) return(i)
}
```

R/api.R
```diff
@@ -191,22 +205,37 @@ handleAPIfailure <- function(code, response) {
...
+                !is.null(err_content$description)
+            ) {
+                msg2 <- err_content$description
+            }
+        }
     }
-    if (!is.error(msg2)) {
+
+    if (!is.null(msg2)) {
         msg <- paste(msg, msg2, sep = ": ")
```
sluga, line 229: Random rant: I always found it annoying that paste includes the separator even with NULL. If `paste('foo', NULL, sep = ':')` returned 'foo', you wouldn't need an if statement in cases like this one.

R/api.R
gergness: This is the easiest way I know how to avoid evaluating the arguments inside `...` at the wrong time, I want to just pass them onto the next function as is, but eg `list(...)` evaluates all of them. (Technically I could use `...names()` & `...n()` to only evaluate body, but it looks uglier)

R/api.R
gergness: Cool, I'll take another look. Partly I'm trying to be as defensive as possible because an error here would mean an unintelligible error to the user

R/api.R
gergness: Made them consistent at least

R/api.R
gergness: Adapted some, but really want to be conservative here because we're by definition working with a response we don't fully understand (because we sent an error) and it makes super confusing behavior for the user if we error in our intrepration

PR #640, opened 2024-04-12
Find variables and their paths in a Crunch dataset or variable folder

sluga:
WIP for now, because I'm not sure if this belongs here. Currently it returns a data.frame because I'm not sure if existing Crunch abstractions are a great fit this function as it'd be very useful if it not only finds the variables but collects their paths as well.

### DIFF
--- a/R/findVariables.R
+++ b/R/findVariables.R
@@ -1,0 +1,57 @@
+#' Find variables and their paths in a Crunch dataset or folder
+#'
+#' @param x Crunch dataset or variable folder
+#' @param deep FALSE (the default) or TRUE; should subfolders
+#' @param include.hidden FALSE (default) or TRUE, should hidden be included in the result?
+#'
+#' @return Data.frame with one row per Crunch variable and columns \code{alias}, \code{path}, \code{hidden}
+#' @export
+findVariables <- function(x, deep = FALSE, include.hidden = FALSE) {
+    if (is.dataset(x)) {
+        x <- cd(x, ".")
+        startpath <- ""
+    } else if (is.folder(x)) {
+        startpath <- name(x)
+    } else {
+        halt('`x` should be "CrunchDataset" or "VariableFolder", not "', paste(class(x), collapse = ", "), '"')
+    }
+    if (!isTRUE(deep) && !isFALSE(deep)) {
+        halt("`deep` should be TRUE or FALSE")
+    }
+    if (!isTRUE(include.hidden) && !isFALSE(include.hidden)) {
+        halt("`hidden` should be TRUE or FALSE")
+    }
+    if (!deep) {
+        vars <- aliases(variables(x))
+        nvars <- length(vars)
+        res <- data.frame(alias = vars, path = rep(startpath, nvars), hidden = rep(FALSE, nvars))
+        return(res)
+    }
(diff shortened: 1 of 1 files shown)


R/findVariables.R
```diff
@@ -0,0 +1,57 @@
+#' Find variables and their paths in a Crunch dataset or folder
+#'
+#' @param x Crunch dataset or variable folder
+#' @param deep FALSE (the default) or TRUE; should subfolders
+#' @param include.hidden FALSE (default) or TRUE, should hidden be included in the result?
+#'
+#' @return Data.frame with one row per Crunch variable and columns \code{alias}, \code{path}, \code{hidden}
+#' @export
+findVariables <- function(x, deep = FALSE, include.hidden = FALSE) {
```
gergness, line 9: I think I'd prefer dropping the `include.hidden` param and if you ever wanted it, you'd have to explicitly do `rbind(findVariables(ds), findVariables(hiddenFolder(ds))`, and make this clear in documentation.

(otherwise needs an `include.private` option as well)

The API & documentation are kind of confusing about this, because it changed since I joined, but hidden variables are always in the hidden folder (before 2021, there could be hidden variables in any folder), so I think I'd prefer having the only indication in the output that it is hidden be the folder name.

R/findVariables.R
```diff
@@ -0,0 +1,57 @@
...
+    dirnames <- names(dirs)
+    res2 <- lapply(seq_along(dirnames), function(i) {
+        if (identical(path, "")) {
+            new_path <- dirnames[i]
+        } else {
+            new_path <- paste(path, dirnames[i], sep = " | ")
+        }
+        .findVariables(dirs[[i]], new_path)
+    })
+    rbind(res, do.call(rbind, res2))
```
gergness, line 56: It probably doesn't matter, but I think this kind of data.frame aggregation can get really bogged down in deeply nested things; I think it's better if you only flatten here (using purrr's terminology, but you should use base functions) and do one `do.call(rbind, )` in the main `findVariables()` function.

(the reason being that you have to allocate memory for consecutive vectors for each rbind, whereas flattening the list doesn't actually require much work because of R's copy-on-modify behavior)

gergness (commented):
Yeah, I struggle with how much this should "fit in" to the existing stuff or not. I think the way that would best fit into existing conventions would be to add a `allVariables()` method for a folder that returns a VariablesCatalog, but then as you say there wouldn't be a great way to get the full paths back out. We could do something like cache the full folder path inside the pseudo-variables catalog that we return and then add a function to get the full path for a variable that optionally uses that cached path, but that seems hack-ish enough that we might as well make it how you suggest 🤷 

Two process things:
- The build failure is because you need to run `make doc` to add this new file to the Collate field in the DESCRIPTION (just a wrapper for `devtools::document()` if you prefer)
- Let's do this through the full Jenkins build system, which requires your git branch name to end with a pivotal ticket number (eg`folder-variables-recursive-12345`)

R/findVariables.R
sluga: I can understand the case for not specifically dealing with hidden & private variables here, though for me the convenience of having include.hidden and include.private wins out. So the current version has them, but if you feel strongly about this, I'll drop them.

PR #642, opened 2024-05-09
[187448845]: don't assume that weight in prefs exists

gergness:


Eduardodudu:
I'd suggest adding one test for this feature on `test-weighting.R` It also makes it easier to evaluate

gergness:
tested here (integration test only): https://github.com/Crunch-io/rcrunch/blob/master/tests/testthat/test-weighting.R#L267-L274

Eduardodudu (approved):

PR #646, opened 2024-07-30
[188037693]: fix bugs with as.data.frame(ds, force = TRUE)

gergness:
Fixes bugs related to as.data.frame(ds, force=TRUE)

- No longer overwrites the wrong data with the right metadata when there are columns with duplicate aliases (eg two array subvariables with the same alias)
- Reads text variables with only numbers as character columns (and doesn't overwrite them all with missing)
- Handles missingness better (text variables have `NA` instead of string "No Data", sometimes numeric arrays would end up all missing if any were missing)

This PR seems bigger than it is because I've had to rerun the automatic fixture generation code for the first time in a while. 

The changes in https://github.com/Crunch-io/rcrunch/pull/646/commits/0325843bdcca484e8be7cd7d43d62f2c14d2ab5e represent only changes to the backend and had no effect on rcrunch behavior. They are only included so that rcrunch's mock tests represent what the server actually sends.

The changes in this commit https://github.com/Crunch-io/rcrunch/pull/646/commits/ed91176939a16f9dcb1a4dce9d42c708a9e26152 represent the new mock dataset I added which has duplicate subvariable aliases and updates to the csv.


sluga (approved):
Maybe you want to update this part? Could probably just refer the reader to the array_strategy argument help.

```
#' * Array variables (Categorical Array, Multiple Response) are decomposed into
#' their constituent categorical subvariables. An array with three subvariables,
#' for example, will result in three columns in the `data.frame`.
#'
#' Column names in the `data.frame` are the variable/subvariable aliases.
```

Issue #662, opened 2025-02-08
`as.data.frame(ds, force  = TRUE)` returns many columns of NA


rossellhayes:
Starting with the recent update to the crunch package, running `as.data.frame(ds, force = TRUE)` now returns a data frame with many (sometimes 100s or even 1000s of) columns full of `NA`s with `NA` names. All the expected data is there, but the columns of `NA`s are interspersed.

gergness:
Sorry, I believe this only happened when there was an array variable with a single subvariable, and then it was put horizontally (describing why there were so many columns).

Can you make sure that `remotes::install_github("Crunch-io/rcrunch#663")` fixes for you too?

PR #663, opened 2025-02-10
Fix for #662

gergness:


gergness:
This had the wrong branch name for the new CI

PR #665, opened 2025-02-19
[ZC-231]: Fix #655

### LINKED ISSUE #655: with_mock() is going away
This is a heads up to let you know that `with_mock()` is deprecated and will be made defunct soon. The API functions that made it work are being removed in R 4.5.0, so we'd suggest that you switch to `with_mocked_bindings()` instead.

gergness:


R/datetime.R
Eduardodudu, line 1: ok, also checked for parser of T pattern directly over packages and couldn't find them.

tests/testthat/test-datetime.R
```diff
@@ -43,6 +43,34 @@ test_that("from8601 parses", {
         class = c("POSIXlt", "POSIXt"), tzone = "UTC"
         )
     )
+    expect_equal(
```
Eduardodudu, line 46: It's missing an example where "T" is used

Eduardodudu:
LGTM, just check if test need the "T" pattern checkup

Eduardodudu (approved):

tests/testthat/test-datetime.R
gergness: New tests adapt existing ones with "T", eg https://github.com/Crunch-io/rcrunch/blob/cff0e0d7624c933849277110b08322864eaf2e2e/tests/testthat/test-datetime.R#L5

PR #679, opened 2026-01-16
remove test to make a filter private

jjdelc:


gergness (approved):
LGTM as long as API returns a reasonable error if someone tries to do this

PR #681, opened 2026-06-02
[ZC-3703]: Handle alias refs when printing expressions

slobodan-ilic:


tests/testthat/test-exclusion-filter.R
```diff
@@ -51,9 +51,16 @@ with_test_authentication({
...
             ## because JSON objects are unordered.
             expect_json_equivalent(zcl(exclusion(ds)), zcl(ds$v4 == "C"))
-            expect_prints(
-                exclusion(ds),
-                'Crunch logical expression: v4 == "C"'
+            vm <- variableMetadata(ds)
+            expect_output(
+                cat(
+                    "Crunch logical expression: ",
+                    formatExpression(exclusion(ds), var_catalog = vm),
```
gergness: Hm, this isn't a good equivalent test. We want to keep existing behavior: `exclusion(ds)` (which calls `show.CrunchLogicalExpr`) should not error.

gergness:
```suggestion
                    exclusion(ds)
```

R/show.R
gergness, line 411: I think this would work?

```
   catalog <- allVariables(loadDataset(object@dataset_url))
    cat("Crunch expression: ", formatExpression(object, catalog), "\n",
        sep = ""
    )
```

Note that both with this (and with you getting the catalog via `variableMetadata()` in the test, return a `VariableTuple` not a `VariableEntity` as we used to. I think the APIs are compatible enough that it will work though.

gergness (changes requested):

gergness:
Superseded by https://github.com/Crunch-io/rcrunch/pull/682

PR #682, opened 2026-06-08
Use var_by_alias view to get entity when encountering zcl with var sy…

gergness:
…ntax

### DIFF
--- a/R/AllClasses.R
+++ b/R/AllClasses.R
@@ -54,7 +54,17 @@ ShojiTuple <- setClass("ShojiTuple",
         body = "list"
     )
 )
-VariableTuple <- setClass("VariableTuple", contains = "ShojiTuple")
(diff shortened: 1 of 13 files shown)


Eduardodudu:
Interesting!

```
expr <- CrunchExpr(
                expression = list(
                    `function` = "==",
                    args = list(
                        list(var = "gender"),
                        list(value = 1)
                    )
                ),
                dataset_url = self(ds)
            )
> expr
Crunch expression: gender == "Male"

 expr <- CrunchExpr(
                expression = list(
                    `function` = "==",
                    args = list(
                        list(var = "gender"),
                        list(value = 1)
                    )
                )
            )
> expr
Crunch expression: gender == 1

```

Eduardodudu (approved):

